PR #1, opened 2020-06-24
Split Compass chart

kjaksik:
**Description**

- extracted provisioner, kyma-environemtn-broker, metrics to eperate chart.
IMPORTANT: we are going to force push the history from compass.

**Related issue(s)**


resources/istio-gateway/values.yaml
```diff
@@ -0,0 +1,22 @@
...
+      namespace: "compass-system"
+    proxy:
+      port: 15020
+
+  connector:
+    # If secrets do not exist they will be created
+    secrets:
+      rootCA:
+        name: kyma-gateway-certs
+        namespace: istio-system # For Ingress Gateway to work properly the namespace needs to be istio-system
```
Szymongib, line 14: > If secrets do not exist they will be created

They won't be created as there is not certs-setup-job. I am also not sure if mTLS Gateway should be created as a part of KCP as I do not think any component uses it (without the connector it does not really make sense).

resources/kyma-environment-broker/values.yaml
```diff
@@ -150,3 +150,63 @@ e2e:
     timeout: "100m"
     preUpgradeVersion: ""
     targetVersion: "master-156dc1a5"
+
+global:
```
Szymongib, line 154: I would suggest keeping global values at the top.

resources/kyma-environment-broker/values.yaml
```diff
@@ -150,3 +150,63 @@ e2e:
...
+      version: "PR-1425"
+    tests:
+      e2e_provisioning:
+        dir:
+        version: "9103d980"
+
+  defaultTenant: 3e64ebae-38b5-46a0-b1ed-9ccee153a0ae
+
+  provisioning:
+    enabled: true
```
Szymongib, line 194: I do not think this value still makes sense. The registration job is labeled with it but the whole chart will still be installed and broker will just not work.

resources/provisioner/values.yaml
```diff
@@ -0,0 +1,148 @@
+global:
```
Szymongib, line 1: There is no registration job for Provisioner.

resources/provisioner/templates/tests/test-provisioner.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if .Values.tests.enabled }}
+{{ if .Values.global.tests.enabled }}
```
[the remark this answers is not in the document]
dbadura: I will set this value as `provisioner.tests.enabled`

installation/resources/installer-overrides-provisioner.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: ConfigMap
+metadata:
+  name: compass-overrides
+  namespace: compass-installer
+  labels:
+    installer: overrides
+    component: compass
+    kyma-project.io/installation: ""
+data:
+  global.provisioning.enabled: true
```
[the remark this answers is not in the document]
dbadura: It's leftover, I will remove it.

PR #10, opened 2020-07-01
Bump compass version to the latest

fhivemind:
**Description**

* Changes the way how Avs configuration is being defined and parsed within KEB chart. This was changed due to Kyma Operator inability to override objects.
Now, KEB Avs configuration uses multiline string which is possible to override.

* Bumps compass version to the latest.

EDIT:
Only bumps compass version to the latest.

**Related issue(s)**


pkosiec:
Hi @fhivemind, tomorrow on 16:00 CET we want to finally synchronize this repo with latest KEB changes from Compass. We can't merge the PR yet, but I'll let you know once we finish the synchronization. Thank you for your patience!

fhivemind:
@pkosiec In that case, I can remove chart updates (since the same changes exist on Compass). However, I do require Compass version bump. Could you look into that?

pkosiec:
@fhivemind Sure thing, sounds good. LGTM

PR #19, opened 2020-07-03
Add Auto Upgrade options to the Provisioners Gardener Config schema

franpog859:
**Description**

- Add Auto Upgrade options to the Provisioners Gardener Config schema


components/provisioner/assets/database/provisioner.sql
```diff
@@ -35,6 +35,8 @@ CREATE TABLE gardener_config
     auto_scaler_max integer NOT NULL,
     max_surge integer NOT NULL,
     max_unavailable integer NOT NULL,
+    auto_update_kubernetes_version boolean NOT NULL,
```
Szymongib, line 38: I wonder if the name is not a little misleading. What do you think about `enable_kubernetes_auto_update`? The same is the case for other files.

franpog859: Yeah, your suggestion is totally right. I used the names of the Gardener structure fields but in this case, this will be better

franpog859:
Here is minor fix enabling to see the auto-update configurations with `runtimeStatus` query - https://github.com/kyma-project/control-plane/pull/39

PR #28, opened 2020-07-06
Add UpgradeShoot() mutation

rafalpotempa:
**Description**

- New graphQL mutation `upgradeShoot` - all fields are optional
- Editing Gardener Shoot CRD and update of stored Gardener Cluster Configuration for Runtime 
- Database migration with adding need enum value for supported operation type
- Implementation of waiting for finishing Gardener shoot upgrade operation based on queue
- Unit tests for new mutation and new operation steps


koala7659:
Following parameters are supported as Gardener Upgrade Shoot input data:

```
input GardenerUpgradeInput {
    kubernetesVersion: String                     # Kubernetes version to be installed on the cluster
    machineType: String                           # Type of node machines, varies depending on the target provider
    diskType: String                              # Disk type, varies depending on the target provider
    volumeSizeGB: Int                             # Size of the available disk, provided in GB
    autoScalerMin: Int                            # Minimum number of VMs to create
    autoScalerMax: Int                            # Maximum number of VMs to create
    maxSurge: Int                                 # Maximum number of VMs created during an update
    maxUnavailable: Int                           # Maximum number of VMs that can be unavailable during an update
    purpose: String                               # The purpose given to the cluster (development, evaluation, testing, production)
    enableKubernetesVersionAutoUpdate: Boolean    # Enable KubernetesVersion AutoUpdate indicates whether the patch Kubernetes version may be automatically updated
    enableMachineImageVersionAutoUpdate: Boolean  # Enable MachineImageVersion AutoUpdate indicates whether the machine image version may be automatically updated
    providerSpecificConfig: ProviderSpecificInput # Additional parameters, vary depending on the target provider
}

```

docs/provisioner/provisioning-api.md
```diff
@@ -87,6 +87,46 @@ The mutation returns OperationID allowing to retrieve the operation status.
```
majakurcius, line 87: image

I got that as input for the documentation related to this PR and it's already [here](https://github.com/kyma-project/control-plane/pull/40), so the changes in the `provisioning-api.md` file can be removed from this PR.

components/provisioner/internal/operations/queue/queues_constructor.go
```diff
@@ -23,6 +24,8 @@ type ProvisioningTimeouts struct {
 	ClusterCreation    time.Duration `envconfig:"default=60m"`
 	Installation       time.Duration `envconfig:"default=60m"`
 	Upgrade            time.Duration `envconfig:"default=60m"`
+	ShootUpgrade       time.Duration `envconfig:"default=30m"`
+	ShootRecreate      time.Duration `envconfig:"default=5m"`
```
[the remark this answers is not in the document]
rafalpotempa: This is the timeout for recreation of `ShootConfig` by Gardener. I changed name to `ShootRefresh`

components/provisioner/internal/api/resolver_integration_with_gardener_test.go
```diff
@@ -212,36 +218,43 @@ func TestProvisioning_ProvisionRuntimeWithDatabase(t *testing.T) {
...
 			require.NoError(t, err)
 
-			shoot := &list.Items[0]
+			var shoot *gardener_types.Shoot
+			for _, s := range list.Items {
+				if s.Annotations["compass.provisioner.kyma-project.io/runtime-id"] == runtimeID {
+					shoot = &s
+					break
+				}
+			}
```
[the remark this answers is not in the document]
rafalpotempa: True, this is some residual from debugging process. There shall indeed be only one.

components/provisioner/internal/provisioning/service_test.go
```diff
@@ -626,6 +628,135 @@ func TestService_UpgradeRuntime(t *testing.T) {
...
+		writeSession.AssertExpectations(t)
+		upgradeShootQueue.AssertExpectations(t)
+	})
+
+	for _, testCase := range []struct {
+		description string
+		mockFunc    func(sessionFactory *sessionMocks.Factory, readWriteSession *sessionMocks.ReadWriteSession, writeSession *sessionMocks.WriteSessionWithinTransaction, provisioner *mocks2.Provisioner)
+	}{
+		{
+			description: "should fail to upgrade Shoot when failed to commit transaction",
```
akgalwas, line 702: Test case for failed shoot update is missing.

akgalwas:
tests/director/api.test file is redundant.

components/provisioner/internal/operations/stages/shootupgrade/new_version_shoot.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/kyma-project/control-plane/components/provisioner/internal/operations"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type WaitForShootClusterNewVersionStep struct {
+	gardenerClient         GardenerClient
+	nextStep               model.OperationStage
+	timeLimit              time.Duration
+	initialResourceVersion string
```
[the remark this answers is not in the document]
rafalpotempa: I created map that stores old config. But indeed there might be some problems on Provisioner restart. The case seems to happen when Gardener shoot config refreshes during a restart. 

@Szymongib What do you think about storing `ResourceVersion` in DB in `operations` table or about adding additional table with initial `ResourceVersion` for each shootUpgrade operation?

components/provisioner/internal/provisioning/service_test.go
rafalpotempa: To be sure. We mean failed DB shoot config update?

components/provisioner/internal/model/gardener_config_test.go
```diff
@@ -94,7 +94,6 @@ func Test_NewGardenerConfigFromJSON(t *testing.T) {
 			assert.Equal(t, testCase.expectedProviderSpecificConfig, providerSpecificConfig)
 		})
 	}
-
```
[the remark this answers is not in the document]
rafalpotempa: I think I missed some test here.

components/provisioner/internal/gardener/provisioner_test.go
```diff
@@ -170,6 +170,43 @@ func TestGardenerProvisioner_DeprovisionCluster(t *testing.T) {
...
+	cluster := newClusterConfig("test-cluster", nil, gcpGardenerConfig, region)
+
+	t.Run("should upgrade shoot", func(t *testing.T) {
+		// given
+		shootClient := clientset.CoreV1beta1().Shoots(gardenerNamespace)
+
+		provisionerClient := NewProvisioner(gardenerNamespace, shootClient, nil, auditLogsPolicyCMName, maintWindowConfigPath)
+
+		// when
+		apperr := provisionerClient.ProvisionCluster(cluster, operationId)
```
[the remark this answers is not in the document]
rafalpotempa: You are right. This is obviously a mistake.

components/provisioner/internal/operations/stages/shootupgrade/new_version_shoot.go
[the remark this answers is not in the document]
Szymongib: I did some investigation around that topic and found that Gardener Shoot has a status field called `ObservedGeneration` and I believe that is exactly what we need, as after we update the Shoot the `generation`  field is incremented by the API Server and we want to make sure that the Gardener Controller picks up the change before we check the operation status and that is where the `ObservedGeneration` comes in, as this status field is set by the Gardener. Therefore in the first step, we can wait until the `generation` is equal to `ObservedGeneration` and only after that wait for the operation to finish.

components/provisioner/internal/provisioning/service.go
```diff
@@ -156,10 +162,60 @@ func (r *service) DeprovisionRuntime(id, tenant string) (string, apperrors.AppEr
...
+	session := r.dbSessionFactory.NewReadWriteSession()
+
+	err := r.verifyLastOperationFinished(session, runtimeID)
+	if err != nil {
+		return &gqlschema.OperationStatus{}, err
+	}
+
+	cluster, dberr := session.GetCluster(runtimeID)
+	if dberr != nil {
+		return &gqlschema.OperationStatus{}, apperrors.Internal("failed to find shoot cluster to upgrade: %s", dberr.Error())
```
Szymongib, line 181: This error message might be confusing indicating that we failed to find Shoot CR not a Cluster in the DB.
I also think we should stick to one convention with capitalizing errors as most of them in this file start with the upper case.

components/provisioner/internal/provisioning/service.go
```diff
@@ -156,10 +162,60 @@ func (r *service) DeprovisionRuntime(id, tenant string) (string, apperrors.AppEr
...
 }
 
+func (r *service) UpgradeGardenerShoot(runtimeID string, input gqlschema.UpgradeShootInput) (*gqlschema.OperationStatus, apperrors.AppError) {
+	log.Infof("Starting Upgrade of Gardener Shoot for Runtime '%s'...", runtimeID)
+
+	if input.GardenerConfig == nil {
+		return &gqlschema.OperationStatus{}, apperrors.Internal("error: Gardener config is nil")
+	}
+
+	session := r.dbSessionFactory.NewReadWriteSession()
```
Szymongib, line 172: Does it need to be the `ReadWrite` session? I think we use it only for reading.

components/provisioner/internal/operations/stages/shootupgrade/upgrade_shoot.go
```diff
@@ -0,0 +1,68 @@
+package shootupgrade
```
Szymongib, line 1: I would suggest changing the name of the file to match the step name

components/provisioner/internal/provisioning/input_converter.go
```diff
@@ -137,6 +177,24 @@ func (c converter) providerSpecificConfigFromInput(input *gqlschema.ProviderSpec
 	return nil, apperrors.BadRequest("provider config not specified")
 }
 
+func (c converter) providerSpecificConfigFromUpgradeInput(input *gqlschema.ProviderSpecificInput) (model.GardenerProviderConfig, error) {
```
Szymongib, line 180: This function is exactly the same as `providerSpecificConfigFromInput` and I do not think it is used.

components/provisioner/internal/provisioning/input_converter.go
```diff
@@ -88,27 +90,65 @@ func (c converter) gardenerConfigFromInput(runtimeID string, input *gqlschema.Ga
...
 		GardenerProviderConfig:              providerSpecificConfig,
 	}, nil
 }
 
+func (c converter) UpgradeShootInputToGardenerConfig(input gqlschema.GardenerUpgradeInput, config model.GardenerConfig) (model.GardenerConfig, apperrors.AppError) {
+
+	providerSpecificConfig, err := c.providerSpecificConfigFromInput(input.ProviderSpecificConfig)
+	if err != nil {
+		providerSpecificConfig = config.GardenerProviderConfig
+	}
```
Szymongib, line 119: I do not like the way we handle it. In case of an error, we will just fail silently and use the old config.

components/provisioner/internal/operations/stages/shootupgrade/new_version_shoot.go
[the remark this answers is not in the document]
koala7659: I have prepared and run test code for first step and proposed approach seem to work correctly. For trivial updates like changing shoot purpose, we have immediately situation when `ObservedGeneration == generation`. For more complicated upgrades - ex. while changing instance type in first attempt we get `generation == ObservedGeneration + 1`. After 5 seconds in second attempt we can see `ObservedGeneration == generation`. So in both cases we are sure we have correct Shoot version and we can start waiting for Shoot Reconcile.

components/provisioner/internal/provisioning/service_test.go
akgalwas: > To be sure. We mean failed DB shoot config update?

Ok, after second look it seems the testcase I was thinking about is there but the name is a bit confusing: "should fail to upgrade Shoot when failed to upgrade cluster". "Should fail when failed to update shoot"?

components/provisioner/internal/operations/stages/shootupgrade/new_version_shoot.go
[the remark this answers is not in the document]
rafalpotempa: This solves the problem. Thanks a lot!

components/provisioner/internal/provisioning/service_test.go
```diff
@@ -238,20 +238,20 @@ func TestService_DeprovisionRuntime(t *testing.T) {
 	t.Run("Should start Runtime deprovisioning and return operation ID", func(t *testing.T) {
 		//given
 		sessionFactoryMock := &sessionMocks.Factory{}
-		readWriteSession := &sessionMocks.ReadWriteSession{}
+		readSession := &sessionMocks.ReadWriteSession{}
```
akgalwas, line 241: Shouldn't we use sessionMocks.ReadSession instead?

rafalpotempa: Indeed. I fixed it

PR #31, opened 2020-07-07
Fix KEB starting

polskikiel:
**Description**

- Fixed KEB starting

**Related issue(s)**


mszostok:
I don't understand how this will solve issue described here: https://github.com/kyma-incubator/compass/issues/1341 
Let's discuss that f2f 🙂

ebensom:
I'm also thinking to increase the liveness probe `initialDelaySeconds` (maybe to 30s?) because the container restarts could be initiated by kubelet because the container is not responding to the liveness probe according to the configured settings. After 2 restarts, the istio-proxy sidecar is ready to proxy traffic so the liveness probe can reach KEB's healthz handler which responds back with 200 OK (edited) 

 I have no knowledge what issue will it solve to wait for the K8S client mapper to be ready

PR #34, opened 2020-07-08
Enable TLS renegotiation for IAS HTTP client

jasiu001:
**Description**

- Enable TLS renegotiation for IAS HTTP client

### DIFF
--- a/components/kyma-environment-broker/cmd/broker/main.go
+++ b/components/kyma-environment-broker/cmd/broker/main.go
@@ -188,11 +188,9 @@ func main() {
 	internalEvalAssistant := avs.NewInternalEvalAssistant(cfg.Avs)
 	externalEvalCreator := provisioning.NewExternalEvalCreator(avsDel, cfg.Avs.Disabled, externalEvalAssistant)
 
-	var clientHTTPForIAS *http.Client
+	clientHTTPForIAS := httpClient
 	if cfg.IAS.TLSRenegotiationEnable {
 		clientHTTPForIAS = httputil.NewRenegotiationTLSClient(30, cfg.Director.SkipCertVerification)
-	} else {
-		clientHTTPForIAS = httpClient
 	}
 	bundleBuilder := ias.NewBundleBuilder(clientHTTPForIAS, cfg.IAS)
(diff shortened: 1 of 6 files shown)


components/kyma-environment-broker/internal/httputil/client.go
```diff
@@ -14,3 +15,17 @@ func NewClient(timeoutSec time.Duration, skipCertVerification bool) *http.Client
 		Timeout:   timeoutSec * time.Second,
 	}
 }
+
+func NewRenegotiationTLSClient(timeoutSec time.Duration, skipCertVerification bool) *http.Client {
+	transport := &http.Transport{
```
mszostok, line 20: you should init transport struct in th same way as above:
```
transport := http.DefaultTransport.(*http.Transport).Clone()
```

that gives you defaults for some fields, and then you can adjust rest of them

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -188,7 +188,13 @@ func main() {
 	internalEvalAssistant := avs.NewInternalEvalAssistant(cfg.Avs)
 	externalEvalCreator := provisioning.NewExternalEvalCreator(avsDel, cfg.Avs.Disabled, externalEvalAssistant)
 
-	bundleBuilder := ias.NewBundleBuilder(httpClient, cfg.IAS)
+	var clientHTTPForIAS *http.Client
+	if cfg.IAS.TLSRenegotiationEnable {
+		clientHTTPForIAS = httputil.NewRenegotiationTLSClient(30, cfg.Director.SkipCertVerification)
+	} else {
+		clientHTTPForIAS = httpClient
+	}
```
mszostok, line 196: **suggestion**
```suggestion
	clientHTTPForIAS := httpClient
	if cfg.IAS.TLSRenegotiationEnable {
		clientHTTPForIAS = httputil.NewRenegotiationTLSClient(30, cfg.Director.SkipCertVerification)
	}
```

maybe it will be simpler to have a default and override with flag enabled? but I don't have a strong opinion about that if/else statements

PR #59, opened 2020-07-17
Remove Config Map creation after runtime is provisioned

Maladie:
**Description**

- Config Map creation removed

**Related issue(s)**

kyma-incubator/compass#1279


components/provisioner/internal/runtime/configuration_test.go
```diff
@@ -29,7 +29,7 @@ func TestProvider_CreateConfigMapForRuntime(t *testing.T) {
 	tenant := "tenant"
 	token := "shdfv7123ygfbw832b"
 
-	namespace := "compass-system"
+	namespace := "kcp-system"
```
Szymongib, line 32:
```suggestion
	namespace := "compass-system"
```
Compass Runtime Agent is still installed in `compass-system` so maybe let's leave it as it was to avoid confusion.

PR #68, opened 2020-07-21
Reorganize internal documents

pkosiec:
**Description**

- Reorganize internal documents:
   - Remove a leftover from Compass split
   - Rename directories and files
   - Update Readme that lists all internal documents


docs/internal/README.md
```diff
@@ -3,5 +3,8 @@
 This folder contains internal documentation, such as proposals, investigations, and standards agreed upon by the Development Team.
 
 - [Custom configuration of provisioned Runtimes](./investigations/provisioning-custom-configuration.md)
-- [Compass as a separate component](./proposals/separate-compass.md)
 - [Runtime upgrade and rollback with the Runtime Provisioner](./runtime-upgrade.md)
+- [Dapr: Remote Runtime management concept](./investigations/dapr-remote-kyma-runtime-configuration.md)
+- [Dapr: Sidecars limitations](./investigations/dapr-sidecars-limitations.md)
+- [Dapr: Proof of Concept scenario](./investigations/dapr-poc-scenario.md)
+- [Runtime Director - Proof of concept](./investigations/runtime-director-poc/README.md)
```
PK85, line 10: Runtime Director -> Runtime Governor. Find all places and rename. Leave Director name only in the context of compass.

PK85, line 10: All docs x-team added about Runtime Governor and POC place in one folder, maybe just name it: Governor

docs/internal/investigations/dapr-sidecars-limitations.md
```diff
@@ -1,5 +1,4 @@
-# Remote Kyma Runtime configuration
-
+# Dapr sidecars limitations
```
PK85, line 1: That title is crazy ;) Maybe just name it like: Dapr reloading investigation, or something like that

docs/internal/README.md
klaudiagrz: Just a few things to reconsider:
1. Do we want to use `Runtime Governor` or just `Governor`? In case of Director, we skip the Runtime part and name the component `Director` throughout the whole documentation. Should we be concise with our approach?
2. What about the Runtime Agent? For now, we have `Kyma Control Plane Runtime Agent` which, in my opinion, is way too long.

PR #83, opened 2020-07-28
Fix panic on Environments Cleanup job

kjaksik:
**Description**

- Fix panic when no `runtime-id` label on shoot

### DIFF
--- a/components/kyma-environment-broker/internal/environmentscleanup/service.go
+++ b/components/kyma-environment-broker/internal/environmentscleanup/service.go
@@ -32,15 +32,17 @@ type Service struct {
 	gardenerService GardenerClient
 	brokerService   BrokerClient
 	instanceStorage storage.Instances
+	logger          *log.Logger
 	MaxShootAge     time.Duration
 	LabelSelector   string
 }
 
-func NewService(gardenerClient GardenerClient, brokerClient BrokerClient, instanceStorage storage.Instances, maxShootAge time.Duration, labelSelector string) *Service {
+func NewService(gardenerClient GardenerClient, brokerClient BrokerClient, instanceStorage storage.Instances, logger *log.Logger, maxShootAge time.Duration, labelSelector string) *Service {
 	return &Service{
 		gardenerService: gardenerClient,
 		brokerService:   brokerClient,
 		instanceStorage: instanceStorage,
+		logger:          logger,
 		MaxShootAge:     maxShootAge,
 		LabelSelector:   labelSelector,
 	}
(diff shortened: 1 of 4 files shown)


components/kyma-environment-broker/internal/environmentscleanup/service.go
```diff
@@ -50,7 +52,7 @@ func (s *Service) PerformCleanup() error {
 	var result *multierror.Error
 	shootsToDelete, err := s.getShootsToDelete(s.LabelSelector)
 	if err != nil {
-		log.Error(errors.Wrap(err, "while getting shoots to delete"))
+		s.logger.Error(errors.Wrap(err, "while getting shoots to delete"))
 		result = multierror.Append(result, err)
```
ksputo, line 56: as disscussed f2f we could refactor this - `getShootsToDelete` returns error only in case when Gardener service was unable to list Shoots - considering this, we can fail fast here and return this error.

components/kyma-environment-broker/internal/environmentscleanup/service.go
```diff
@@ -59,25 +61,28 @@ func (s *Service) PerformCleanup() error {
...
+
+	if len(runtimeIDsToDelete) == 0 {
+		return result.ErrorOrNil()
+	}
 
 	instancesToDelete, err := s.getInstancesForRuntimes(runtimeIDsToDelete)
 	if err != nil {
-		log.Error(errors.Wrap(err, "while getting instance IDs for Runtimes"))
+		s.logger.Error(errors.Wrap(err, "while getting instance IDs for Runtimes"))
 		result = multierror.Append(result, err)
```
ksputo, line 78: Underlying logic in `getInstancesForRuntimes` has retries when accessing the database. Entering this code block means that we could not reach the db and get instances, so we can also return error immediately right here (with possible errors appended before in the loop)

components/kyma-environment-broker/internal/environmentscleanup/service.go
```diff
@@ -59,25 +61,28 @@ func (s *Service) PerformCleanup() error {
...
 	for _, instance := range instancesToDelete {
-		log.Infof("Triggering environment deprovisioning for instance ID %q", instance.InstanceID)
+		s.logger.Infof("Triggering environment deprovisioning for instance ID %q", instance.InstanceID)
 		currentErr := s.triggerEnvironmentDeprovisioning(instance)
 		if currentErr != nil {
-			log.Error(errors.Wrapf(currentErr, "while triggering deprovisioning for instance ID %q", instance.InstanceID))
+			s.logger.Error(errors.Wrapf(currentErr, "while triggering deprovisioning for instance ID %q", instance.InstanceID))
 			result = multierror.Append(result, currentErr)
 		}
 	}
```
ksputo, line 88: as discussed, we could extract formatting function:
```
 func(i []error) string {
			var s []string
			for _, v := range i {
				s = append(s, v.Error())
			}
			return strings.Join(s, ", ")
		}
```
and assign it to `result.ErrorFormat` each time it is going to be returned

Issue #82, opened 2020-07-28
Provisioner recognizes categorized Director errors 


franpog859:
**Description**

This issue is a follow-up for [this one](https://github.com/kyma-incubator/compass/issues/1438)

Provisioner should recognize categorized Director errors and map them into its own responses. 
GraphQL does return HTTP code 200 even if there was an error so the error handling is much more difficult here. Director has implemented custom error responses with categorized error codes. The change was implemented [in this PR](https://github.com/kyma-incubator/compass/pull/1408). Provisioner should use these values and map them into its own categorized error responses which were implemented [in this PR](https://github.com/kyma-project/control-plane/pull/18)

You may consider writing your own custom GraphQL client to read the error codes, forking the existing one, for example [machinebox/graphql](https://github.com/machinebox/graphql) and using the changed code on your fork, contributing to the [existing one](https://github.com/machinebox/graphql) and using it after the contribution

**Reasons**

Provisioner will return more accurate errors instead of always returning internal server error code. If the solution will be generic enough, it could also be used for example in Kyma Environment Broker component to recognize Provisioner's errors

**Acceptance criteria**
- [Director Client](https://github.com/kyma-project/control-plane/blob/master/components/provisioner/internal/director/directorclient.go) reads the error codes and maps them on Provisioner's ones

PR #96, opened 2020-07-29
Map Director error codes in Provisioner error responses

franpog859:
**Description**

- add `machinebox/graphql` library to the Provisioner
- map Director's error codes to Provisioner's response codes

**Related issue(s)**

See the issue #82


components/provisioner/third_party/machinebox/graphql/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Installation
+Make sure you have a working Go environment. To install graphql, simply run:
+
+```
+$ go get github.com/machinebox/graphql
+```
+
+## Usage
+
```
majakurcius, line 20: Please add an introductory sentence describing what to do with the code below.

franpog859: Hey! I don't think I should change it. It's just a copied library from the [machinebox repository](https://github.com/machinebox/graphql). The only change I made is a minor contribution which allows us to use it in our specific way

majakurcius: I see. In that case, please make sure that the appropriate credits and sources are included (see point 4 in the license). 
I think a note in the form of a comment could also be added in modified parts of code. For example, in `client.go` where we have line 9 (`"github.com/kyma-project/control-plane/components/provisioner/third_party/machinebox/graphql"`), we could add a comment saying something like "The original code comes from `https://github.com/machinebox/graphql`, however, this line was modified." or something like this.

components/provisioner/internal/director/directorclient.go
```diff
@@ -247,10 +248,34 @@ func (cc *directorClient) executeDirectorGraphQLCall(directorQuery string, tenan
...
 
 	return nil
 }
+
+func mapDirectorErrorToProvisionerError(egErr gcli.ExtendedError) apperrors.AppError {
+	errorCodeValue, present := egErr.Extensions()["error_code"]
+	if !present {
+		return apperrors.Internal("Failed to read the error code from the error response. Original error: %v", egErr)
+	}
+	errorCode, ok := errorCodeValue.(float64)
```
Szymongib, line 266: Questions: why casting the error code to `float64`?

components/provisioner/internal/director/directorclient.go
```diff
@@ -247,10 +248,34 @@ func (cc *directorClient) executeDirectorGraphQLCall(directorQuery string, tenan
...
+	if !ok {
+		return apperrors.Internal("Failed to cast the error code from the error response. Original error: %v", egErr)
+	}
+	switch directorApperrors.ErrorType(errorCode) {
+	case directorApperrors.InternalError, directorApperrors.UnknownError:
+		return apperrors.Internal(egErr.Error())
+	case directorApperrors.Unauthorized, directorApperrors.InsufficientScopes:
+		return apperrors.Forbidden(egErr.Error())
```
Szymongib, line 274: I am not sure about returning the `Forbidden` status in both cases. It seems like a good fit for `InsufficientScopes` but not sure about the `Unauthorized`.

components/provisioner/internal/director/directorclient.go
franpog859: I checked the type of incoming error code from Director and it was `float64` even if it looks just like `21` for example. I tried to cast it to `int` anyway but it didn't work at all

components/provisioner/internal/director/directorclient.go
franpog859: Hm, will the `BadRequest` work better for you? The `Forbidden` one was just my first thought for these two

Szymongib: What about some equivalent of 502 or even Internal error? It is not actually user error if Provisioner gets 401/403 from Director but more so some issue with the configuration of Control Plane itself, so I am thinking something like `Bad Gateway` would make the most sense.

components/provisioner/internal/director/directorclient.go
Szymongib: Thanks for insights. It's probably some GraphQL quirks.

PR #109, opened 2020-08-04
Make allow privileged containers configurable

franpog859:
**Description**

- Make allow privileged containers configurable


components/provisioner/internal/provisioning/input_converter.go
```diff
@@ -107,6 +111,7 @@ func (c converter) gardenerConfigFromInput(runtimeID string, input *gqlschema.Ga
 		LicenceType:                         input.LicenceType,
 		EnableKubernetesVersionAutoUpdate:   util.UnwrapBoolOrDefault(input.EnableKubernetesVersionAutoUpdate, c.defaultEnableKubernetesVersionAutoUpdate),
 		EnableMachineImageVersionAutoUpdate: util.UnwrapBoolOrDefault(input.EnableMachineImageVersionAutoUpdate, c.defaultEnableMachineImageVersionAutoUpdate),
+		AllowPrivilegedContainers:           util.UnwrapBoolOrDefault(input.AllowPrivilegedContainers, c.shouldAllowPrivilegedContainers(tillerYaml)),
```
Szymongib, line 114: I wonder if we do not want some mechanism to override this field for example by Env var. As in the case, the 1.15 without the Tiller will not work correctly with disallowed privileged containers the Provisioner will end up blocking the release and the code changes will be needed to unblock it.

I would suggest adding something temporarily to prevent such a scenario as well as add follow up issue for cleaning it up and testing when `allowPrivilegedContainers` if set to false.

components/provisioner/internal/provisioning/graphql_converter_test.go
```diff
@@ -422,3 +428,12 @@ func fixKymaRelease() model.Release {
...
 }
+
+func fixKymaReleaseWithoutTiller() model.Release {
+	return model.Release{
+		Id:            "e829b1b5-2e82-426d-91b0-f94978c0c140",
+		Version:       kymaVersionWithoutTiller,
+		TillerYAML:    "",
+		InstallerYAML: "installer yaml",
+	}
+}
```
Szymongib, line 439: I think this function is used only in `input_converter_tests.go` so that is a probably better place for it.

components/provisioner/internal/provisioning/input_converter.go
```diff
@@ -74,7 +78,7 @@ func (c converter) ProvisioningInputToCluster(runtimeID string, input gqlschema.
 	}, nil
 }
 
-func (c converter) gardenerConfigFromInput(runtimeID string, input *gqlschema.GardenerConfigInput) (model.GardenerConfig, apperrors.AppError) {
+func (c converter) gardenerConfigFromInput(runtimeID string, input *gqlschema.GardenerConfigInput, tillerYaml string) (model.GardenerConfig, apperrors.AppError) {
```
Szymongib, line 81: What about just passing boolean if the privileged containers should be allowed?

components/provisioner/internal/provisioning/input_converter.go
franpog859: It makes sense. But there will be three sources of this value then - the environment variable, the graphql mutation field, and the shouldAllowPrivilegedContainers mechanism
What priorities do you suggest? At least two of them won't be nil because the env variable will always have a value and the shouldAllowPrivilegedContainers mechanism will always return a value

components/provisioner/internal/provisioning/input_converter.go
franpog859: I was thinking about that and I just couldn't decide on which layer of abstraction should the shouldAllowPrivilegedContainers mechanism appear :D If you say it'll be cleaner I change it

PR #113, opened 2020-08-05
POC of relasing tenant subscriptions into the pool

koala7659:
The following functionalities were implemented:
- HAP marks secret as released on deprovisioning if there is only one runtime using particular subscription
- Subscription Clean up Job introduced - the job removes Azure resource groups for secrets marked as released


components/kyma-environment-broker/internal/hyperscaler/account_pool.go
```diff
@@ -38,19 +40,114 @@ type Credentials struct {
...
+
+	if err != nil {
+		return false, errors.Wrapf(err, "error while looking for a secret used by the tenant %s and hyperscaler %s", tenantName, hyperscalerType)
+	}
+
+	if secret == nil {
+		return false, errors.Errorf("accountPool failed to find subscription secret used by the tenant %s and hyperscaler %s", tenantName, hyperscalerType)
+	}
+
+	labelSelector = fmt.Sprintf("shared!=true, released=true, tenantName=%s,hyperscalerType=%s", tenantName, hyperscalerType)
```
Szymongib, line 75: `released` label name is a little misleading as it implies that it is clean. I guess the better name would be `dirty` or `toRelease`.

akgalwas: I think "dirty" is a good name.

components/subscription-cleanup-job/internal/cloudprovider/gcp.go
```diff
@@ -0,0 +1,12 @@
+package cloudprovider
+
+type gcpResourceCleaner struct {
+}
+
+func NewGCPeResourcesCleaner(secretData map[string][]byte) ResourceCleaner {
+	return &azureResourceCleaner{}
```
koala7659, line 7: I'm not sure about that. I think you meant just to return `&gcpResourceCleaner {}` here instead of `azureResourceCleaner`. It is not supported anyway and will return `nil` at the end but still it is confusing.

koala7659 (commented):
LGTM

PR #157, opened 2020-08-24
Generate Shoot template in Provisioner

Szymongib:
**Description**

- Generate Shoot template
- Render template with values

**Related issue(s)**


components/provisioner/internal/templates/render.go
```diff
@@ -0,0 +1,29 @@
...
+func RenderTemplate(text string, values Values) ([]byte, error) {
+	templ, err := template.New("").Parse(text)
+	if err != nil {
+		return nil, fmt.Errorf("error while parsing renderTemplate: %s", err.Error())
+	}
+
+	buffer := new(bytes.Buffer)
+	err = templ.Execute(buffer, values)
+	if err != nil {
+		return nil, fmt.Errorf("error while rendering renderTemplate: %s", err.Error())
```
rafalpotempa, line 25: I'm not sure about meaning of this error message. Did you mean (`error while rendering template '%s'`, text)
```suggestion
		return nil, fmt.Errorf("error while rendering template '%s': %s", text, err.Error())
```
?

components/provisioner/templates/cluster-bom.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: "hub.k8s.sap.com/v1"
+kind: ClusterBom
+metadata:
+  name: {{ .ShootName }}
+spec:
+  secretRef: {{ .ShootName }}.kubeconfig
+  applicationConfigs:
+  - id: kyma
+    configType: helm
+    readyRequirements:
```
rafalpotempa, line 10: OIDC config could go here

components/provisioner/templates/shoot.yaml
```diff
@@ -0,0 +1,65 @@
...
+  region: '{{ .Region }}'
+  secretBindingName: '{{ .GardenerSecretName }}'
+status:
+  gardener:
+    id: ""
+    name: ""
+    version: ""
+  hibernated: false
+  technicalID: ""
+  uid: ""
```
rafalpotempa, line 65: How do you think: should we use the full possible shoot config, or just the pure minimum?

Szymongib: I think we should keep it as close as possible to the Shoot created by the Provisioner.

components/provisioner/templates/cluster-bom.yaml
Szymongib: Yes, I am waiting to confirm the configuration.

components/provisioner/internal/templates/render.go
Szymongib: Good catch. Did a little too much renaming, I will fix it with the next changes.

PR #184, opened 2020-09-07
Fix race when downloading artifacts

Szymongib:
**Description**

- Fix possible race condition when downloading artifacts

**Related issue(s)**


rafalpotempa (approved):
OK

PR #204, opened 2020-09-16
Move KEB internal packages to common directory

ksputo:
**Description**

- Added missing methods used in `tests/e2e/provisioning/internal` to `director` package in `components/kyma-environment-broker`
- Moved `director`, `gardener` packages containing clients from `components/kyma-environment-broker/internal` to `components/kyma-environment-broker/common` in order to allow importing shared logic in `tests/e2e/provisioning`

**Related issue(s)**

[#1303](https://github.com/kyma-incubator/compass/issues/1303)


kjaksik (commented):
Looks good, however I'm wondering about the `common` directory name. Is it some kind of a standard or did you come up with this? I encountered some projects where the client code is located in the `pkg` directory but I don't have a strong opinion on this

kjaksik:
As we've discussed face to face we will leave it be for the moment. However it would be a good idea to move the clients one level higher so they are not common for KEB, but for the whole Control Plane repo. Right now they are duplicated. @ksputo Could you please open an issue about that and mention that we have to discuss this topic with the framefrog team, what do they think about that.

kjaksik (approved):

ksputo:
@kjaksik 

Here's the follow-up issue for other teams: #216 

cc: @PK85 @anishj0shi @janmedrek

PR #211, opened 2020-09-18
Keb upgrade test

piotrmiskiewicz:
**Description**

- introduce a unit/module test which covers all logic connected with orchestration and upgrades
- one orchestration can be processed in the same time
- orchestration setup moved to separate function which is reused in the test


components/kyma-environment-broker/internal/process/upgrade_kyma/upgrade_kyma_step.go
```diff
@@ -32,7 +42,7 @@ func (s *UpgradeKymaStep) Name() string {
 }
 
 func (s *UpgradeKymaStep) Run(operation internal.UpgradeKymaOperation, log logrus.FieldLogger) (internal.UpgradeKymaOperation, time.Duration, error) {
-	if time.Since(operation.UpdatedAt) > UpgradeKymaTimeout {
```
polskikiel, line 35: IMO `UpgradeKymaTimeout` field doesn't really fit in the `IntervalConfig` struct, maybe we can call it `TimeSchedule` or `OperationSchedule`? Anyway if we want it to be parametrized we can get rid of the `UpgradeKymaTimeout` const from that file: https://github.com/kyma-project/control-plane/blob/master/components/kyma-environment-broker/internal/process/upgrade_kyma/upgrade_kyma_step.go#L16.

polskikiel (approved):

Issue #180, opened 2020-09-03
Implement API to list runtimes


polskikiel:
**Description**

Implement in KEB API to expose data about runtimes

Think about paging. Max page 100 is a default. Could be overridden by ENV.

PR #234, opened 2020-09-28
Change Runtime pagination to page and limit

kjaksik:
**Description**

- Change Runtime pagination to page and limit

**Related issue(s)**

#180

### DIFF
--- a/components/kyma-environment-broker/internal/runtime/handler.go
+++ b/components/kyma-environment-broker/internal/runtime/handler.go
@@ -46,14 +46,14 @@ func (h *Handler) AttachRoutes(router *mux.Router) {
 }
 
 func (h *Handler) getRuntimes(w http.ResponseWriter, req *http.Request) {
-	var toReturn []RuntimeDTO
+	toReturn := make([]RuntimeDTO, 0)
 	limit, page, err := h.getParams(req)
 	if err != nil {
 		httputil.WriteErrorResponse(w, http.StatusBadRequest, errors.Wrap(err, "while getting query parameters"))
 		return
 	}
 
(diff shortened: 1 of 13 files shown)


components/kyma-environment-broker/internal/runtime/handler.go
```diff
@@ -99,39 +99,42 @@ func (h *Handler) getOperationsForInstance(instance internal.Instance) (*interna
...
-	var cursor string
+	var page int
 	var err error
 
 	params := req.URL.Query()
-	limitArr, ok := params[limitParam]
+	limitArr, ok := params[pageSizeParam]
 	if len(limitArr) > 1 {
-		return 0, "", errors.New("limit has to be one parameter")
+		return 0, 0, errors.New("limit has to be one parameter")
```
polskikiel, line 110: please change limit to page_size in the below error messages

components/kyma-environment-broker/internal/storage/driver/memory/instance.go
```diff
@@ -160,34 +158,25 @@ func (s *Instance) GetInstanceStats() (internal.InstanceStats, error) {
...
 	var toReturn []internal.Instance
-
-	offset, err := pagination.DecodeOffsetCursor(cursor)
-	if err != nil {
-		return nil, nil, 0, err
+	var offset int
+	if page < 2 {
+		offset = 0
+	} else {
+		offset = page*pageSize - 1
```
polskikiel, line 169: maybe an overhead but we have similar logic in the `ConvertPageLimitAndOrderedColumnToSQL` function, we could export it to e.g. separate function

polskikiel (approved):

PR #245, opened 2020-10-01
Fix field type

polskikiel:
**Description**

- Fixed field type

it fails on dev with the following message:

```
"while reading Operation from the storage: Failed to get operation: sql: Scan error on column index 10, name \"orchestration_id\": converting NULL to string is unsupported
```

proposed solution: http://devtidbits.com/2020/08/03/go-sql-error-converting-null-to-string-is-unsupported/ 

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/storage/driver/postsql/operation.go
+++ b/components/kyma-environment-broker/internal/storage/driver/postsql/operation.go
@@ -1,6 +1,7 @@
 package postsql
 
 import (
+	"database/sql"
 	"encoding/json"
 	"fmt"
 	"time"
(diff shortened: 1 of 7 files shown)


kjaksik (approved):
lgtm

components/kyma-environment-broker/internal/storage/driver/postsql/operation.go
```diff
@@ -551,5 +555,25 @@ func operationToDB(op *internal.Operation) dbmodel.OperationDTO {
...
+
+	return result
+}
+
+func orchestrationIDtoString(orchestrationID sql.NullString) string {
+	if orchestrationID.Valid {
+		return orchestrationID.String
+	}
+
+	return ""
```
kjaksik, line 578: What do you think about moving those functions to a common plane and calling them something like `StringToSQLNullString` and `SQLNullStringToString`? I think this logic is not exclusive to orchestrationID and can be reused

kjaksik (approved):
LGTM

PR #248, opened 2020-10-05
Exclude internal secrets from being cleaned up

Maladie:
**Description**

- HAP now checks if secret is labeled as `internal` before marking it as `dirty`. This is because some secrets (for example dedicated e2e test secret) should not be returned to the pool despite being unused.

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/common/hyperscaler/account_provider_test.go
+++ b/components/kyma-environment-broker/common/hyperscaler/account_provider_test.go
@@ -36,6 +36,23 @@ func TestMarkUnusedGardenerSecretAsDirty(t *testing.T) {
 		assert.Equal(t, secret.Labels["dirty"], "true")
 	})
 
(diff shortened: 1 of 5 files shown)


components/kyma-environment-broker/common/hyperscaler/account_provider_test.go
```diff
@@ -36,6 +36,23 @@ func TestMarkUnusedGardenerSecretAsDirty(t *testing.T) {
 		assert.Equal(t, secret.Labels["dirty"], "true")
 	})
 
+	t.Run("should not mark secret as dirty if internal", func(t *testing.T) {
+		//given
```
piotrmiskiewicz, line 40: remove one "given"

Maladie:
duplicate #https://github.com/kyma-project/control-plane/pull/251

Issue #193, opened 2020-09-11
HTTP client error from sending request treat as a temporary


jasiu001:
**Description**

Many provisioning steps in KEB use HTTP client (EDP, IAS etc), usually clients use temporary error when e.g. response code is 500 or higher. A temporary error results in a repeat of the steppes. 
A step is not repeated if execute `Do` method of HTTP client returns an error (the problem may be caused by a temporary lack of networks). 
All clients should return a temporary error if `Do` method returns error.

**Reasons**

A temporary lack of networks should not result in a permanent error.

**AC**
- All clients (EDP, IAS, LMS...) should return a [temporary error](https://github.com/kyma-project/control-plane/blob/master/components/kyma-environment-broker/internal/ias/client.go#L247) when method `Do` returns [error](https://github.com/kyma-project/control-plane/blob/master/components/kyma-environment-broker/internal/ias/client.go#L236)
- Every step which uses the HTTP client should handle [temporary error](https://github.com/kyma-project/control-plane/blob/master/components/kyma-environment-broker/internal/process/provisioning/ias_registration.go#L92)
- If the client use `Post`, `Get` [method](https://github.com/kyma-project/control-plane/blob/master/components/kyma-environment-broker/internal/edp/client.go#L139), the method should be split for two `createRequest` (if an error occur in this method then the error should be permanent) and `Do`

PR #258, opened 2020-10-08
Upgrade HTTP clients - treat connection error as a temporary

jasiu001:
**Description**
PR changes HTTP clients for external services bahavior. Now steps are repeated if the execution of `Do()` method from HTTP client returns an error.

The provisioning process consists of 13 steps of which 7 of them uses HTTP client
- _provideLmsTenantStep_, _lmsCertStep_,  _InternalEvaluationStep_ - error from client treat as temporary already (all errors, no matter what kind of error), handling of temporary errors has been added, other errors are repeated for a shorter period of time
- _EDPRegistrationStep_
- _IASRegistrationStep_
- _CreateRuntimeStep_ - already support temporary error for network connection
- _InitialisationStep_ - already support temporary error for network connection

**Related issue(s)**
[#193](https://github.com/kyma-project/control-plane/issues/193)

### DIFF
--- a/components/kyma-environment-broker/internal/avs/client.go
+++ b/components/kyma-environment-broker/internal/avs/client.go
@@ -170,6 +170,10 @@ func (c *Client) execute(request *http.Request, allowNotFound bool, allowResetTo
 		return &http.Response{}, kebError.AsTemporaryError(err, "while executing request by http client")
 	}
 
+	if response.StatusCode >= http.StatusInternalServerError {
+		return response, kebError.NewTemporaryError("avs server returned %d status code", response.StatusCode)
+	}
+
 	switch response.StatusCode {
 	case http.StatusOK, http.StatusCreated:
 		return response, nil
@@ -186,10 +190,7 @@ func (c *Client) execute(request *http.Request, allowNotFound bool, allowResetTo
 			return c.execute(request, allowNotFound, false)
 		}
 		return response, fmt.Errorf("avs server returned %d status code twice for %s (response body: %s)", http.StatusUnauthorized, request.URL.String(), responseBody(response))
-	case http.StatusInternalServerError,
-		http.StatusRequestTimeout,
-		http.StatusServiceUnavailable,
-		http.StatusGatewayTimeout:
+	case http.StatusRequestTimeout:
 		return response, kebError.NewTemporaryError("avs server returned %d status code", response.StatusCode)
 	default:
 		return response, fmt.Errorf("unsupported status code: %d for %s (response body: %s)", response.StatusCode, request.URL.String(), responseBody(response))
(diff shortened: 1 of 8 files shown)


components/kyma-environment-broker/internal/avs/client.go
```diff
@@ -186,6 +186,11 @@ func (c *Client) execute(request *http.Request, allowNotFound bool, allowResetTo
 			return c.execute(request, allowNotFound, false)
 		}
 		return response, fmt.Errorf("avs server returned %d status code twice for %s (response body: %s)", http.StatusUnauthorized, request.URL.String(), responseBody(response))
+	case http.StatusInternalServerError,
```
piotrmiskiewicz, line 189: I think all http errors with code grater or equal to 500 should be handled this way

components/kyma-environment-broker/internal/edp/client.go
```diff
@@ -111,40 +119,65 @@ func (c *Client) DeleteMetadataTenant(name, env, key string) error {
...
-		return []MetadataItem{}, errors.Wrap(err, "while requesting about dataTenant metadata")
+		return metadata, errors.Wrap(err, "while creating GET metadata tenant request")
+	}
+	response, err := c.httpClient.Do(request)
+	if err != nil {
+		return metadata, kebError.AsTemporaryError(err, "while requesting about dataTenant metadata")
 	}
-	defer func() {
-		err = multierror.Append(err, errors.Wrap(c.closeResponseBody(response), "while trying to close body reader")).ErrorOrNil()
-	}()
```
piotrmiskiewicz, line 127: removing a `defer` may cause not closing the response, for examle when there is a response with broken JSON

jasiu001: I removed `defer` but in each client before `return` and after handle the response, I close `response.Body` treat error from this operation as a temporary, e.g.:
```go
if err := c.closeResponseBody(response); err != nil {
    return kebError.AsTemporaryError(err, "while closing delete MetadataTenant response")
}
```
it is because when `multierror` is used we lose the temporary error kind, e.g. in this solution temporary case will not work:
```go
func main() {
	err := test()

	if err != nil {
		switch {
		case IsTemporaryError(err):
			fmt.Println("[TEMP] error ", err)
		default:
			fmt.Println(err)
		}
	}
}

func test() (err error) {
	err = test2()
	defer func() {
		err = multierror.Append(err, errors.Wrap(closeResponseBody(), "while trying to close body reader")).ErrorOrNil()
	}()
	if err != nil {
		return errors.Wrap(err, "while test")
	}
	return nil
}

func test2() (err error) {
	err = test3()
	if err != nil {
		return AsTemporaryError(err, "while test3")
	}
	return nil
}

func test3() error {
	return fmt.Errorf("something bad")
}

func closeResponseBody() error {
	return fmt.Errorf("cannot close")
}
```

piotrmiskiewicz (approved):

PR #266, opened 2020-10-12
Implement strategies in KEB

polskikiel:
**Description**

- Implemented strategies in KEB

**Related issue(s)**


components/kyma-environment-broker/internal/orchestration/handlers/kyma_handler.go
```diff
@@ -206,3 +210,28 @@ func (h *kymaHandler) resolveErrorStatus(err error) int {
 		return http.StatusInternalServerError
 	}
 }
+
+func (h *kymaHandler) defaultTarget(spec *internal.TargetSpec) {
```
ebensom, line 214: I suggest we make at least one include target mandatory. Defaulting to target all can lead to major accidents.

components/kyma-environment-broker/internal/orchestration/strategy_parallel.go
```diff
@@ -11,42 +10,61 @@ import (
...
 		})
 	}
 
 	for _, op := range operations {
-		if isMaintenanceWindowMode {
-			time.Sleep(time.Until(op.MaintenanceWindowEnd))
+		if isMaintenanceWindowSchedule {
+			until := time.Until(op.MaintenanceWindowBegin)
+			p.log.Infof("Waiting %v to start upgrade operation %s", until, op.Operation.ID)
+			time.Sleep(until)
```
ebensom, line 49: This logic doesn't guarantee that all operations will be executed within the requested time window. 
Think of an example where typical operation takes 30 minutes. Executing 10 operations with 2 workers would take ~150 minutes. If time window is is the same 1 hour for each runtime, tbis logic executes beyond the time frame.

ebensom: I have an idea how to implement scheduling properly by relying on the underlying `DelayedQueue`'s `AddAfter()` method. PLease ping me to discuss.

components/kyma-environment-broker/internal/orchestration/kyma/manager.go
```diff
@@ -120,9 +122,10 @@ func (u *upgradeKymaManager) resolveOperations(o *internal.Orchestration, params
 					SubAccountID:           r.SubAccountID,
 					OrchestrationID:        o.OrchestrationID,
 				},
+				ProvisioningParameters: po.ProvisioningParameters,
```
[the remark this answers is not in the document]
polskikiel: I save it to use it later in the [converter](https://github.com/kyma-project/control-plane/blob/master/components/kyma-environment-broker/internal/orchestration/handlers/converter.go#L43)
that prevents the decoding error which happens when we try to fetch data about a single operation while it didn't reach the call to the provisioner. In the converter, we can also just omit `planName` and `planID` fields in converter if we get an error on decoding the provisioning parameters WDYT?

jasiu001 (approved):

jasiu001 (approved):

jasiu001 (approved):

PR #277, opened 2020-10-19
Fix problem with failing deprovisioning when Director returns 'Tenant not found' error

Maladie:
**Description**

- Internal app error code added
- provisioner no longer returns error when director returns 'Tenant not found' error

**Related issue(s)**


components/provisioner/internal/apperrors/apperrors.go
```diff
@@ -9,42 +9,58 @@ const (
...
+	IntCodeForbidden      IntErrCode = 12
+	IntCodeBadRequest     IntErrCode = 13
+	IntCodeTenantNotFound IntErrCode = 13
+)
+
 type ErrCode int
 
+type IntErrCode int
+
 type AppError interface {
```
akgalwas, line 24: We could make the solution more readable by applying the following change:
```
type AppError interface {
        Append(string, ...interface{}) AppError
	Code() ErrCode
	Cause() CauseCode
	Error() string
}
```
If we use "Cause" term the meaning of the additional error code gets obvious (Internal Code is enigmatic a bit).

components/provisioner/internal/apperrors/apperrors.go
```diff
@@ -9,42 +9,58 @@ const (
...
-	code    ErrCode
-	message string
+	code         ErrCode
+	internalCode IntErrCode
+	message      string
 }
 
-func errorf(code ErrCode, format string, a ...interface{}) AppError {
-	return appError{code: code, message: fmt.Sprintf(format, a...)}
+func errorf(code ErrCode, internalCode IntErrCode, format string, a ...interface{}) AppError {
```
akgalwas, line 37: When we introduce cause field with the following contants:
- Unknown
- TenantNotFound

We will avoid duplicating information as in the following code:
`return errorf(CodeForbidden, IntCodeForbidden, format, a...)`

By default cause field will have Unknown value.

components/provisioner/internal/apperrors/apperrors.go
```diff
@@ -54,3 +70,7 @@ func (ae appError) Code() ErrCode {
 func (ae appError) Error() string {
 	return ae.message
 }
+
+func (ae appError) IntCode() IntErrCode {
```
akgalwas, line 74: Rename to Cause() or similar.

components/provisioner/internal/apperrors/apperrors.go
```diff
@@ -9,42 +9,58 @@ const (
...
-	return errorf(CodeForbidden, format, a...)
+	return errorf(CodeForbidden, IntCodeForbidden, format, a...)
 }
 
 func BadRequest(format string, a ...interface{}) AppError {
-	return errorf(CodeBadRequest, format, a...)
+	return errorf(CodeBadRequest, IntCodeBadRequest, format, a...)
+}
+
+func WrongTenant(format string, a ...interface{}) AppError {
```
akgalwas, line 57: InvalidTenant instead of WrongTenant?

components/provisioner/internal/director/directorclient.go
```diff
@@ -171,6 +171,9 @@ func (cc *directorClient) RuntimeExists(id, tenant string) (bool, apperrors.AppE
 	var response GetRuntimeResponse
 	err := cc.executeDirectorGraphQLCall(runtimeQuery, tenant, &response)
 	if err != nil {
+		if err.IntCode() == apperrors.IntCodeTenantNotFound {
```
akgalwas, line 174: If we change IntCode to Cause the code should look as follows:
```
if err.Code() == apperrors.CodeBadRequest && err.Cause() == apperrors.CodeTenantNotFound {
   ....
} 
```

akgalwas (approved):

PR #282, opened 2020-10-20
Expose kyma upgrade operation state in API

ebensom:
**Description**

Expose kyma upgrade operation state in `/orchestrations/{orchestration_id}/operations`


polskikiel (approved):
lgtm, you can also bump the image

polskikiel (approved):

PR #295, opened 2020-10-28
Add retries to Director calls

Maladie:
**Description**

- Retries added
- Tests added

**Related issue(s)**


components/provisioner/internal/util/retry.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"time"
+
+	"github.com/kyma-project/control-plane/components/provisioner/internal/apperrors"
+
+	"github.com/sirupsen/logrus"
+)
+
+func RetryOnError(interval time.Duration, count int, errMsgFmt string, function func() apperrors.AppError) apperrors.AppError {
```
akgalwas, line 11: There is a numer of libraries that implement retries (e.g. github.com/avast/retry-go,  k8s.io/apimachinery/pkg/util/wait).What is the reason we implemented this function from scratch?

akgalwas, line 11: I would add tests for RetryOnError function.

akgalwas (approved):

components/provisioner/internal/util/retry.go
Maladie: We want to return apperrors.AppError from Director calls

akgalwas (approved):

PR #300, opened 2020-10-30
Fix k8s.io/apiserver in Provisioner

rafalpotempa:
**Description**

- Fix k8s.io/apiserver in Provisioner


koala7659 (approved):
Seems to be OK. I'm worried a little bit about old version of libraries listed in go.sum file.

PR #304, opened 2020-11-03
Add possibility to pass seed specific audit log tenant

Maladie:
**Description**

- Shoot controller will now search for seed specific audit log tenant before checking generic one

**Related issue(s)**

### DIFF
--- a/components/provisioner/internal/gardener/shoot_reconciler.go
+++ b/components/provisioner/internal/gardener/shoot_reconciler.go
@@ -153,10 +153,9 @@ func getAuditLogTenant(seed string, data map[string]string) string {
 }
 
 func findTenantStrictly(seed string, data map[string]string) string {
-	for key, tenant := range data {
-		if seed == key {
-			return tenant
-		}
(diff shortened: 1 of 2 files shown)


components/provisioner/internal/gardener/shoot_reconciler.go
```diff
@@ -143,13 +140,34 @@ func (r *Reconciler) enableAuditLogs(logger logrus.FieldLogger, namespacedName t
...
+	return findTenantLoosely(seed, data)
+}
+
+func findTenantStrictly(seed string, data map[string]string) string {
+	for key, tenant := range data {
+		if seed == key {
+			return tenant
+		}
+	}
+	return ""
```
majakurcius, line 161: No need to iterate over the map:

```suggestion
	tenant, found := data[seed]
	if found {
		return tenant
	}
	return ""
```

Maladie: Right, thanks

koala7659 (approved):
LGTM

PR #328, opened 2020-11-13
fixing issue with trial cluster

maksd:
**Description**

- fix issue with trial clusters not reporting metrics because they have no event hub resource groups.


k15r (approved):
lgtm

PR #330, opened 2020-11-15
Configure evaluation purpose for trial runtimes

ebensom:
**Description**

- Default shoot purpose applies only if not set by provider specific defaults
- Set evaluation purpose as default for trial plan


piotrmiskiewicz (approved):
lgtm

PR #334, opened 2020-11-16
Add random suffix to trial runtimes

kjaksik:
**Description**

- Add random suffix to trial runtimes

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/process/input/input.go
+++ b/components/kyma-environment-broker/internal/process/input/input.go
@@ -3,6 +3,10 @@ package input
 import (
 	"time"
 
+	"github.com/Masterminds/goutils"
+
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/broker"
+
 	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal"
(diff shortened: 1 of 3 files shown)


components/kyma-environment-broker/internal/process/input/input.go
```diff
@@ -268,6 +278,15 @@ func (r *RuntimeInput) applyGlobalOverridesForUpgradeRuntime() error {
 	return nil
 }
 
+func (r *RuntimeInput) addRandomStringToTrialRuntimeName() error {
+	rand.Seed(time.Now().UnixNano())
+	if r.provisioningParameters.PlanID == broker.TrialPlanID {
```
piotrmiskiewicz, line 283: there is an IsTrial func in the broker package

piotrmiskiewicz: Why the random part is used only for trial? are there any business reasons?

kjaksik: User cannot specify the name when provisioning a trial skr. A random name comes from CP and there were occurences when the name was repeated. We want to make sure that the name is random and it won't cause any problems in the Director DB

piotrmiskiewicz (approved):

PR #344, opened 2020-11-23
Make orchestration states external

ebensom:
**Description**

Needed to move KCP CLI out from `components/kyma-environment-broker` to `components/cli`

**Related issue(s)**


polskikiel (approved):
lgtm

PR #346, opened 2020-11-24
Service Manager client

piotrmiskiewicz:
**Description**

- implement Service Manager client for necessary methods
- provide an easy way to get SM client from provisioning step
- implement dummy "doing readonly" uaa step (this implementation must be changed in the next PRs) to proove it works fine
- change sm_overrides step to use new implementation of SM credentails logic

**Related issue(s)**
Related: https://github.tools.sap/kyma/backlog/issues/908

### DIFF
--- a/components/kyma-environment-broker/internal/servicemanager/client.go
+++ b/components/kyma-environment-broker/internal/servicemanager/client.go
@@ -1,0 +1,265 @@
+package servicemanager
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/Peripli/service-manager-cli/pkg/query"
+	"github.com/Peripli/service-manager-cli/pkg/types"
+	"github.com/Peripli/service-manager/pkg/web"
+	kebError "github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/error"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/iosafety"
+	"github.com/pkg/errors"
+)
+
+type (
+	client struct {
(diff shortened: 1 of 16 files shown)


components/kyma-environment-broker/internal/servicemanager/client.go
```diff
@@ -0,0 +1,243 @@
...
+	client struct {
+		creds      Credentials
+		httpClient *http.Client
+	}
+)
+
+func New(credentials Credentials) Client {
+	return &client{
+		creds:      credentials.WithNormalizedURL(),
+		httpClient: http.DefaultClient,
```
crabtree, line 29: I think it is worth to configure at least some timeouts on the HTTP client.

crabtree (changes requested):

components/kyma-environment-broker/internal/servicemanager/client.go
```diff
@@ -0,0 +1,243 @@
...
+	req, err := c.createHTTPRequest(method, url, body)
+	if err != nil {
+		return 0, err
+	}
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return 0, err
+	}
+	defer func() {
```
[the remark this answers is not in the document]
piotrmiskiewicz: left by accident, fixed

components/kyma-environment-broker/internal/servicemanager/client.go
piotrmiskiewicz: good point, let's do a new constructor with http.Client as an argument and the factory, which creates sm client has one shared and configured http.Client

crabtree (approved):

crabtree (approved):

PR #347, opened 2020-11-24
Save console URL from provisioner gardenerClusterDomain label

ebensom:
**Description**

Currently KEB only saves the dashboard URL in instance state after the provisioning has finished successfully, using the `runtime_ConsoleUrl` label set by compass-runtime-agent in director.

The `/runtimes` API, and the `kcp runtimes` and `kcp kubeconfig` CLI commands rely on the dashboard URL to display the shoot name, or to fetch kubeconfig by given shoot. 
As a result, only the successfully provisioned runtimes have their corresponding shoot names displayed. kubeconfig cannot be downloaded by shoot name for failed clusters. 
 
- KEB gets console URL by `gardenerClusterDomain` label set on the runtime in Director. This label is [set by provisioner](https://github.com/kyma-project/control-plane/blob/master/components/provisioner/internal/operations/stages/provisioning/wait_for_cluster_domain.go#L99) as the first step of the provisioning operation
- Query and save the dashboard URL in `Provision_Initialization` step even during the provisioner operation is in progress.


ebensom:
Closing in favor of a more general solution implemented in the near future.

PR #351, opened 2020-11-25
Get Kyma version per GA for upgrade

ebensom:
**Description**

Consider the `kyma-versions` `ConfigMap` containing the Kyma versions per global account during upgrade operation


piotrmiskiewicz (approved):
lgtm

PR #352, opened 2020-11-26
Replace latest-published source flag with master source flag

shorim:
**Description**

- `latest-published` source flag will no longer exist for the CLI. So, its usage is replaced with the new `master` source flag

**Related issue(s)**

kyma-project/cli#501

### DIFF
--- a/installation/cmd/run.sh
+++ b/installation/cmd/run.sh
@@ -13,7 +13,7 @@ if [[ $KYMA_RELEASE == *PR-* ]]; then
   if [ -z "$KYMA_TAG" ]; then echo "ERROR: Kyma artifacts for ${KYMA_RELEASE} not found."; exit 1; fi
   KYMA_SOURCE="eu.gcr.io/kyma-project/kyma-installer:${KYMA_TAG}"
 elif [[ $KYMA_RELEASE == master ]]; then
-  KYMA_SOURCE="latest-published"
+  KYMA_SOURCE="master"
 elif [[ $KYMA_RELEASE == *master-* ]]; then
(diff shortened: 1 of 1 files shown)


installation/cmd/run.sh
```diff
@@ -13,7 +13,7 @@ if [[ $KYMA_RELEASE == *PR-* ]]; then
   if [ -z "$KYMA_TAG" ]; then echo "ERROR: Kyma artifacts for ${KYMA_RELEASE} not found."; exit 1; fi
   KYMA_SOURCE="eu.gcr.io/kyma-project/kyma-installer:${KYMA_TAG}"
 elif [[ $KYMA_RELEASE == master ]]; then
-  KYMA_SOURCE="latest-published"
+  KYMA_SOURCE="master"
```
janmedrek, line 16: Why are we changing this? I do not think we want the installer to deploy the `master` version as it is not recognized as a stable one. Or is the `latest-published` related to the last commit on the master branch? I would need more insights on what is the goal here and how it will affect the process of installation (what will be installed on a cluster).

suleymanakbas91: It's just a renaming from `latest-published` to `master` as `master` is more obvious for the users.

suleymanakbas91: Just to clarify, this is the Kyma version that CLI is installing. So, it has nothing to do with `stable` binary for CLI. `latest-published` was installing the latest master from Kyma, so we renamed it to make it more obvious.

janmedrek (approved):

Issue #261, opened 2020-10-09
KEB: allow to cancel given runtime upgrade orchestration


PK85:
**Description**

In case of mistake or upgrades issues we need to allow operator to cancel given orchestration.

**AC**
- add documentation
- update swagger
- if operation is still not taken is should have status "pending"
- implement functionality where all "pending" operations are chnaged to "canceled"
- set orchestration status to "cancel"
- do not cancel "in progress" operations

PR #355, opened 2020-11-27
Allow canceling in progress orchestrations

polskikiel:
**Description**

- Added `Canceling` and `Canceled` state
- Added an endpoint for canceling PUT /orchestrations/:id/cancel
   - It just set the orchestration state to `Canceling`, which is handled through different services - orchestrationKymaManager, strategy, and kymaUpgradeManager.
- UpgradeKymaOperations now uses extended orchestration states with included Canceling, Canceled, and Pending states. They are compatible with `domain.LastOperationState` states that we are using for provisioning and deprovisioning.
- Refactor kyma_handler to two separate structs - orchestration_handler and kyma_handler
- adjusted docs and swagger  

**Related issue(s)**

https://github.com/kyma-project/control-plane/issues/261


components/kyma-environment-broker/common/orchestration/strategies/parallel.go
```diff
@@ -0,0 +1,174 @@
+package strategies
```
polskikiel, line 1: diff in /strategies files is misleading because I moved them to a separate package. I've changed mostly processOperation function. To see the difference, look at the deleted file.

components/kyma-environment-broker/internal/orchestration/handlers/cancel.go
```diff
@@ -0,0 +1,59 @@
...
+	return &Canceler{
+		orchestrations: orchestrations,
+		operations:     operations,
+		log:            logger,
+	}
+}
+
+// Cancel finds in progress orchestration and cancels it
+func (c *Canceler) Cancel() (string, error) {
+	orchestrations, _, _, err := c.orchestrations.List(dbmodel.OrchestrationFilter{States: []string{orchestrationExt.InProgress}})
```
ksputo, line 28:
```suggestion
	orchestrations, _, size, err := c.orchestrations.List(dbmodel.OrchestrationFilter{States: []string{orchestrationExt.InProgress}})
```
Can we use returned parameter here? We already return size of the returned Orcherstration list

components/kyma-environment-broker/internal/orchestration/handlers/cancel.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "while listing orchestrations")
+	}
+	size := len(orchestrations)
```
ksputo, line 32:
```suggestion
```

components/kyma-environment-broker/internal/orchestration/handlers/cancel_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	fixOrchestrationID = "test-id"
+)
+
+func TestNewCanceler(t *testing.T) {
```
ksputo, line 19: You don't test NewCanceler constructor in this code block. Consider changing this to more descriptive approach, like:
```golang
func TestCanceler_Cancel(t *testing.T) {
	t.Run("should cancel existing orchestration" , func(t *testing.T) {
		/* 
		...
		 */
	})
	t.Run("should return error when orchestration not found" , func(t *testing.T) {
		/*
			...
		*/
	})
}

func TestCanceler_CancelForID(t *testing.T) {
	t.Run("should cancel orchestration" , func(t *testing.T) {
		/*
			...
		*/
	})
	t.Run("should return error when orchestration not found" , func(t *testing.T) {
		/*
			...
		*/
	})
}
```

components/kyma-environment-broker/common/orchestration/strategies/parallel.go
```diff
@@ -0,0 +1,179 @@
...
+			return false, nil
+		}
+		if o.State == orchestration.Canceled {
+			isCanceled = true
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil && err != wait.ErrWaitTimeout {
+		return false, errors.Wrap(lastErr, "while pooling operations")
```
ksputo, line 176:
```suggestion
		return false, errors.Wrap(lastErr, "while polling operations")
```

components/kyma-environment-broker/internal/process/upgrade_kyma_operation.go
```diff
@@ -4,8 +4,11 @@ import (
 	"errors"
 	"time"
 
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/common/orchestration"
+
```
ksputo, line 8:
```suggestion
```
unnecessary newline

components/kyma-environment-broker/internal/process/upgrade_kyma_operation.go
```diff
@@ -40,6 +43,16 @@ func (om *UpgradeKymaOperationManager) OperationFailed(operation internal.Upgrad
 	return updatedOperation, 0, errors.New(description)
 }
 
+// OperationSucceeded marks the operation as succeeded and only repeats it if there is a storage error
+func (om *UpgradeKymaOperationManager) OperationCancelled(operation internal.UpgradeKymaOperation, description string) (internal.UpgradeKymaOperation, time.Duration, error) {
```
ksputo, line 47:
```suggestion
func (om *UpgradeKymaOperationManager) OperationCanceled(operation internal.UpgradeKymaOperation, description string) (internal.UpgradeKymaOperation, time.Duration, error) {
```

let's decide on one spelling for cohesion:
```
canceled - American English
cancelled - British English
```

resources/kcp/charts/kyma-environment-broker/files/swagger.yaml
```diff
@@ -58,6 +58,24 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/StatusResponseList'
+    delete:
+      summary: Cancels in progress orchestration
+      operationId: cancelDynamic
+      description: |
+        Cancels in progress orchestration depends on their state from storage
```
ksputo, line 65: how about
```
Cancels all orchestration in progress
```

resources/kcp/charts/kyma-environment-broker/files/swagger.yaml
```diff
@@ -85,6 +103,31 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/errObj'
+    delete:
+      summary: Cancels in progress orchestration
+      operationId: cancelByID
+      description: |
+        Cancels in progress orchestration with declared ID
```
ksputo, line 110: how about
```
Cancels given orchestration in progress
```

ksputo (commented):
some minor comments

components/kyma-environment-broker/internal/process/upgrade_kyma_operation.go
polskikiel: I wasnt decided at first but lets go with the American version

resources/kcp/charts/kyma-environment-broker/files/swagger.yaml
polskikiel: I will go with Cancels in progress orchestration

ksputo (approved):

components/kyma-environment-broker/common/orchestration/strategies/parallel.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+type Executor interface {
+	Execute(operationID string) (time.Duration, error)
+}
+
+type ParallelOrchestrationStrategy struct {
+	executor        Executor
+	dq              workqueue.DelayingInterface
+	orchestrations  storage.Orchestrations
```
ebensom, line 28: Strategy must be independent from storage. I'm about to raise the taskrun PR in CLI which leverages strategy and resolver. Please move all storage operations (cancellation checks) to the operation's Execute() or one of it's steps.

ebensom (changes requested):

ebensom:
Please do not merge this PR, strategy must not be dependent on storage.

components/kyma-environment-broker/internal/process/upgrade_kyma/initialisation.go
```diff
@@ -59,6 +64,37 @@ func (s *InitialisationStep) Name() string {
...
+
+		op, err := s.operationStorage.UpdateUpgradeKymaOperation(operation)
+		if err != nil {
+			log.Errorf("while updating operation: %v", err)
+			return operation, s.timeSchedule.Retry, nil
+		}
+		operation = *op
+	}
+
+	// if schedule is maintenanceWindow and time window for this operation has finished we reprocess on next time window
```
ebensom, line 93: rescheduling logic should not be added back here

components/kyma-environment-broker/internal/process/upgrade_kyma/initialisation.go
```diff
@@ -59,6 +64,37 @@ func (s *InitialisationStep) Name() string {
...
+	o, err := s.orchestrationStorage.GetByID(operation.OrchestrationID)
+	if err != nil {
+		if dberr.IsNotFound(err) {
+			log.Errorf("orchestration %s not found: %v", err)
+			return s.operationManager.OperationFailed(operation, fmt.Sprintf("orchestration %s not found", operation.OrchestrationID))
+		}
+		log.Errorf("while getting orchestration from storage: %v", err)
+		return operation, s.timeSchedule.Retry, nil
+	}
+	if o.State == orchestrationExt.Canceled {
```
ebensom, line 76: Why do we need to check canceled orchestration state if canceled state is propagated down to operations?

components/kyma-environment-broker/common/orchestration/strategies/parallel.go
polskikiel: I think we need to check somehow if orchestration wasn't canceled in case of `maintenanceWindow` schedule type. After adding it to the queue I'm not able to remove it so I decided to go with the polling storage function. Instead of using storage in strategies, can we just send a pointer to this waiter function and use different waiting strategies if it's provided or not? It would look like that: 
```
	if p.isCanceled != nil {
		waiter := *p.isCanceled
		isCanceled, err := waiter(until, log)
		if err != nil {
			return errors.Wrap(err, "while checking if orchestration was canceled")
		}
		if isCanceled {
			log.Info("Operation was canceled")
			return nil
		}
		log.Infof("Upgrade operation is scheduled now")
		p.dq.Add(id)
	} else {
		p.dq.AddAfter(id, until)
	}
```

ebensom: So in case of maintenancewindow the workers are blocked at `dq.Get()` . These internally used delaying queues cound only be cancelled with `Shutdown()` call which will signal and wake all blocked worker goroutines. So all you need is to store dq in strategy, and expose dq.Shutdown() call in strategy.Cancel() function.

polskikiel: Didn't know that delaying queues have that fancy shutdown logic, exactly what I needed there thx!

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -471,6 +474,29 @@ func processOrchestration(state string, op storage.Orchestrations, queue *proces
 	return nil
 }
 
+// processCanceledOrchestration reprocess canceled orchestrations only when some in progress operations exists
+func processCanceledOrchestration(orchestrationsStorage storage.Orchestrations, operationsStorage storage.Operations, queue *process.Queue, log logrus.FieldLogger) error {
```
ebensom, line 478: Should we reprocess all cancelled orchestrations every time KEB is started? I know there is a corner case when KEB is restarted/crashes during cancellation, but I'm thinking we should only reprocess the last one, as ne number of cancelled orchestrations could be huge over time.

components/kyma-environment-broker/common/orchestration/strategies/parallel.go
```diff
@@ -16,42 +17,43 @@ type Executor interface {
 
 type ParallelOrchestrationStrategy struct {
 	executor Executor
-	log      logrus.FieldLogger
+	dq       workqueue.DelayingInterface
```
ebensom, line 20: We should keep `dq` local to execution, otherwise if there were concurrent executions launched by the same strategy ovject, operations would be mixed up in the shared dq. I know right now each orchestration creates a separate strategy object, but the strategy was designed to be thread safe and this change breaks that. 

You could do similar logic which is done with execution specific waitgroup: save it in a map (access protected by the mutex), and when `Cancel(execID)` is called, look up the `dq` based in `execID` and call `Shutdown()` on looked-up object.

components/kyma-environment-broker/internal/orchestration/kyma/manager.go
```diff
@@ -150,12 +152,18 @@ func (u *upgradeKymaManager) resolveOperations(o *internal.Orchestration, params
...
 	} else {
-		// Resume processing of in progress upgrade operations after restart
+		// Resume processing of not finished upgrade operations after restart
 		var err error
-		result, _, _, err = u.operationStorage.ListUpgradeKymaOperationsByOrchestrationID(o.OrchestrationID, dbmodel.OperationFilter{States: []string{string(domain.InProgress)}})
+		inProgress, _, _, err := u.operationStorage.ListUpgradeKymaOperationsByOrchestrationID(o.OrchestrationID, dbmodel.OperationFilter{States: []string{orchestration.InProgress}})
 		if err != nil {
 			return result, err
 		}
+		pending, _, _, err := u.operationStorage.ListUpgradeKymaOperationsByOrchestrationID(o.OrchestrationID, dbmodel.OperationFilter{States: []string{orchestration.Pending}})
```
ebensom, line 161: You could merge these two list with passing `States: []string{orchestration.Pending, orchestration.InProgress}`

components/kyma-environment-broker/internal/orchestration/kyma/manager.go
```diff
@@ -200,37 +208,84 @@ func (u *upgradeKymaManager) updateOrchestration(o *internal.Orchestration, stat
...
-		return errors.Wrap(err, "while waiting for scheduled operations to finish")
+		return nil, errors.Wrap(err, "while waiting for scheduled operations to finish")
 	}
 
-	orchestrationState := orchestration.Succeeded
-	if stats[domain.Failed] > 0 {
-		orchestrationState = orchestration.Failed
+	if o.State == orchestration.Canceled {
+		strategy.Cancel(execID)
+		err = u.resolveCanceledOperations(o)
```
ebensom, line 262: I think the correct order is to first update cancelled state for each operation in DB, then signal `strategy.Cancel()`

components/kyma-environment-broker/internal/orchestration/kyma/manager.go
polskikiel: OK, I used a single list as you suggested and then I split operations per pending and in-progress states in the separate logic because I want to reprocess already in-progress operations at first to keep the order of executing operations

ebensom: Ah, I you need to process them separately, you could also keep the two queries, it shouldn't be too much of a difference performance wise.

components/kyma-environment-broker/cmd/broker/main.go
polskikiel: We can implement it to check only the latest operation to decrease the complexity but we need to keep in mind that we might have canceled orchestration with unfinished operations in our storage. I'd prefer to have that mechanism reliable to not have situations like with never-ending provisioning/deprovisioning operations but as you wish :)

polskikiel: Added `return nil` after we find the first matching canceled orchestration to reprocess to skip the others

components/kyma-environment-broker/common/orchestration/strategies/parallel.go
polskikiel: ok I'm saving now delaying queues into the map and I'm using them in the context of executionID, it should be thread-safe now again

components/kyma-environment-broker/internal/orchestration/kyma/manager.go
polskikiel: it's quite readable so I'll leave it as it's now, also one query less is always better

ebensom (dismissed):

ksputo (dismissed):

docs/kyma-environment-broker/03-10-orchestration.md
```diff
@@ -42,3 +43,11 @@ The example strategy configuration looks as follows:
...
 }
 ```
+
+## Cancellation
+
+You may cancel any in progress orchestration using one of two DELETE endpoints described above.
+
+After you cancel an orchestration, the KEB sets its state to `Canceled`. Orchestration with such state is not scheduling any new operations. Every pending operation in case of cancel are also marked with `Canceled` state.
+
+To provide consistency, cancelled orchestration waits for already processed operations to finish. When operations are finished the orchestration queue processing becomes unblocked.
```
klaudiagrz, line 53:
```suggestion
To provide consistency, a canceled orchestration waits for already processed operations to finish. When operations are finished, the cancelation is processed.
```

docs/kyma-environment-broker/03-10-orchestration.md
```diff
@@ -12,7 +12,8 @@ If Kyma Environment Broker is restarted, it reprocesses the orchestration with t
 Orchestration API consist of the following handlers:
 
 - `GET /orchestrations` - exposes data about all orchestrations.
-- `GET /orchestrations/{orchestration_id}` - exposes data about a single orchestration status.
+- `GET /orchestrations/{orchestration_id}` - exposes status of a single orchestration.
```
klaudiagrz, line 15:
```suggestion
- `GET /orchestrations/{orchestration_id}` - exposes the status of a single orchestration.
```

docs/kyma-environment-broker/03-10-orchestration.md
```diff
@@ -42,3 +43,11 @@ The example strategy configuration looks as follows:
   }
 }
 ```
+
+## Cancelation
+
+You can cancel any orchestration that is in progress or pending using the `PUT /orchestrations/{orchestration_id}/canceled` endpoint described above.
```
klaudiagrz, line 49:
```suggestion
You can cancel any orchestration that is in progress or pending using the `PUT /orchestrations/{orchestration_id}/canceled` endpoint.
```

docs/kyma-environment-broker/03-10-orchestration.md
```diff
@@ -42,3 +43,9 @@ The example strategy configuration looks as follows:
   }
 }
 ```
+
+## Cancelation
+
+You can cancel any orchestration that is in progress or pending using the `PUT /orchestrations/{orchestration_id}/cancel` endpoint. 
+After you cancel an orchestration, KEB sets its state to `Canceling`. An orchestration with such a state does not schedule any new operations.
+To provide consistency, a canceled orchestration waits for already processed operations to finish. When operations are finished, the processed orchestration's state is set to `Canceled` and the next orchestration from the queue is start to being processed.
```
klaudiagrz, line 51:
```suggestion
To provide consistency, a canceled orchestration waits for already processed operations to finish. When operations are finished, the processed orchestration's state is set to `Canceled` and the next orchestration from the queue starts to be processed.
```

docs/kyma-environment-broker/03-10-orchestration.md
```diff
@@ -5,14 +5,15 @@ type: Details
 
 Orchestration is a mechanism that allows you to upgrade Kyma Runtimes. To create an orchestration, [follow this tutorial](#tutorials-orchestrate-kyma-upgrade). After sending the request, the orchestration is processed by `KymaUpgradeManager`. It lists Shoots (Kyma Runtimes) in the Gardener cluster and narrows them to the IDs that you have specified in the request body. Then, `KymaUpgradeManager` performs the [upgrade steps](#details-runtime-operations) logic on the selected Runtimes.
 
-If Kyma Environment Broker is restarted, it reprocesses the orchestration with the `IN PROGRESS` state.
+If Kyma Environment Broker is restarted, it reprocesses the orchestration with the `CANCELING`, `IN PROGRESS`, and `PENDING` state.
```
klaudiagrz, line 8:
```suggestion
If Kyma Environment Broker is restarted, it reprocesses the orchestrations that were previously in the `CANCELING`, `IN PROGRESS`, and `PENDING` state.
```

docs/kyma-environment-broker/03-10-orchestration.md
```diff
@@ -42,3 +43,9 @@ The example strategy configuration looks as follows:
   }
 }
 ```
+
+## Cancelation
+
+You can cancel any orchestration that is in progress or pending using the `PUT /orchestrations/{orchestration_id}/cancel` endpoint. 
+After you cancel an orchestration, KEB sets its state to `Canceling`. An orchestration with such a state does not schedule any new operations.
+To provide consistency, a canceled orchestration waits for already processed operations to finish. When operations are finished, the processed orchestration's state is set to `Canceled` and the next orchestration from the queue is processed.
```
klaudiagrz, line 51:
```suggestion
To provide consistency, a canceled orchestration waits for already processed operations to finish. When operations are finished, the processed orchestration's state is set to `Canceled` and the next orchestration from the queue starts being processed.
```

klaudiagrz (approved):

ksputo (approved):
lgtm

PR #359, opened 2020-11-30
Trim runtime name if its too long

kjaksik:
**Description**

- Trim runtime name if its too long

**Related issue(s)**


crabtree (commented):
Can you cover this with some test cases?

crabtree (approved):

PR #368, opened 2020-12-03
Fix for upgrade concurrency issue in Provisioner

akgalwas:
**Description**

UpgradeKymaStep which is a first step in upgrade doesn't requeue  operation. As a result operation occupies a queue and can starve upgrades waiting in the queue (and they will timeout instantaneously after picking from the queue). 

- Concurrency level for upgrade operation increased


koala7659 (approved):
It seems to work fine. I tested it with 20 parallel upgrades. No problems seen in the logs

PR #373, opened 2020-12-04
Correcting images for provisioner and provisioner tests

koala7659:
Update both images to the latest version


franpog859 (approved):
Please, keep your eye on the `pre-master-control-plane-gke-provisioner-integration` job and the test logs themself. The job will probably fail but we can check on the running cluster if everything is ok or not

PR #376, opened 2020-12-08
fix trials removal

k15r:
**Description**

- introduce ratelimiting
- improve tracing

**Related issue(s)**


components/metris/internal/provider/azure/azure.go
```diff
@@ -103,93 +106,130 @@ func (a *Azure) Run(ctx context.Context) {
...
+	if err != nil {
+		errorOccurred = true
+		eventData, rateLimited, instanceDeleted = processError(ctx, workerlogger, instance, eventData, err, a.config.MaxRetries, a.instanceStorage)
+	}
 
-				// requeue item after X duration if client still in storage
-				if !a.queue.ShuttingDown() {
-					workerlogger.Debugf("requeuing cluster in %s", a.config.PollInterval)
-					a.queue.AddAfter(clusterid, a.config.PollInterval)
+	if !errorOccurred {
```
maksd, line 165: Is this "if" really necessary, could it be merge with the previous error check? I don't see the variable "errorOccurred" used anywhere else.

components/metris/internal/provider/azure/azure.go
```diff
@@ -103,93 +106,130 @@ func (a *Azure) Run(ctx context.Context) {
...
+				a.queue.Done(clusterid)
+				if !rateLimited {
+					a.queue.Forget(clusterid)
+				}
+				if a.queue.ShuttingDown() {
+					workerlogger.Debugf("queue is shutting down, can't requeue cluster, processing cluster %s one last time", clusterid)
 				} else {
-					workerlogger.Warnf("vm capabilities for region %s not found, some metrics won't be available", instance.cluster.Region)
+					workerlogger.Debugf("enqueueing '%s'", clusterid)
+					a.queue.AddRateLimited(clusterid)
```
maksd, line 119: If `rateLimited` is **false**  you should use `AddAfter` with `config.PollInterval`, and if it is rate limited then use [AddRateLimited](https://github.com/kubernetes/client-go/blob/v0.19.4/util/workqueue/rate_limiting_queue.go#L59-L61), because `AddRateLimited` [increment the number of failure](https://github.com/kubernetes/client-go/blob/v0.19.4/util/workqueue/default_rate_limiters.go#L94) before calculating the delay. So, for example if everything was OK, then it would retry after 10min `(5min*2^1)` instead of the 5min.

maksd:
You should update the file README.md with the provider config changes, and also add them to the chart templates and values.

components/metris/internal/provider/azure/azure.go
k15r: `AddAfter` does not increase the failure increment. So when the next scheduled execution fails and would cause a ratelimited retry it would not be delayed by `duration *2^1` but by `duration*2^0`
Thats why we always use AddRateLimited and check if the call should be ratelimited or if it should be scheduled with the normal delay. In case we need the normal delay we call `forget` before adding the new schedule, otherwise we just add ratelimited

components/metris/internal/provider/azure/azure.go
k15r: done

components/metris/internal/provider/azure/types.go
```diff
@@ -143,35 +138,35 @@ type ClientSecretMap struct {
 	EnvironmentName string
 }
 
-// VMType defines the event format for the virtual machine metrics.
+// VMType defines the eventData format for the virtual machine metrics.
```
marcobebway, line 141: Do we need this change?

components/metris/internal/provider/azure/types.go
```diff
@@ -143,35 +138,35 @@ type ClientSecretMap struct {
...
 type VMType struct {
 	Name  string `json:"name"`
 	Count uint32 `json:"count"`
 }
 
-// ProvisionedVolume defines the event format for the volume metrics.
+// ProvisionedVolume defines the eventData format for the volume metrics.
```
marcobebway, line 147: Do we need this change?

components/metris/internal/provider/azure/types.go
```diff
@@ -143,35 +138,35 @@ type ClientSecretMap struct {
...
 type ProvisionedVolume struct {
 	SizeGBTotal   uint32 `json:"size_gb_total"`
 	SizeGBRounded uint32 `json:"size_gb_rounded"`
 	Count         uint32 `json:"count"`
 }
 
-// Compute defines the event format for the compute metrics.
+// Compute defines the eventData format for the compute metrics.
```
marcobebway, line 154: Do we need this change?

components/metris/internal/provider/azure/types.go
```diff
@@ -143,35 +138,35 @@ type ClientSecretMap struct {
...
 type Compute struct {
 	VMTypes            []VMType          `json:"vm_types"`
 	ProvisionedRAMGB   float64           `json:"provisioned_ram_gb"`
 	ProvisionedVolumes ProvisionedVolume `json:"provisioned_volumes"`
 	ProvisionedCpus    uint32            `json:"provisioned_cpus"`
 }
 
-// Networking defines the event format for the network metrics.
+// Networking defines the eventData format for the network metrics.
```
marcobebway, line 162: Do we need this change?

components/metris/internal/provider/azure/types.go
```diff
@@ -143,35 +138,35 @@ type ClientSecretMap struct {
...
 type Networking struct {
 	ProvisionedLoadBalancers uint32 `json:"provisioned_loadbalancers"`
 	ProvisionedVnets         uint32 `json:"provisioned_vnets"`
 	ProvisionedIps           uint32 `json:"provisioned_ips"`
 }
 
-// EventHub defines the event format for the event hub metrics.
+// EventHub defines the eventData format for the eventData hub metrics.
```
marcobebway, line 169: Do we need this change?

components/metris/internal/provider/azure/types.go
```diff
@@ -182,7 +177,7 @@ type EventHub struct {
 	MaxOutgoingBytesPT5M float64 `json:"max_outgoing_bytes_pt5m"`
 }
 
-// EventData defines the event information to send to EDP.
+// EventData defines the eventData information to send to EDP.
```
marcobebway, line 180: Do we need this change?

components/metris/internal/provider/azure/types.go
```diff
@@ -48,10 +48,7 @@ const (
 	// PT5M ...
 	PT5M TimeGrain = "PT5M"
 
-	// maximum number of failed attempt to get metrics, after that instance is remove from cache/storage.
-	maxRetryAttempts int = 5
-
-	responseErrCodeResourceGroupNotFound string = "ResourceGroupNotFound"
+	ResponseErrCodeResourceGroupNotFound string = "ResourceGroupNotFound"
```
marcobebway, line 51: Why exporting const?

components/metris/internal/provider/azure/instance_test.go
```diff
@@ -188,7 +188,7 @@ func TestInstance_getComputeMetrics(t *testing.T) {
 			want:   metricsNoDisk,
 		},
 		{
-			name:    "with last event",
+			name:    "with last eventData",
```
marcobebway, line 191: Do we need this change?

components/metris/internal/provider/azure/azure.go
```diff
@@ -103,93 +106,133 @@ func (a *Azure) Run(ctx context.Context) {
...
+func processError(ctx context.Context, workerlogger log.Logger, instance *Instance, eventData *EventData, err error, maxRetries int, instanceStorage storage.Storage) (*EventData, bool, bool) {
+	var span *trace.Span
+	if tracing.IsEnabled() {
+		ctx, span = trace.StartSpan(ctx, "metris/provider/azure/processError")
+		defer span.End()
+	}
+	eventData = instance.lastEvent
+	workerlogger.With("error", err).Error("could not get metrics, using information from cache")
+	if eventData == nil {
+		workerlogger.With("error", err).Error("could not get metrics from cache, dropping events because no cached information")
```
marcobebway, line 189: I believe this log can be moved to whoever called `processError`. Because the decision to act on nil `eventData` is done later.

components/metris/internal/provider/azure/azure.go
```diff
@@ -241,7 +284,7 @@ func (a *Azure) clusterHandler(parentctx context.Context) {
 
 				if rg, err := instance.client.GetResourceGroup(parentctx, "", filter, logger); err != nil {
 					if !cluster.Trial {
-						logger.Warnf("could not find a resource group for event hub, cluster may not be ready, retrying in %s: %s", a.config.PollInterval, err)
+						logger.Warnf("could not find a resource group for eventData hub, cluster may not be ready, retrying in %s: %s", a.config.PollInterval, err)
```
marcobebway, line 287: Do we need this change?

components/metris/internal/provider/azure/types.go
k15r: goland magically decided to do this

components/metris/internal/provider/azure/instance_test.go
k15r: i will revert them, goland was running rogue

components/metris/internal/provider/azure/azure.go
```diff
@@ -103,93 +106,131 @@ func (a *Azure) Run(ctx context.Context) {
...
+	if errdetail, ok := err.(autorest.DetailedError); ok {
+		err = errdetail
 
-				// requeue item after X duration if client still in storage
-				if !a.queue.ShuttingDown() {
-					workerlogger.Debugf("requeuing cluster in %s", a.config.PollInterval)
-					a.queue.AddAfter(clusterid, a.config.PollInterval)
+		switch errdetail.StatusCode {
+		// Check if the error is a resource group not found, then it would mean
+						// that the cluster may have been deleted, and gardener did not trigger
```
marcobebway, line 197: Whitespaces.

components/metris/internal/provider/azure/clients_test.go
```diff
@@ -744,7 +744,7 @@ func Test_client_getMetricValues(t *testing.T) {
 		urlEscape("%s?%s", rguri, novalueparams):  novalueList,
 		urlEscape("%s?%s", rguri, notsparams):     notsList,
 		urlEscape("%s?%s", rguri, notsdataparams): notsdatalist,
-	})
+	}components/metris/internal/provider/azure/instance.go)
```
marcobebway, line 747: ?

marcobebway (dismissed):

components/metris/README.md
```diff
@@ -18,12 +18,16 @@ Metris is a metering component that collects data and sends them to EDP.
 | `--edp-buffer` | **EDP_BUFFER** | Number of events that the buffer can have | `100` |
 | `--edp-workers` | **EDP_WORKERS** | Number of workers to send metrics | `5` |
 | `--edp-event-retry` | **EDP_RETRY** | Number of retries for sending an event | `5` |
-| `--provider-poll-interval` | **PROVIDER_POLLINTERVAL** | Interval at which metrics are fetched | `1m` |
+| `--provider-poll-interval` | **PROVIDER_POLLINTERVAL** | Interval at which metrics are fetched | `5m` |
+| `--provider-poll-max-interval` | **PROVIDER_POLLMAXINTERVAL** | maximum Interval at which metrics are fetch | `15m` |
 | `--provider-poll-duration` | **PROVIDER_POLLDURATION** | Time limit for requests made by the provider client | `5m` |
+| `--provider-max-retries` | **PROVIDER_MAXRETRIES** | Number of maximum retries until a cluster is removed from the cache in case of not being found on the provider. This will stop sending events for the removed cluster! | `20` |
```
alexandra-simeonova, line 24:
```suggestion
| `--provider-max-retries` | **PROVIDER_MAXRETRIES** | Maximum number of retries before a cluster is removed from the cache if it is not found on the provider. NOTE: This will stop sending events for the removed cluster | `20` |
```

It's not a big deal, but we normally avoid exclamation points and use "Note" or "Caution" [instead](https://kyma-project.io/community/guidelines/content/#formatting-formatting-panels).

alexandra-simeonova (approved):
🚀

marcobebway (dismissed):

marcobebway (approved):

alexandra-simeonova (approved):

janmedrek (approved):

components/metris/internal/provider/azure/azure.go
maksd: Hi, I looked again at the `When` function and saw that they increase the value after checking it, and you are right, it's all good then.

maksd (approved):

PR #379, opened 2020-12-09
Keb context updates

piotrmiskiewicz:
**Description**

- service contains a flag "allow_context_updates" which enables context updates by KEB client
- added more logging to the update endpoint logic to see what we get

### DIFF
--- a/components/kyma-environment-broker/internal/broker/instance_update.go
+++ b/components/kyma-environment-broker/internal/broker/instance_update.go
@@ -33,7 +33,7 @@ func (b *UpdateEndpoint) Update(ctx context.Context, instanceID string, details 
 
 	instance, err := b.instanceStorage.GetByID(instanceID)
 	if err != nil {
-		logger.Errorf("unable to get instance")
+		logger.Errorf("unable to get instance: %s", err.Error())
 		return domain.UpdateServiceSpec{}, errors.New("unable to get instance")
 	}
 	logger.Infof("Plan ID/Name: %s/%s", instance.ServicePlanID, PlanIDsMapping[instance.ServicePlanID])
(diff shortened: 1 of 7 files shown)


components/kyma-environment-broker/internal/broker/instance_update.go
```diff
@@ -2,26 +2,60 @@ package broker
...
+	var ersContext internal.ERSContext
+	err = json.Unmarshal(details.RawContext, &ersContext)
+	if err != nil {
+		logger.Errorf("unable to decode context: %s", err.Error())
+		return domain.UpdateServiceSpec{}, nil
+	}
+	logger.Infof("Global account ID: %s active: %v", instance.GlobalAccountID, ersContext.Active)
+
+	var contextData map[string]interface{}
+	_ = json.Unmarshal(details.RawContext, contextData)
```
kjaksik, line 50: Why don't we handle this error?

kjaksik: Btw my IDE says that `contextData` should be a pointer 😄

components/kyma-environment-broker/internal/broker/instance_update.go
```diff
@@ -2,26 +2,60 @@ package broker
...
-	b.log.Infof("Update instanceID: %s", instanceID)
-	b.log.Infof("Update details: %+v", details)
-	b.log.Infof("Update asyncAllowed: %v", asyncAllowed)
+	logger := b.log.WithField("instanceID", instanceID)
+	logger.Infof("Update instanceID: %s", instanceID)
+	logger.Infof("Update asyncAllowed: %v", asyncAllowed)
+
+	instance, err := b.instanceStorage.GetByID(instanceID)
+	if err != nil {
+		logger.Errorf("unable to get instance")
```
kjaksik, line 36: What do you think about logging the error that we get from the `GetByID` method too?

components/kyma-environment-broker/internal/broker/instance_update.go
piotrmiskiewicz: the second unmarshalling is just for a POC, but will handle the error

kjaksik (approved):

PR #395, opened 2020-12-16
KEB instance model with versioning

piotrmiskiewicz:
**Description**

- add optimistick locking for instsance updates (KEB)

### DIFF
--- a/components/kyma-environment-broker/internal/model.go
+++ b/components/kyma-environment-broker/internal/model.go
@@ -104,6 +104,8 @@ type Instance struct {
 	CreatedAt time.Time
 	UpdatedAt time.Time
 	DeletedAt time.Time
+
+	Version int
 }
 
 func (instance Instance) GetProvisioningParameters() (ProvisioningParameters, error) {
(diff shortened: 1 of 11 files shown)


components/kyma-environment-broker/internal/storage/dbsession/write.go
```diff
@@ -73,10 +75,19 @@ func (ws writeSession) UpdateInstance(instance internal.Instance) dberr.Error {
 		Set("provisioning_parameters", instance.ProvisioningParameters).
 		Set("provider_region", instance.ProviderRegion).
 		Set("updated_at", time.Now()).
+		Set("version", instance.Version+1).
 		Exec()
 	if err != nil {
 		return dberr.Internal("Failed to update record to Instance table: %s", err)
 	}
+	rAffected, e := res.RowsAffected()
```
ksputo, line 83: i think there is no need to change the error naming convention, we can overwrite previous `err` value here
```suggestion
	rAffected, err := res.RowsAffected()
```

piotrmiskiewicz: done

ksputo (approved):

PR #404, opened 2020-12-22
Version mapping for PR must be lowercase

ksputo:
**Description**
When performing account version mapping, you can't use uppercase for ConfigMaps `metadata.name` field:
```
Error from server (Invalid): error when creating "overrides": ConfigMap "tracing-global-overrides-PR-123" is invalid: metadata.name: Invalid value: "tracing-global-overrides-PR-123": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

On the other hand, the uppercase format is needed for fetching `kyma-installer.yaml` file properly:
https://github.com/kyma-project/control-plane/blob/9465c2f0d2e65053657d94d2769602f82b02d15f/components/kyma-environment-broker/internal/runtime/provider.go#L166-L171

- Changed `kymaVersion` string handling when processing versions from PR to lowercase for example `pr-123`

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/process/input/builder_test.go
+++ b/components/kyma-environment-broker/internal/process/input/builder_test.go
@@ -118,15 +118,15 @@ func TestInputBuilderFactory_ForPlan(t *testing.T) {
 		// given
 		componentsProvider := &automock.ComponentListProvider{}
 		componentsProvider.On("AllComponents", "1.10").Return([]v1alpha1.KymaComponent{}, nil).Once()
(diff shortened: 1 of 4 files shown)


ksputo:
not needed: logic works fine, only the suffix in ConfigMap name needs to be lowercase.

PR #405, opened 2020-12-22
Fix AVS external evaluation creation

ksputo:
**Description**
Provisioning a SKR with `TrialPlanID` caused setting the `disabled` flag in `externalEvalCreator` to `false` permanently

- Fixed AVS external evaluation creation

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/process/provisioning/external_eval.go
+++ b/components/kyma-environment-broker/internal/process/provisioning/external_eval.go
@@ -24,6 +24,7 @@ func NewExternalEvalCreator(delegator *avs.Delegator, disabled bool, assistant *
 
 func (eec *ExternalEvalCreator) createEval(operation internal.ProvisioningOperation, url string, logger logrus.FieldLogger) (internal.ProvisioningOperation, time.Duration, error) {
 	if eec.disabled {
+		logger.Infof("Creating AVS external evaluation is disabled")
 		return operation, 0, nil
 	} else {
 		return eec.delegator.CreateEvaluation(logger, operation, eec.assistant, url)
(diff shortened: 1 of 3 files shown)


components/kyma-environment-broker/internal/process/provisioning/initialisation.go
```diff
@@ -256,6 +249,24 @@ func (s *InitialisationStep) launchPostActions(operation internal.ProvisioningOp
...
+		log.Errorf("cannot fetch provisioning parameters from operation: %s", err)
+		return s.operationManager.OperationFailed(operation, "invalid operation provisioning parameters")
+	}
+	if pp.PlanID == broker.TrialPlanID {
+		log.Info("skipping AVS external evaluation creation for trial plan")
+		return operation, 0, nil
+	}
+	log.Infof("creating external evaluation for instance %", instance.InstanceID)
+	operation, repeat, err := s.externalEvalCreator.createEval(operation, instance.DashboardURL, log)
+	if err != nil || repeat != 0 {
```
polskikiel, line 264: nit: maybe it'll be better to log error message if it's not null and just `return operation, retry, nil`

components/kyma-environment-broker/internal/process/provisioning/external_eval.go
```diff
@@ -24,6 +24,7 @@ func NewExternalEvalCreator(delegator *avs.Delegator, disabled bool, assistant *
 
 func (eec *ExternalEvalCreator) createEval(operation internal.ProvisioningOperation, url string, logger logrus.FieldLogger) (internal.ProvisioningOperation, time.Duration, error) {
 	if eec.disabled {
```
polskikiel, line 26: do we need this check at all?

ksputo: imo we should, because this is set by this flag globally:
https://github.com/kyma-project/control-plane/blob/81c77a1ef304437c78d877c740bf814e68c34804/components/kyma-environment-broker/internal/avs/config.go#L17

and we use it on `stage`

polskikiel (approved):
lgtm

polskikiel (approved):

PR #412, opened 2020-12-30
Set custom cloudsql-proxy image

crabtree:
**Description**

- Set custom cloudsql-proxy image

**Related issue(s)**


franpog859 (approved):
Looks fine! I'm afraid though that the `pre-master-control-plane-gke-provisioner-integration` job may not test the migrator jobs correctly. As far as I remember, the `.Values.global.database.embedded.enabled` equals `true` in this case so there is no cloudsql-proxy container in the pod.  We may want to test this cloudsql-proxy image manually on the DEV environment. Anyway, it should be fine because it's not a big change as far as I understand it correctly

kjaksik (approved):

Issue #399, opened 2020-12-17
KEB stores all necessary information for next processes


piotrmiskiewicz:
**Description**

Currently, the provisioning operation contains all necessary IDs of external services, which are used for deprovisioning.
The first step of the deprovisioning process finds the ProvisioningOperation and copies necessary field. The problem is that such data could be stored (in the near future) also in the UpgradeKyma operation. 
AC:
 - find a solution how to store it and a way to get it, compare proposed solutions by writing pros and cons

Discussed solutions:
1. Create a new entity, similar to RuntimeState, which contains all those information and such entity would be created at the end of every process (even if the process fails, so it should not be a part of the last step). Then taking such information would be easy - just take the last one.
2. load all operations for given instance and process them one by one to collect all IDs

piotrmiskiewicz:
## Option 1
Instead of defining fields in the XXXOperation like this:
```
type ProvisioningOperation struct {
	Operation `json:"-"`

	Lms Lms `json:”lms”`
}
```

all IDs, names are inside a struct:

```
type ProvisioningOperation struct {
	Operation `json:"-"`

	InstanceDetails InstanceDetails `json:"-"`
}

type InstanceDetails struct {
	Lms Lms `json:”lms”`
}
```

Then, when we are creating some operation, which is not the first one, we only need to do:
```
lastOper := b.operationsStorage.GetLastOperation(instanceID) // we need to implement such method in the storage
operation.InstanceDetails = lastOper.InstanceDetails
```

This solution gives us a confidence that all IDs, names of external resources are saved and could be used in the next operation. This solution stores snapshots of data for all operations but not in a separate table (it is a technical detail, we can do it on a storage layer, but then we need a transaction).

## Option 2
Separate entity, table which stores all IDs, names.
It must be up to date all the time, so we can't just create such "snapshot" only and the end of processing. The KEB Pod could be restarted just after an operation update. The operation failure is saved inside a step, for example:
```
return s.operationManager.OperationFailed(operation, msg)
```

## Option 3
Everytime we need such IDs, names, we are reading all previous operations and extract such data.

## Summary:
Option 1:
(-) data migration - we need to update all existing upgradeKyma operation
(-) code refactor
(+) when the work above is done, no changes for the future, everything will work the same
(+) clean state in the DB (just get last operation and you have all those values)

Option 2:
(-) we need to provide a logic which copies all data to the new entity. So every new field in the ProvisioningOperation needs a line in the copying logic
[...]

PR #417, opened 2021-01-05
Return only non-pending operations when getting last operation

ksputo:
**Description**

- Added extra condition when getting last operation for given `instanceID` to ensure the operation is not in `pending` state

**Related issue(s)**
#399

### DIFF
--- a/components/kyma-environment-broker/internal/storage/storage_test.go
+++ b/components/kyma-environment-broker/internal/storage/storage_test.go
@@ -491,6 +491,24 @@ func TestSchemaInitializer(t *testing.T) {
 					},
 				},
 			}
+			latestPendingOperation := internal.ProvisioningOperation{
+				Operation: internal.Operation{
+					ID:    "latest-id-pending",
+					State: orchestration.Pending,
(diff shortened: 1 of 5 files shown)


components/kyma-environment-broker/internal/storage/storage_test.go
```diff
@@ -491,6 +491,24 @@ func TestSchemaInitializer(t *testing.T) {
...
+					OrchestrationID:        orchestrationID,
+					ProvisionerOperationID: "target-op-id",
+					Description:            "description",
+					Version:                1,
+					ProvisioningParameters: internal.ProvisioningParameters{},
+					InstanceDetails: internal.InstanceDetails{
+						Lms: internal.LMS{TenantID: "tenant-id"},
+					},
+				},
+			}
```
piotrmiskiewicz, line 512: please add also a pending operation in the test:
```
t.Run("Instances", func(t *testing.T) {
		t.Run("Should create and update instance", func(t *testing.T) {```

with different lms tenant Id just to be sure that such operation is not taken into account when the storage fetches instance details.

piotrmiskiewicz (approved):

PR #423, opened 2021-01-06
Fix parameters decryption

polskikiel:
**Description**

- Fixed parameters decryption

**Related issue(s)**


ksputo:
change processing operation flow in every place where `toOperation()` method is used: first, we need to unmarshall `operation.Data` field to get `InstanceDetails`, save it and then after calling `toOperation` which decrypts creds we need to overwrite `InstanceDetails` field in operation

jasiu001 (approved):

PR #426, opened 2021-01-07
Fix AVS trial eval removal during deprovisioning

ebensom:
**Description**

Wrong parent evaluation ID is being used when trying to remove trial evaluation from parent (compound) evaluation.

### DIFF
--- a/components/kyma-environment-broker/internal/avs/client_test.go
+++ b/components/kyma-environment-broker/internal/avs/client_test.go
@@ -163,7 +163,7 @@ func TestClient_RemoveReferenceFromParentEval(t *testing.T) {
 		assert.NoError(t, err)
 
 		// When
-		err = client.RemoveReferenceFromParentEval(resp.Id)
+		err = client.RemoveReferenceFromParentEval(parentEvaluationID, resp.Id)
 
 		// Then
 		assert.NoError(t, err)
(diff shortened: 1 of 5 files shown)


components/kyma-environment-broker/internal/avs/client_test.go
```diff
@@ -181,7 +181,7 @@ func TestClient_RemoveReferenceFromParentEval(t *testing.T) {
 		assert.NoError(t, err)
 
 		// When
-		err = client.RemoveReferenceFromParentEval(111)
+		err = client.RemoveReferenceFromParentEval(parentEvaluationID, 111)
```
ksputo, line 184: can we add another test case to cover the case that you've fixed?
something like:

```golang
	t.Run("should return error when parent evaluation does not contain subevaluation", func(t *testing.T) {
		// Given
		server := newServer(t)
		mockServer := fixHTTPServer(server)
		client, err := NewClient(context.TODO(), Config{
			OauthTokenEndpoint: fmt.Sprintf("%s/oauth/token", mockServer.URL),
			ApiEndpoint:        fmt.Sprintf("%s/api/v2/evaluationmetadata", "http://not-existing"),
			ParentId:           parentEvaluationID,
		}, logrus.New())
		assert.NoError(t, err)

		// When
		err = client.RemoveReferenceFromParentEval(int64(9999), 111)

		// then
		assert.Error(t, err)
	})
	```
	
and then adjusting fake client behaviour here:
https://github.com/kyma-project/control-plane/blob/0b7b8c6a4c18a1bc0a23094e948c84080d76118a/components/kyma-environment-broker/internal/avs/client_test.go#L411

I think it should return `400` instead of `404` in case subevaluation we're trying to delete does not "belong" to given parent id evaluation

ksputo (approved):
lgtm

PR #434, opened 2021-01-11
Evaluation status flow on Kyma upgrade

fhivemind:
**Description**

Injects pre- and post- logic during Kyma upgrades. Specifically, sets Avs evaluation monitors status to Maintenance on upgrade initialization, and reverts to original eval status on finish.

**Related issue(s)**


components/kyma-environment-broker/internal/avs/client.go
```diff
@@ -96,6 +96,39 @@ func (c *Client) AddTag(evaluationID int64, tag *Tag) (*BasicEvaluationCreateRes
...
+		return &responseObject, errors.Wrap(err, "while creating SetStatus request")
+	}
+	request.Header.Set("Content-Type", "application/json")
+
+	response, err := c.execute(request, false, true)
+	if err != nil {
+		return &responseObject, errors.Wrap(err, "while executing SetStatus request")
+	}
+	defer func() {
+		if closeErr := c.closeResponseBody(response); closeErr != nil {
```
ebensom, line 119: Misc bug: `closeResponseBody()` doesn't drain the response body before closing it.

components/kyma-environment-broker/internal/avs/delegator.go
```diff
@@ -88,17 +93,54 @@ func (del *Delegator) AddTags(logger logrus.FieldLogger, operation internal.Prov
...
+	return updatedOperation, d, nil
+}
+
+func (del *Delegator) SetStatus(logger logrus.FieldLogger, operation internal.UpgradeKymaOperation, evalAssistant EvalAssistant, status string) (internal.UpgradeKymaOperation, time.Duration, error) {
+	logger.Infof("starting the SetStatus to avs internal id [%d]", operation.Avs.AvsEvaluationInternalId)
+	var updatedOperation internal.UpgradeKymaOperation
+	d := 0 * time.Second
+
+	evalId := evalAssistant.GetEvaluationId(operation.Avs)
+	currentStatus := evalAssistant.GetEvalStatus(operation.Avs)
```
ebensom, line 117: NOTE: Avs lifecycle data in `UpgradeKymaOperation` will be present after https://github.com/kyma-project/control-plane/pull/437

components/kyma-environment-broker/internal/process/upgrade_kyma/initialisation.go
```diff
@@ -64,6 +68,14 @@ func (s *InitialisationStep) Name() string {
 func (s *InitialisationStep) Run(operation internal.UpgradeKymaOperation, log logrus.FieldLogger) (internal.UpgradeKymaOperation, time.Duration, error) {
 	if operation.State == orchestrationExt.Canceled {
 		log.Infof("Skipping processing because orchestration %s was canceled", operation.OrchestrationID)
+
+		// revert to original evaluation status
+		operation, _, err := s.internalEvalUpdater.RestoreStatusToEval(operation, log)
```
ebensom, line 73: state can only be canceled in rare condition situation when the operation was pending (not in progress) and the operation started to be executed concurrently with the cancellation. So this code is not needed

components/kyma-environment-broker/internal/process/upgrade_kyma/internal_eval_updater.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type InternalEvalUpdater struct {
+	delegator *avs.Delegator
+	assistant *avs.InternalEvalAssistant
+	avsConfig avs.Config
+}
+
+func NewInternalEvalUpdater(delegator *avs.Delegator, assistant *avs.InternalEvalAssistant, config avs.Config) *InternalEvalUpdater {
```
ebensom, line 17: I suggest to refactor this to `EvalUpdater` and add logic to also turn the status of the external evaluation.

So one function to set status for both the internal and external eval, and another function to restore the status for both of them.

components/kyma-environment-broker/internal/process/upgrade_kyma/initialisation.go
```diff
@@ -171,6 +183,33 @@ func (s *InitialisationStep) configureKymaVersion(operation *internal.UpgradeKym
...
+		if !inMaintenance {
+			operation, _, errStatus = s.internalEvalUpdater.SetStatusToEval(avs.StatusMaintenance, operation, log)
+			if errStatus != nil {
+				err = errors.Wrap(err, errStatus.Error())
+			}
+		}
+	} else if operation.State == orchestrationExt.Succeeded || operation.State == orchestrationExt.Failed {
+		// restore previous evaluation state on finish
+		if inMaintenance {
+			operation, _, errStatus = s.internalEvalUpdater.RestoreStatusToEval(operation, log)
```
ebensom, line 203: So since the AVS statuses are introduced with this PR, the `Original` and `Current` statuses doesn't exist in DB for existing operations records. As a result, JSON deserialization would create the `UpgradeKymaOperation` objects with empty (`""`) AVS status values, and you would restore the empty status both in the DB record and when making the AVS API call which would most probably result in `500 Internal Server Error`. 

I'm thinking of adding a small logic to obtain the `Current` status from AVS through API call when it's empty in DB.

components/kyma-environment-broker/internal/process/upgrade_kyma/internal_eval_updater.go
fhivemind: It does makes sense, I completely forgot about the fact that external monitors will still ping the systems during the upgrade.

components/kyma-environment-broker/internal/avs/delegator.go
fhivemind: The current logic relies on migrations, and this shouldn't be the case. To support the scenario when the migrations are not working, it's possible to have default states (`Maintenance` on upgrade init and `Active` on finish). The worst-case scenario is that we'll set `Active` on upgrade finish for a monitor that had `originalState != Active` before the upgrade, due to inability to preserve states. However, this edge-case will be resolved as soon as #437 is merged.

githubgnagy:
PLease note and prepare for situation when external evaluation does not exist. Which is the case for recently provisioned clusters and trial clusters.

components/kyma-environment-broker/internal/process/upgrade_kyma/upgrade_kyma_step.go
```diff
@@ -19,10 +19,16 @@ type UpgradeKymaStep struct {
...
+	evaluationManager   *EvaluationManager
 	timeSchedule        TimeSchedule
 }
 
-func NewUpgradeKymaStep(os storage.Operations, runtimeStorage storage.RuntimeStates, cli provisioner.Client, timeSchedule *TimeSchedule) *UpgradeKymaStep {
+func NewUpgradeKymaStep(
+	os storage.Operations,
+	runtimeStorage storage.RuntimeStates,
+	cli provisioner.Client,
+	em *EvaluationManager,
```
ebensom, line 30: `EvaluationManager` is not used now in upgrade kyma step

components/kyma-environment-broker/internal/process/upgrade_kyma/evaluation_manager.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// SetStatus updates evaluation monitors (internal and external) status.
+// Note that this operation should be called twice (reason behind the zero delay)
+// to configure both monitors.
+func (em *EvaluationManager) SetStatus(status string, operation internal.UpgradeKymaOperation, logger logrus.FieldLogger) (internal.UpgradeKymaOperation, time.Duration, error) {
+	avsCfg := operation.Avs
+
+	// do internal monitor status update
+	if em.internalAssistant.IsInMaintenance(avsCfg) {
```
ebensom, line 34: Erroneous check for maintenance status, call from `SetMaintenanceStatus` will not set the maintenance status because these conditions skip processing.

components/kyma-environment-broker/internal/avs/delegator.go
```diff
@@ -1,33 +1,39 @@
...
 	"github.com/sirupsen/logrus"
 )
 
 type Delegator struct {
-	operationManager  *process.ProvisionOperationManager
-	avsConfig         Config
-	client            *Client
-	operationsStorage storage.Operations
+	provisionManager   *process.ProvisionOperationManager
+	deprovisionManager *process.DeprovisionOperationManager
```
ebensom, line 17: `deprovisionManager` is not used anywhere.

components/kyma-environment-broker/internal/process/upgrade_kyma/evaluation_manager.go
fhivemind: Good observation!

ebensom (approved):

kjaksik (approved):

PR #436, opened 2021-01-12
Dump KEB stack traces

polskikiel:
**Description**

- Dump KEB stack traces
- 
**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/process/queue.go
+++ b/components/kyma-environment-broker/internal/process/queue.go
@@ -1,6 +1,7 @@
 package process
 
 import (
+	"runtime/debug"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


components/kyma-environment-broker/internal/process/queue.go
```diff
@@ -68,7 +69,7 @@ func worker(queue workqueue.RateLimitingInterface, process func(key string) (tim
 				log = log.WithField("operationID", id)
 				defer func() {
 					if err := recover(); err != nil {
-						log.Errorf("panic error from process: %v", err)
+						log.Errorf("panic error from process: %v\n%s", err, debug.Stack())
```
ebensom, line 72: After looking at the result log, this should be modified like "panic error from process: %v. Stacktrace: %s", and also in parallel orchestration strategy. I think this should be merged as it's very useful to capture bugs in operation processing.

ebensom (approved):

kjaksik (dismissed):

kjaksik (approved):

PR #443, opened 2021-01-14
Reset CreatedAt during deprovision reprocessing

### LINKED ISSUE #429: Cannot reprocess operation because of global timeout
**Description**

In KEB we implemented the global timeout for operation processing which is based on the operation's `createdAt` time

It fails when user reprocess the request after time longer then timeout. The solution could be to reset createdAt time in case of reprocessing

**Expected result**

I can reprocess any operation whenever i want to

ebensom:
**Description**

Resetting the `CreatedAt` when reprocessing a failed deprovision operation is needed for the global operation timeout to work correctly. 
Also, operation duration in metrics is exposed erroneously when the `CreatedAt` field is not reset.

**Related issue(s)**
Resolves #429


piotrmiskiewicz:
it looks like we are changing the history. Maybe it would be easy to just create a new operation instead of changing the history (createdAt field looks like once is created, then nobody will changed it)

ebensom:
> it looks like we are changing the history. Maybe it would be easy to just create a new operation instead of changing the history (createdAt field looks like once is created, then nobody will changed it)

I agree, let's implement this in the proper way: create a new deprovision operation instead of resetting fields in the failed operation, and change code not assume there can be only one deprovision operation.

PR #447, opened 2021-01-15
add edp test server

k15r:
**Description**

- add edp test server with validation of edp events
- ...

**Related issue(s)**


maksd (commented):
Would it be a good idea to put the `edp_data_schema.json` and `edp_example_data.json` files into a `/configs` or `/assets` folder?
Other than that everything looks good to me.

PR #450, opened 2021-01-18
Add suspension and unsuspension operations to runtimes list

ksputo:
**Description**
With introducing suspension and unsuspension, we could have multiple provisioning and deprovisioning operations for a single instance. To distinguish them, we should consider following definitions:
- each `Deprovisioning` operation with `Temporary: true` field is a suspension operation
- each `Provisioning` operation after the **first, original** (looking at `created_at` field in db) one is unsuspension operation

- Added suspension and unsuspension operations to `StatusDTO` when listing runtimes on `/runtimes` endpoint

**Related issue(s)**


ebensom:
Please don't forget to allow the `PATCH` request in the ory-oathkeeper rule

piotrmiskiewicz (approved):
lgtm

PR #455, opened 2021-01-19
Additional logic on Kyma upgrade

fhivemind:
**Description**

Adds tests and fixes Kyma upgrade issues for pre- and post- logic on monitor upgrades.
Failures of the additional steps should fail the operation as well. 

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/process/upgrade_kyma/initialisation.go
+++ b/components/kyma-environment-broker/internal/process/upgrade_kyma/initialisation.go
@@ -22,8 +22,8 @@ import (
 )
 
 const (
-	UpgradePreSteps int = iota + 1
-	UpgradePostSteps
+	UpgradeInitSteps int = iota + 1
+	UpgradeFinishSteps
 )
 
 const (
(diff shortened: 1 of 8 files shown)


components/kyma-environment-broker/internal/process/upgrade_kyma/initialisation.go
```diff
@@ -232,13 +245,28 @@ func (s *InitialisationStep) checkRuntimeStatus(operation internal.UpgradeKymaOp
 		msg = *status.Message
 	}
 
+	// execute pre- steps
+	operation, delay, err := s.performRuntimeTasks(UpgradePreSteps, operation, instance, log)
```
ebensom, line 249: The problem with this naming is that this is not a "pre" step, it's executed right after the upgrade has been started.
I would just not use this additional layer of abstraction to try to make it more generic, it increases code complexity (extra levels of function calls and if conditions), without bringing too much benefit.

I would just fall back to ensure AVS evaluations being in maintenance  by using `em.SetMaintenanceStatus()` / `em.RestoreStatus()`  directly when `status.State` is not terminal, and AVS evaluations being restored when they are. 
The logic to return prematurely when there is AVS error even if the provisioning status is Succeeded is good as it won't termonate this KEB operations and will allow retry to ensure AVS status. 

With this you could extract/eliminate functions `executePreUpgradeTasks`, `executePostUpgradeTasks` and `performRuntimeTasks`, saving two levels of function calls, and increasing readability a LOT.

fhivemind: The initial approach was focused on extending the tasks. As mentioned, it doesn't do a good job to verify calls as it can retrigger the same flow (`executePreUpgradeTasks`, `executePostUpgradeTasks`). 

This has been replaced with init and finish steps that will be triggered in `performRuntimeTasks` as that should be an independent handler of all Kyma Upgrade vital dependency calls at some stage of execution.

ebensom (dismissed):

components/kyma-environment-broker/internal/process/upgrade_kyma/initialisation_test.go
```diff
@@ -236,3 +617,266 @@ func fixInstanceRuntimeStatus() internal.Instance {
...
+		t: t,
+		evaluations: &evaluationRepository{
+			basicEvals:   make(map[int64]*avs.BasicEvaluationCreateResponse, 0),
+			evalSet:      make(map[int64]bool, 0),
+			parentIDrefs: make(map[int64][]int64, 0),
+		},
+	}
+}
+
+func fixHTTPServer(srv *server) *httptest.Server {
```
kjaksik, line 669: What do you think about reusing the avs mock client from the avs tests?

kjaksik (approved):

PR #459, opened 2021-01-20
Use the correct fluent bit plugin based on the kyma version

rakesh-garimella:
**Description**

- Pass the correct fluentbit plugin based on the kyma version

**Related issue(s)**


ebensom (commented):
The change doesn't take the fact into account that kyma version for an SKR can come from 3 different sources (in this order of precedence), and you only take the default version:

1. provisioning parameters
2. GA / SA version configuration
3. default version

rakesh-garimella:
Closing this PR in favor of https://github.com/kyma-project/control-plane/pull/461

PR #463, opened 2021-01-22
Implement fake edp client for tests

szwedm:
**Description**

EDP fake client has been introduced to replace EDPClient mocks in tests.

- New source file: client_fake.go, which contains fake EDP client implementation
- Mock replaced with fake client in: edp_deregistration_test.go
- Mock replaced with fake client in: edp_registration_test.go

### DIFF
--- a/components/kyma-environment-broker/internal/edp/client_fake.go
+++ b/components/kyma-environment-broker/internal/edp/client_fake.go
@@ -1,25 +1,79 @@
 package edp
 
+import "errors"
+
 // FakeClient implements the edp client interface but does not process data nor call real external system
-type FakeClient struct{}
+type FakeClient struct {
+	data map[string]interface{}
+}
 
 // NewFakeClient creates edp fake client
 func NewFakeClient() *FakeClient {
-	return &FakeClient{}
+	return &FakeClient{
+		data: make(map[string]interface{}),
+	}
 }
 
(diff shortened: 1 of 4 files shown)


components/kyma-environment-broker/internal/edp/client_fake.go
```diff
@@ -0,0 +1,79 @@
+package edp
+
+import "errors"
+
+// FakeClient implements the edp client interface but does not process data nor call real external system
+type FakeClient struct {
+	data map[string]interface{}
```
piotrmiskiewicz, line 7: you can define the struct as a type then use it in the map

piotrmiskiewicz: Imagine that I'd like to use the fake client and create two different DataTenants. Will it work? I mean can I assert it at the end?

szwedm: Good point, we will not be able to check the fields of map value since it is an anonymous struct.

piotrmiskiewicz (approved):
lgtm

piotrmiskiewicz (approved):
lgtm

piotrmiskiewicz (dismissed):

piotrmiskiewicz (approved):

PR #465, opened 2021-01-25
KEB: Implement CLS provisioning

skhalash:
**Description**

- CLS config (service manager credentials, SAML params, etc.)
- Implement a CLS provisioning/offering steps
- In-memory storage implementation
- Show-casing the usage of the above-mentioned steps in a form of an integration test

**Related issue(s)**


components/kyma-environment-broker/internal/cls/provisioner_test.go
```diff
@@ -0,0 +1,178 @@
...
+func TestProvisionCreatesNewInstanceIfNoneFoundInDB(t *testing.T) {
+	const (
+		fakeGlobalAccountID = "fake-global-account-id"
+		fakeSKRInstanceID   = "fake-skr-instance-id"
+		fakeBrokerID        = "fake-broker-id"
+		fakeServiceID       = "fake-service-id"
+		fakePlanID          = "fake-plan-id"
+	)
+
+	storageMock := &automock.InstanceStorage{}
```
piotrmiskiewicz, line 25: use in memory implementation instead of mock (we don't need mock for the storage)

skhalash: one advantage of using mocks in this case that I see is that I can test e.g. DB connection problems (when the storage returns `dberrors`)

skhalash: for instance, this test case: https://github.com/kyma-project/control-plane/blob/e691b02a17f411333bfd2dcad0579f7cc5abe7fb/components/kyma-environment-broker/internal/cls/provisioner_test.go#L79 tests that an instance is not created if it cannot be saved to the DB first

piotrmiskiewicz: for errors using a mock looks fine, for happy path - you can use in-memory

piotrmiskiewicz:
We should think about extract some code from existing offering step and reuse it in CLS offering step because it is almost the same. The main difference is that is uses different way to get SM client

skhalash:
@piotrmiskiewicz, I fully agree. I was playing around with the generic offering step and unfortunately didn't find an elegant way to make it usable for CLS. Would it be OK to unify them after we are done with step implementation in a separate PR?

components/kyma-environment-broker/internal/storage/driver/memory/cls_instance.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/storage/dberr"
+)
+
+type clsInstances struct {
+	mu sync.Mutex
+
+	data map[clsKey]internal.CLSInstance
+}
+
+type clsKey struct {
```
suleymanakbas91, line 16: I see that we're building a 1-to-1 mapping of cls instances and the global accounts. What if we are required to create more than one instance for a global account at some point in the future? Would it be too difficult to design a solution supporting more than one instance from the beginning? What do you think?

skhalash: Well, there should be a way to identify _our_ instance. Currently, we are using the `GlobalAccountID` as a unique key to search for. But you are right, we might use a combination of a `GlobalAccountID` + `Region` in the future, so I'll add a `Region` field.

skhalash: Added here: 5b64a86f089473ce2d653af54a4fa9fca965c97b

components/kyma-environment-broker/internal/cls/provisioner.go
```diff
@@ -0,0 +1,113 @@
...
+	p.log.Infof("Found existing cls instance for global account %s", request.GlobalAccountID)
+
+	if !exists {
+		p.log.Infof("No cls instance found for global account %s", request.GlobalAccountID)
+		return p.createNewInstance(smClient, request)
+	}
+
+	err = p.storage.Reference(instance.Version, instance.GlobalAccountID, request.SKRInstanceID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "while deleting a cls instance for global account %s", request.GlobalAccountID)
```
suleymanakbas91, line 70: Error message should be about referencing instead.

suleymanakbas91 (changes requested):

components/kyma-environment-broker/internal/cls/provisioner.go
skhalash: True, fixed

piotrmiskiewicz (approved):
lgtm

suleymanakbas91 (approved):

PR #474, opened 2021-01-29
Change KEB's service instance update to async

ksputo:
**Description**
When suspending/unsuspending a runtime, a deprovisioning/provisioning operation is created and process continues asynchronously. For better OSB API spec compliance and testability KEB should return the corresponding operation ID for the deprovisioning/provisioning action underneath the suspension/unsuspension.

- `PATCH /v2/service_instances/{instance_id}` changed to async behaviour according to OSB API spec 2.14

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/suspension/handler.go
+++ b/components/kyma-environment-broker/internal/suspension/handler.go
@@ -72,10 +72,15 @@ func (h *ContextUpdateHandler) suspend(instance *internal.Instance, log logrus.F
 		return "", err
 	}
 
-	// no error, deprovisioning operation exists and is in progress - fail
+	// no error, deprovisioning operation exists and is in progress
 	if err == nil && (lastDeprovisioning.State == domain.InProgress || lastDeprovisioning.State == orchestration.Pending) {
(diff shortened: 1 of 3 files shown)


components/kyma-environment-broker/internal/suspension/handler.go
```diff
@@ -63,30 +65,47 @@ func (h *ContextUpdateHandler) Handle(instance *internal.Instance, newCtx intern
...
 
-	// no error, operation exists and is in progress
+	// no error, deprovisioning operation exists and is in progress
 	if err == nil && (lastDeprovisioning.State == domain.InProgress || lastDeprovisioning.State == orchestration.Pending) {
-		log.Infof("Suspension already started")
-		return nil
+		if !lastDeprovisioning.Temporary {
+			// found ordinary deprovisioning operation in progress - fail suspension
+			log.Warnf("Deprovisioning operation already started for instance %s", instance.InstanceID)
+			return "", errors.New("cannot suspend instance - deprovisioning operation already started")
```
piotrmiskiewicz, line 80: I think we should not return an error here. When the instance is being deprovisioned we should not do anything. The context update response could be HTTP 200

PR #478, opened 2021-02-01
Allow retriggering runtime suspension

ebensom:
**Description**

Only skip context update  during unsuspension, but allow retriggering suspension, as it is needed to reprocess failed deprovisionings.

### DIFF
--- a/components/kyma-environment-broker/internal/suspension/handler.go
+++ b/components/kyma-environment-broker/internal/suspension/handler.go
@@ -51,7 +51,7 @@ func (h *ContextUpdateHandler) Handle(instance *internal.Instance, newCtx intern
 		isActivated = *instance.Parameters.ErsContext.Active
 	}
 
-	if newCtx.Active == nil || isActivated == *newCtx.Active {
+	if newCtx.Active == nil || isActivated && *newCtx.Active {
 		logrus.Debugf("Context.Active flag was not changed, the current value: %v", newCtx.Active)
 		return nil
 	}
--- a/resources/kcp/values.yaml
+++ b/resources/kcp/values.yaml
@@ -12,7 +12,7 @@ global:
       version: "PR-435"
     kyma_environment_broker:
       dir:
-      version: "PR-477"
+      version: "PR-478"
     kyma_environments_cleanup_job:
       dir:
       version: "PR-416"
--- a/resources/kcp/values.yaml
+++ b/resources/kcp/values.yaml
@@ -12,7 +12,7 @@ global:
       version: "PR-435"
     kyma_environment_broker:
       dir:
-      version: "PR-466"
+      version: "PR-477"
     kyma_environments_cleanup_job:
       dir:
       version: "PR-416"


components/kyma-environment-broker/internal/suspension/handler.go
```diff
@@ -51,7 +51,7 @@ func (h *ContextUpdateHandler) Handle(instance *internal.Instance, newCtx intern
 		isActivated = *instance.Parameters.ErsContext.Active
 	}
 
-	if newCtx.Active == nil || isActivated == *newCtx.Active {
+	if newCtx.Active == nil || isActivated && *newCtx.Active {
```
ksputo, line 54: I don't think this solution is right as it will allow to call suspension on already suspended instance. Consider following unit test:
```golang
func TestSuspensionRetrigger(t *testing.T) {
	// given
	provisioning := NewDummyQueue()
	deprovisioning := NewDummyQueue()
	st := storage.NewMemoryStorage()

	svc := NewContextUpdateHandler(st.Operations(), provisioning, deprovisioning, logrus.New())
	instance := fixInstance(fixInactiveErsContext())
	st.Instances().Insert(*instance)
	st.Operations().InsertDeprovisioningOperation(internal.DeprovisioningOperation{
		Operation:       internal.Operation{
			ID:                     "suspended-op-id",
			Version:                0,
			CreatedAt:              time.Now(),
			UpdatedAt:              time.Now(),
			InstanceID:             instance.InstanceID,
			State: domain.Succeeded,
		},
		Temporary:       true,
	})

	// when
	err := svc.Handle(instance, fixInactiveErsContext())
	require.NoError(t, err)

	// then
	op, _ := st.Operations().GetDeprovisioningOperationByInstanceID("instance-id")
	assertQueue(t, deprovisioning, op.ID)
	assertQueue(t, provisioning)

	assert.Equal(t, domain.LastOperationState("pending"), op.State)
	assert.Equal(t, instance.InstanceID, op.InstanceID)
}
```

so, we have an instance in db with `Context.Active: false` - this happens whenever we successfully call context update (no matter if the deprovisioning operation underneath fails or not). Moreover, we have a temporary deprovisioning operation, which succeeded (which means that cluster was deprovisioned as a result of context update - suspension) - with this change, KEB will update the context again (from `Active: false` -> `Active: false`, no change here), but will also create another deprovisioning operation and which will finally fail (or worst, hang in queue) because we will try to call Provisioner with deprovision mutation, taking `runtimeID` from this suspended instance - and this `runtimeID` in such case would be empty as we remove `runtimeID` when deprovisioning succeeded with `operation.Temporary == true`.

I'd rather consider additional logic for checking if we should retrigger the operation in case another context update call comes (with the same `Active` value as current instance state.

ksputo: See my proposal here:

https://github.com/kyma-project/control-plane/pull/479

cc: @piotrmiskiewicz

PR #479, opened 2021-02-01
Add retriggering runtime suspension

ksputo:
**Description**
Currently updating context with the same `Active` value as the one in `instance.Parameters.ErsContext.Active` in database results in skipping the update. However, `instance.Parameters.ErsContext.Active` is always updated if there is no error here:

https://github.com/kyma-project/control-plane/blob/b5e8b6e1163a47c6e819895af45bb3204d11df3c/components/kyma-environment-broker/internal/broker/instance_update.go#L69

`Handle()` method does not check, if the deprovisioning operation created underneath succeeded or failed in the process, and afterwards the instance is updated anyways which prevents from retriggering the failed suspension call when user tries updating context again (instance state in DB is `Active: false` and incoming call also contains `Active: false` parameter)

- Check if there is an existing temporary operation when `Context.Active` flag was not changed - if it exists and is in state `failed`, retrigger the deprovisioning process

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/suspension/handler.go
+++ b/components/kyma-environment-broker/internal/suspension/handler.go
@@ -46,14 +46,36 @@ func (h *ContextUpdateHandler) Handle(instance *internal.Instance, newCtx intern
 		return nil
 	}
 
+	return h.handleContextChange(newCtx, instance, l)
+}
+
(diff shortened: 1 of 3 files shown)


piotrmiskiewicz (approved):
lgtm

piotrmiskiewicz (approved):

PR #485, opened 2021-02-02
KEB reverts erscontext cleanup

piotrmiskiewicz:
**Description**

- the context update (http PATCH) is a partial update. First version was taking all values so we have cleaned values.
- this is the reason we need to take provisioning parameters from last provisioning operation

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/broker/instance_update_test.go
+++ b/components/kyma-environment-broker/internal/broker/instance_update_test.go
@@ -1,0 +1,82 @@
+package broker
+
+import (
+	"context"
+	"encoding/json"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/ptr"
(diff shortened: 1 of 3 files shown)


components/kyma-environment-broker/internal/broker/instance_update_test.go
```diff
@@ -0,0 +1,83 @@
...
+		RawContext:      json.RawMessage("{\"active\":false}"),
+		MaintenanceInfo: nil,
+	}, true)
+
+	// then
+	inst, _ := st.Instances().GetByID(instanceID)
+	// check if original ERS context is set again in the Instance entity
+	assert.NotEmpty(t, inst.Parameters.ErsContext.ServiceManager.Credentials.BasicAuth.Password)
+	// check if the handler was called
+	assert.Equal(t, instance, handler.Instance)
```
ksputo, line 79: This assertion throws error, because you are expecting fixed instance with empty ERS Context with the one processed by `Handle()` method which has modified ERS Context (with SM credentials)

ksputo (approved):
lgtm

PR #487, opened 2021-02-03
Migrate OIDC JWT verification from Istio to ORY oathkeeper

ebensom:
**Description**

The new Istio auth model (`RequestAuthentication`) doesn't support workloads relying on hybrid auth strategies, i.e. using oauth2 access token for certain endpoints, while using OIDC ID token in JWT format for other endpoints.


piotrmiskiewicz (approved):
lgtm

PR #494, opened 2021-02-05
Add admin group to Oathkeeper rule to list runtimes

lumi017:
**Description**

Add admin group to Oathkeeper rule to list runtimes

**Related issue(s)**

### DIFF
--- a/resources/kcp/charts/kyma-environment-broker/templates/oathkeeper-rule.yaml
+++ b/resources/kcp/charts/kyma-environment-broker/templates/oathkeeper-rule.yaml
@@ -43,7 +43,7 @@ spec:
     config:
       jwks_urls: ["{{ tpl .Values.oidc.keysURL $ }}"]
       scope_strategy: exact
-      required_scope: ["{{ .Values.oidc.groups.operator }}"]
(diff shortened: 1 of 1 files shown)


resources/kcp/charts/kyma-environment-broker/templates/oathkeeper-rule.yaml
```diff
@@ -43,7 +43,7 @@ spec:
     config:
       jwks_urls: ["{{ tpl .Values.oidc.keysURL $ }}"]
       scope_strategy: exact
-      required_scope: ["{{ .Values.oidc.groups.operator }}"]
+      required_scope: ["{{ .Values.oidc.groups.operator }}", "{{ .Values.oidc.groups.admin }}"]
```
ebensom, line 46: Unfortunately this is not going to work as intended: the rule will require both scopes to be present in and claim, so he relationship is AND and not OR.
We have the solution but it will require changing L2 and L3 CAM profiles. Just FYI I'm on it.

lumi017: Yes indeed it's not so straight forward, and you can't create a second rule matching the same URL, so changing CAM profile is the only option...

PR #497, opened 2021-02-08
Fix suspension when last deprovisioning not exist

ksputo:
**Description**
When last deprovisioning operation does not exist we return `nil`, which can cause crash

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/suspension/handler.go
+++ b/components/kyma-environment-broker/internal/suspension/handler.go
@@ -69,6 +69,10 @@ func (h *ContextUpdateHandler) handleContextChange(newCtx internal.ERSContext, i
 			if err != nil && !dberr.IsNotFound(err) {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


components/kyma-environment-broker/internal/suspension/handler.go
```diff
@@ -69,6 +69,10 @@ func (h *ContextUpdateHandler) handleContextChange(newCtx internal.ERSContext, i
 			if err != nil && !dberr.IsNotFound(err) {
 				return err
 			}
+			if dberr.IsNotFound(err) || lastDeprovisioning == nil {
```
piotrmiskiewicz, line 72: I would do simpler: ```if lastDeprovisioning == nil {```

piotrmiskiewicz (changes requested):

piotrmiskiewicz (approved):

piotrmiskiewicz (approved):

PR #504, opened 2021-02-09
Overrides Conflict Resolution Strategy

ralikio:
**Description**

Introduce merge or replace flags to allow differ between algorithms used for override resources merge.

- Addition of onConflict flag to GraphQl mutations for merging/replacing config maps and secrets.

**Related issue(s)**


components/provisioner/pkg/gqlschema/schema.graphql
```diff
@@ -212,6 +213,7 @@ input ComponentConfigurationInput {
     namespace: String!                    # Namespace to which component should be installed
     configuration: [ConfigEntryInput]     # Component specific configuration
     sourceURL: String                     # Custom URL for the source files of the given component
+    onConflict: String                            # Defines merging strategy if conflicts occur for component overrides
```
akgalwas, line 216: onConflict field could be defined as Enum type (similar to e.g. KymaProfile). It makes sense as there are two possible values and you get validation for free. Here is the error message I got whn I specified incorrect KymaProfile value: "Expected type KymaProfile, found Evaluationdd. Did you mean the enum value Evaluation?"

akgalwas (approved):

akgalwas (approved):

PR #473, opened 2021-01-29
Hibernation test resilence

akgalwas:
**Description**
The general idea is to make Hibernation test more resilient.

Changes proposed in this pull request:

- Environments Clean Up job removes clusters provisioned with Provisoner
- Hibernation test timeouts refactoring

PR #505, opened 2021-02-10
Enable PERMISSIVE mTLS for provisioner for all endpoints

ebensom:
**Description**

#473 introduced a DestionationRule for provisioner to disable mTLS for downstream peers of provisioner. It results in an error for peers within the Istio mesh as provisioner's istio-proxy is configured with STRICT TLS settings (i.e. not allow clients with plain-text HTTP connection). 
To fix this issue while still allowing KEB environments-cleanup job to reach provisioner's GraphQL endpoint, provisioner's istio-proxy ios configured in PERMISSIVE mode which will allow both TLS and plain-text connections from within the cluster.


akgalwas (approved):

kjaksik:
It won't pass, there's a pr like [this](https://github.com/kyma-project/test-infra/pull/3275) in the works but for the integration job

ksputo (approved):

Issue #428, opened 2021-01-08
Refactor storage tests in KEB


polskikiel:
**Description**

What can be done:

- split storage_test.go to separate files so e.g. 
   - `operation_test.go`
   - `instance_test.go`
   - `lmsTenant_test.go`
   - ...
- place above tests in a separate package // maybe `internal/storage/driver/postsql` is a good place for this kind of tests
- fulfil every structures' fields with a test data and assert if it was propagated through the storage
- fixtures can be moved to the separate place `fixtures.go` 
- think about splitting tests to a smaller parts - test only the particular functions without having something like e2e storage test for operations

**AC**
- Tests covers all scenarios, more assertions etc
- Developer can run single test (without commenting out other tests etc)
- No single monolith test file 
- Decrease test execution from 2 min to 30 seconds

ksputo:
Suggestion: we could move fixtures next to the model in code, instead of creating separate fixtures in each place where for example `Instance` is needed

szwedm:
Proposed changes have been implemented. Closing.

PR #510, opened 2021-02-11
Fixtures for storage_test

szwedm:
**Description**

storage_test.go in KEB contains functions, which are creating fixed structures required for tests. Some fields of the structures are left with zero values, what could lead to a missed test case. Similar pattern occurs in a lot of other KEB components tests, so it would be a good practice to place fixed structure creation functions in a separate file, where every KEB component test could invoke creational function from one place. This favors code reuse and eliminates redundancy.

- implement widely used structures in tests as fixtures with fields populated with default values,
- keep fixtures in one source file, which provides default structure creation functions.

**Related issue(s)**
See #428.


components/kyma-environment-broker/internal/storage/storage_test.go
```diff
@@ -1134,104 +1132,44 @@ func fixInstance(testData instanceData) *internal.Instance {
...
-		},
-		PlatformRegion: "region",
-	}
+	pp := internal.FixProvisioningParameters("test")
+	pp.PlanID = broker.TrialPlanID
+	pp.ServiceID = broker.KymaServiceID
+	return pp
 }
 
 func fixInstanceDetails() internal.InstanceDetails {
```
polskikiel, line 1175: `fixInstanceDetails` can also be moved to fixture.go

components/kyma-environment-broker/internal/fixture.go
```diff
@@ -0,0 +1,257 @@
...
+
+func FixProvisioningParametersDTO() ProvisioningParametersDTO {
+	trialCloudProvider := TrialCloudProvider("provider")
+	return ProvisioningParametersDTO{
+		Name:                        "cluster-name",
+		TargetSecret:                pointer.StringPtr("TargetSecret"),
+		VolumeSizeGb:                ptr.Integer(50),
+		MachineType:                 pointer.StringPtr("MachineType"),
+		Region:                      pointer.StringPtr("Region"),
+		Purpose:                     pointer.StringPtr("Purpose"),
```
polskikiel, line 70: why do we use a new pointer package? you can just use String function from our ptr package

polskikiel (changes requested):
Some comments from my side, do you plan to use common fixtures also in the other KEB's packages? As I suppose it will be added in the next PRs? 
I was also thinking about putting fixtures to a separate package e.g. `fixture/fixture.go` to have it under different name - `fixture.FixDeprovisioningOperation(operationId, testData)` instead of `internal.FixDeprovisioningOperation(operationId, testData)`. Please let me know what do you think.

szwedm:
> Some comments from my side, do you plan to use common fixtures also in the other KEB's packages? As I suppose it will be added in the next PRs?
That's correct, I'm currently replacing fixtures implemented locally in test files with new ones. This will be visible in next PR.
> I was also thinking about putting fixtures to a separate package e.g. `fixture/fixture.go` to have it under different name - `fixture.FixDeprovisioningOperation(operationId, testData)` instead of `internal.FixDeprovisioningOperation(operationId, testData)`. Please let me know what do you think.
Brilliant, I will put fixtures in fixture package.

components/kyma-environment-broker/internal/fixture.go
szwedm: Good point, forgot to change that.

components/kyma-environment-broker/internal/storage/storage_test.go
szwedm: Will do.

polskikiel (dismissed):

polskikiel (approved):

PR #513, opened 2021-02-12
Decrease retries count in calls to Director

akgalwas:
**Description**

The idea behind that is to align timeouts in KEB with timeouts in Provisioner. 

- Decreased number of retries in calls to Director


koala7659 (approved):
This is OK and should mitigate the possible problem. I think we could improve the way we handle errors from graphQL calls to Director to retry only when it makes sense to do that. I will log an issue for that

koala7659:
I have created follow up issue to process errors from Director RegisterRuntime call. https://github.com/kyma-project/control-plane/issues/517

PR #514, opened 2021-02-15
Encrypt instance parameters

polskikiel:
**Description**

- Encrypted instance parameters with migration

**Related issue(s)**


components/kyma-environment-broker/internal/migrations/instance_parameters_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"testing"
+
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/migrations"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/storage"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInstanceParametersMigration_Migrate(t *testing.T) {
```
kjaksik, line 13: What do you think about adding a check after this migration that would check if the auth is encrypted?

polskikiel: ok so i have changed the test to use postgres storage and added legacy methods that allows me to write a properly working migration along with the unit test

kjaksik (approved):

PR #541, opened 2021-02-22
Adjust catalog metadata

polskikiel:
**Description**

- Changed service image to base64 encoded svg file
- Use separate yaml file mounted using config map to provide data about KEB's services and plans
- Added pattern validation for cluster name allowing for all alphanumeric strings with '-'

**Related issue(s)**


resources/kcp/charts/kyma-environment-broker/values.yaml
```diff
@@ -81,7 +81,23 @@ osbUpdateProcessingEnabled: "false"
 
 brokerService:
   displayName: "Kyma Environment"
```
PK85, line 83: please offer some solution that we can override any field frm schema by yaml, e.g.:

```
services.ID.description
services.ID.metadata.longDescription
plans.ID.description
plans.ID.metadata.displayName
```

components/kyma-environment-broker/internal/broker/services.go
```diff
@@ -33,8 +35,10 @@ func NewServices(cfg Config, log logrus.FieldLogger) *ServicesEndpoint {
 //   GET /v2/catalog
 func (b *ServicesEndpoint) Services(ctx context.Context) ([]domain.Service, error) {
 	var availableServicePlans []domain.ServicePlan
+	// we scope to the kymaruntime service only
+	class := b.servicesConfig[KymaServiceName]
```
jasiu001, line 39: if the key `kymaruntime` does not exist in catalog.yaml (or has e.g. type) you will not get any error message (errror log) and plans will not contain description/metadata info.

components/kyma-environment-broker/internal/broker/broker.go
```diff
@@ -26,18 +30,56 @@ type KymaEnvironmentBroker struct {
...
 
+type ServicesConfig map[string]Service
+
+func NewServicesConfigFromFile(path string) (ServicesConfig, error) {
+	yamlFile, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, errors.Wrap(err, "while reading YAML file with managed components list")
+	}
+	fmt.Println("XD")
+	fmt.Println(string(yamlFile))
```
jasiu001, line 45: it is not necessary I think ;)

jasiu001 (approved):

PR #547, opened 2021-02-23
Add eventing as optional component

radufa:
**Description**

- Add eventing as a new optional component

**Related issue(s)**


radufa:
For SKR, the new Eventing will not be an optional component.

PR #550, opened 2021-02-24
KEB: Implement SQL storage for CLS

skhalash:
**Description**

- Wrote a SQL implementation of  the `CLSInstances` interface (the one that was used by far was in-memory)
- Adjusted the `CLSInstances` interface based on findings from the SQL implementation
- Added an integration test, which tests the new storage implementation with a real Postgre DB
- Added migration scripts

**Related issue(s)**


suleymanakbas91 (approved):

components/kyma-environment-broker/internal/cls/deprovisioner.go
```diff
@@ -35,15 +35,14 @@ func NewDeprovisioner(storage DeprovisionerStorage, remover InstanceRemover, log
...
+	SKRInstanceID string
+	Instance      servicemanager.InstanceKey
 }
 
 func (d *Deprovisioner) Deprovision(smClient servicemanager.Client, request *DeprovisionRequest) error {
-	instance, _, err := d.storage.FindInstance(request.GlobalAccountID)
+	instance, _, err := d.storage.FindByID(request.Instance.InstanceID)
 	if err != nil {
-		return errors.Wrapf(err, "while trying to lookup an instance for global account: %s", request.GlobalAccountID)
+		return errors.Wrapf(err, "while trying to find the cls instance %s", request.Instance.InstanceID)
```
piotrmiskiewicz, line 45: if the CLS instance is not found, do not return any error. Maybe we can log it, but it would be nice to call "Deprovision" method twice and do not see any error.

components/kyma-environment-broker/internal/model.go
```diff
@@ -35,13 +36,109 @@ const (
...
+	version         int
+	id              string
+	globalAccountID string
+	region          string
+	createdAt       time.Time
+	references      []string
+	removedBy       string
+
+	events []interface{}
+}
```
piotrmiskiewicz, line 48: I know everything is here (in this file) but maybe let's move CLSInstance to another go file

components/kyma-environment-broker/internal/model.go
```diff
@@ -35,13 +36,109 @@ const (
...
+type CLSInstanceReferencedEvent struct {
+	SKRInstanceID string
+}
+
+type CLSInstanceUnreferencedEvent struct {
+	SKRInstanceID string
+}
+
+func NewCLSInstance(version int, id, globalAccountID, region string, createdAt time.Time, references []string, removedBy string) *CLSInstance {
```
piotrmiskiewicz, line 57: We need a constructor which could be used to create a "fresh" (completely new) CLSInstance. So the only arguments list is: id (it can be also generated inside the constructor), GAID, region, SKRInstanceID. All other fields you can fill, let's say:
```
func NewCLSInstance(globalAccountID string, region string, skrInstanceID string) CLSInstance {
return CLSInstance{
  version: 0,
  id: uuid.New(),
  globalAccountID: globalAccountID,
region: region,
createdAt: time.New(),
references: []string{skrInstanceId},
removedBy: "",
}
}

```

piotrmiskiewicz: I know the storage needs another constructor - you can have the second one (find a good name)

skhalash: Agree, I refactored it completely to use function options, which seems to be a good fit here.

components/kyma-environment-broker/internal/model.go
skhalash: Yeah, makes sense. Will do it

components/kyma-environment-broker/internal/cls/deprovisioner.go
skhalash: I'll note it down, but I'd rather tackle it in our final PR (since we already noticed some other similar issues in other places)

components/kyma-environment-broker/internal/model.go
skhalash: Done

piotrmiskiewicz (approved):

PR #556, opened 2021-02-26
Add download kubeconfig for SKR functionality

jasiu001:

- KEB during provisioning initiation returns URL to kubeconfig for future SKR
- when SKR is ready, a new endpoint will return properly prepared kubeconfig


components/kyma-environment-broker/internal/kubeconfig/handler.go
```diff
@@ -0,0 +1,131 @@
...
+		return
+	}
+
+	if operation.InstanceID != instanceID {
+		h.handleErrorResponse(w, http.StatusBadRequest, errors.New("mismatch between operation and instance"))
+		return
+	}
+
+	switch operation.State {
+	case domain.InProgress:
```
piotrmiskiewicz, line 92: what about pending state?

components/kyma-environment-broker/internal/kubeconfig/handler.go
```diff
@@ -0,0 +1,131 @@
...
+		h.handleInfoResponse(w, fmt.Sprintf("provisioning operation for instance %s is in progress state, kubeconfig not exist yet, please try again later.", instanceID))
```
piotrmiskiewicz, line 93: if the kubeconfig is not available - please return http 404

components/kyma-environment-broker/internal/kubeconfig/handler.go
```diff
@@ -0,0 +1,131 @@
...
+	vars := mux.Vars(r)
+	instanceID := vars["instance_id"]
+	operationID := vars["operation_id"]
+
+	if !h.isValidUUID(operationID) {
+		h.handleErrorResponse(w, http.StatusBadRequest, errors.New("wrong operationID"))
+		return
+	}
+
+	if !h.isValidUUID(instanceID) {
```
piotrmiskiewicz, line 64: KEB allows the instance ID to be not valid UUID, so why do we need to validate it here?

piotrmiskiewicz:
please, update the swagger or create a task for that and then swagger will be updated in another PR

components/kyma-environment-broker/internal/kubeconfig/handler.go
```diff
@@ -0,0 +1,131 @@
...
+	return &Handler{
+		instanceStorage:   storage.Instances(),
+		operationStorage:  storage.Operations(),
+		kubeconfigBuilder: b,
+		log:               log,
+	}
+}
+
+func (h *Handler) AttachRoutes(router *mux.Router) {
+	router.HandleFunc("/kubeconfig/{operation_id}/instance/{instance_id}", h.GetKubeconfig).Methods(http.MethodGet)
```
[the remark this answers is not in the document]
PK85: And instanceId could have many related operationIDs. What is a reason of having operationID here at all?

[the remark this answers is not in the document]
jasiu001: only instanceID will be present in the URL, so the path will be `/kubeconfig/{instance_id}`

components/kyma-environment-broker/internal/kubeconfig/handler.go
jasiu001: IMO `Pending` is a status of orchestration which is part of the update operation, during the update cluster already exist so kubeconfig could be downloaded. There is more [update statuses](https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/common/orchestration/dto.go#L29) which I'm not supported. BTW this is a little mess to have statuses in two different places (domain from pivotal and orchestration from KEB)

components/kyma-environment-broker/internal/kubeconfig/handler.go
jasiu001: right, I forgot we support not uuid type of ID

components/kyma-environment-broker/internal/kubeconfig/handler.go
piotrmiskiewicz: We have introduced a Pending state which indicates that the operation didn't really started. The GetProvisioningOperationByInstanceID returns the latest provisioning operation which could be triggered by unsuspension. Such operation has state Pending until every other operation has finished.
The Pending state does not match OSB API and we are not able to define it in the same place as others states.

piotrmiskiewicz (dismissed):

components/kubeconfig-service/pkg/transformer/kubeconfig.go
```diff
@@ -49,6 +49,5 @@ users:
       - get-token
       - "--oidc-issuer-url={{ .OIDCIssuerURL }}"
       - "--oidc-client-id={{ .OIDCClientID }}"
-      - "--oidc-client-secret={{ .OIDCClientSecret }}"
```
ebensom, line 52: Feel free to remove the corresponding config item from the chart, and https://github.com/kyma-project/control-plane/blob/main/components/kubeconfig-service/pkg/env/env.go

piotrmiskiewicz (approved):

PR #557, opened 2021-03-01
add gardener quota and azure secret usage tables in grafana dashboard

lauraanddola:
to implement the task [Grafana Dashboard for Gardener Usage Exporter](https://github.tools.sap/kyma/backlog/issues/1281)

### DIFF
--- a/resources/kcp/charts/kyma-environment-broker/files/dashboard.json
+++ b/resources/kcp/charts/kyma-environment-broker/files/dashboard.json
@@ -18,6 +18,7 @@
   "links": [],
   "panels": [
     {
+      "collapsed": true,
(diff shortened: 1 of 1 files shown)


polskikiel (approved):

ebensom (approved):

lumi017:
I think it's better if we don't tie gardener exporter metrics to KCP dashboard. IMO it should be a new dashboard that we inject to Grafana.

lauraanddola:
> I think it's better if we don't tie gardener exporter metrics to KCP dashboard. IMO it should be a new dashboard that we inject to Grafana.

I plan to create a new dashboard and move those to new one.

PR #559, opened 2021-03-01
New provisioning process approach (Do not merge!)

jasiu001:
**Description**

This is a proposal for the new provisioning process. Only for tests!


jasiu001:
concept implemented in [this PR](https://github.com/kyma-project/control-plane/pull/663)

PR #569, opened 2021-03-03
Orchestration of cluster (shoot) upgrades, part 1

ebensom:
**Description**

- Model for `upgradeCluster` orchestrations and operations
- Storage changes for `upgradeCluster` orchestrations and operations
- Reusable generic orchestration manager. Define `upgradeKymaManager` and `upgradeClusterManager` in terms of orchestration manager. 

**Related issue(s)**
https://github.tools.sap/kyma/backlog/issues/1165


components/kyma-environment-broker/internal/orchestration/handlers/cluster_handler.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// defaults strategy if not specified to Parallel with Immediate schedule
+	defaultOrchestrationStrategy(&params.Strategy)
+
+	now := time.Now()
+	o := internal.Orchestration{
+		OrchestrationID: uuid.New().String(),
+		Type:            orchestration.UpgradeClusterOrchestration,
+		State:           orchestration.Pending,
+		Description:     "started processing of Kyma upgrade",
```
polskikiel, line 67: I know it's not really needed but we can change Kyma to Cluster or we can just use different description like `added to queue` for both types

polskikiel (approved):

PR #580, opened 2021-03-09
Remove clusterbrokers finalizers before uninstallation

polskikiel:
**Description**

- Remove clusterbrokers finalizers before uninstallation to speed up the process

**Related issue(s)**

### DIFF
--- a/components/provisioner/internal/installation/resource_cleanup.go
+++ b/components/provisioner/internal/installation/resource_cleanup.go
@@ -91,6 +91,7 @@ func (s *serviceCatalogClient) PerformCleanup(resourceSelector string) error {
 	}
 
 	s.deleteServiceInstances(serviceInstances)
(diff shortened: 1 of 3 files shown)


components/provisioner/internal/installation/resource_cleanup.go
```diff
@@ -231,6 +232,24 @@ func (s *serviceCatalogClient) deleteServiceInstances(serviceInstances []v1beta1
 	}
 }
 
+func (s *serviceCatalogClient) removeClusterBrokersFinalizers(brokers []v1beta1.ClusterServiceBroker) {
+	for _, broker := range brokers {
+		logrus.Debugf("trying to delete ClusterServiceBroker %q", broker.Name)
```
ksputo, line 237: this log line might be confusing, because we do not try to delete `ClusterServiceBrokers` here, just clearing their finalizers.

ksputo (approved):
lgtm

PR #590, opened 2021-03-12
Fix deprovisioning of suspended trial runtimes

ebensom:
**Description**

If a deprovision is triggered after suspension (so without having the runtime unsuspended first), KEB doesn't clean up the instance from DB properly. As a consequence the subsequent provisioning requests always fail with `The Trial Kyma was created for the global account, but there is only one allowed` error message.

**Related issue(s)**
https://github.tools.sap/kyma/backlog/issues/1301

### DIFF
--- a/components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
+++ b/components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
@@ -114,6 +114,10 @@ func (s *InitialisationStep) run(operation internal.DeprovisioningOperation, log
 	instance, err := s.instanceStorage.GetByID(operation.InstanceID)
 	switch {
 	case err == nil:
+		// If the instance is currently suspended, let this deprovision operation clean up the instance
+		if instance.Parameters.ErsContext.Active != nil && !*instance.Parameters.ErsContext.Active {
+			return s.operationManager.OperationSucceeded(operation, "instance has been suspended previously", log)
+		}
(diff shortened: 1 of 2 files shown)


components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
```diff
@@ -114,6 +114,10 @@ func (s *InitialisationStep) run(operation internal.DeprovisioningOperation, log
 	instance, err := s.instanceStorage.GetByID(operation.InstanceID)
 	switch {
 	case err == nil:
+		// If the instance is currently suspended, let this deprovision operation clean up the instance
+		if instance.Parameters.ErsContext.Active != nil && !*instance.Parameters.ErsContext.Active {
```
piotrmiskiewicz, line 118: in general it should work. I'm afraid if the suspension failed (something was not removed) - such resource will be orphaned.

ebensom: As far as I can see on just need to add the `&& instance.RuntimeID == ""` to the condition, because if it's empty, it means that the runtime was either successfully removed before during suspension, or never existed (due to some failure). If the `RuntimeID` is not empty, the operation could proceed until `Remove_Runtime`, triggering new `deleteRuntime` in provisioner and polling the status. 

What do you think?

ebensom:
Abandon in favor of https://github.com/kyma-project/control-plane/pull/595

PR #601, opened 2021-03-17
Cluster upgrade orchestration, part 2

ebensom:
**Description**

- Implement cluster (Shoot) upgrade process steps, manager and operation data manager
- Refactor AVS evaluation manager to make it independent from Kyma Upgrade operations, and reuse it in cluster upgrade process
- Moved `OperationType` from `dbmodel` to internal, expose `OperationType` in `Operation` to better check for concurrent operation(s) and their types.
- refactor kyma upgrade initialization:
    - Fix to NOT CANCEL in progress operation when the orchestration is canceled
    - Eliminate instance use
    - eliminate orchestration use in kyma version initialization from parameter
    - factor out faulty rescheduling behavior (see below)
    - Preempt upgrade operation by deprovisioning (suspension) instead of waiting for upgrade to finish
    - behavior when checking concurrent operations on the same runtime
- Refactored maintenance window rescheduling logic to be in parallel strategy, used commonly by kyma and cluster upgrade orchestration
- Allow concurrent upgrade orchestrations, while making sure concurrent operations are not in progress on the same runtime. Eliminate 1-length orchestration queues restriction. See https://github.tools.sap/kyma/backlog/issues/1084
- in-memory test for E2E cluster upgrade orchestration

**Related issue(s)**
https://github.tools.sap/kyma/backlog/issues/1165
https://github.tools.sap/kyma/backlog/issues/1084


components/kyma-environment-broker/internal/model.go
```diff
@@ -122,14 +123,31 @@ type Instance struct {
...
 
-	ID        string    `json:"-"`
-	Version   int       `json:"-"`
-	CreatedAt time.Time `json:"-"`
-	UpdatedAt time.Time `json:"-"`
+	ID        string        `json:"-"`
+	Version   int           `json:"-"`
+	CreatedAt time.Time     `json:"-"`
+	UpdatedAt time.Time     `json:"-"`
+	Type      OperationType `json:"-"`
```
piotrmiskiewicz, line 150: why did you add a type to the model? when you have an instance of ProvisioningOperation you know the type is "Provisioning", if you have an instance of DeprovisioningOperation - type is "deprovisioning". where are you using the type?

components/kyma-environment-broker/internal/orchestration/manager/upgrade_kyma.go
```diff
@@ -60,6 +61,9 @@ func (u *upgradeKymaFactory) NewOperation(o internal.Orchestration, r orchestrat
 		},
 		SMClientFactory: u.smcf,
 	}
+	if o.Parameters.Kyma.Version != "" {
```
piotrmiskiewicz, line 64: why do we need this? let's discuss

piotrmiskiewicz: maybe we have to compute the version according to the GA/SA->version mapping and the default one.

piotrmiskiewicz: ok, I see the reason. It should be nice to cover the scenario with manual version set in the orchestration by the orchestration test. let's discuss it

components/kyma-environment-broker/internal/model.go
ebensom: In KEB model, the instance current state is represented in the last operation's current state. There are already some places where we check for concurrent operations to guard against certain conditions - these places actually meant to check the "current instance state".

Now with upgrade kyma and upgrade cluster, there are too much different operation types, and to get the "current instance state", it would be very cumbersome to implement proper conditions for ensuring concurrent operations are not performed in parallel on the same runtime with looking up each (5-6) different operation type specific data from DB and compare. A much natural and easier way is to use the `GetLastOperation()` method, and use the `Operation` object's `Type` and `State` .

I use it mainly [here](https://github.com/kyma-project/control-plane/pull/601/files#diff-53706f0b7a9c8887c58d6376e430c426578722a233dec634591344b110e02cceR70) and [here](https://github.com/kyma-project/control-plane/pull/601/files#diff-69186aec10682e7b7131a19426024d2a79d81cceae25a9b405c2dbdfda0d8e3eR84)

components/kyma-environment-broker/internal/orchestration/manager/upgrade_kyma.go
ebensom: OK, will create a test case for that scenario

ebensom: The main reason for this change is that I wanted to eliminate orchestration usage in upgrade operation's initialization step. It is a bad pattern introduced lately in by my colleague, and wanted to eliminate that. Operation data MUST enclose all data needed to perform the operation, it should not "fetch" some shadow data from parent orchestration.

components/kyma-environment-broker/internal/process/input/builder.go
```diff
@@ -291,3 +292,58 @@ func mergeMaps(maps ...map[string]struct{}) map[string]struct{} {
 	}
 	return res
 }
+
+func (f *InputBuilderFactory) CreateUpgradeShootInput(pp internal.ProvisioningParameters) (internal.ProvisionerInputCreator, error) {
+	if !f.IsPlanSupport(pp.PlanID) {
+		return nil, errors.Errorf("plan %s in not supported", pp.PlanID)
+	}
+
+	provider, err := f.getHyperscalerProviderForPlanID(pp.PlanID, pp.Parameters.Provider)
```
piotrmiskiewicz, line 301: the provider could be changed for a plan (see Trial). Do you expect provider change? what is the expectation of upgradeshoot with  a change of the provider?

components/kyma-environment-broker/internal/orchestration/manager/upgrade_kyma.go
ebensom: Scenerio covered in a new orchestration test case in latest commit

piotrmiskiewicz (approved):

PR #610, opened 2021-03-23
Upgrade containerd dependency to version 1.4.4 for provisioner and provisioner tests

koala7659:
The reason for a change is to mitigate CVE-2021-21334


franpog859 (approved):
We don't have Provisioner's `go.sum` on the master branch 🤔🤔🤔 I'm curious what happened here 

LGTM anyway

PR #563, opened 2021-03-02
KEB: Combine CLS steps

skhalash:
**Description**

It's the final PR in a series of CLS steps, which combines them all together and enabled them in `main.go` (still hidden behind a feature flag)

Other changes proposed in this pull request:

- Fix FluentBit config indentation
- Pass the logger as a method argument (instead of a constructor argument) 
- Add all Azure regions in the region mapping routine
- Disable CLS steps for trial clusters 

**Related issue(s)**
See also #550  #536 #465 #565

PR #614, opened 2021-03-25
KEB: Improve CLS stability

skhalash:
After a rollout to Canary, we figured out that CLS steps didn't handle certain Service Manager errors gracefully.

**Description**

- Implement a separate step that checks provisioning status and deletes an instance if provisioning fails
- Add an integration test with real provisoning/deprovisioning managers, processing queues in 2 flavours (with a mock service manager client and with a real one)
- Improve stability by introducing retries where necessary
- Fix a problem with stuck deprovisioning

**Related issue(s)**
See also #563


piotrmiskiewicz (approved):
let's go!

PR #633, opened 2021-04-01
New way of setting audit logging tenant

Maladie:
**Description**

- setting audit log tenant is now based on seed region

**Related issue(s)**


components/provisioner/internal/gardener/auditlogging.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func getAuditLogInstanceIdentifier(seed gardener_types.Seed) string {
+	message := findAuditLogConditionMessage(seed)
+
+	if message == "" {
+		return ""
+	}
+
+	pat := regexp.MustCompile(audiInstanceCodePattern)
```
franpog859, line 94: Maybe we could compile the regexp in the `NewAuditLogConfigurator` function just to do it once and not every time we want to use it

components/provisioner/internal/gardener/auditlogging.go
```diff
@@ -0,0 +1,107 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"regexp"
+
+	gardener_types "github.com/gardener/gardener/pkg/apis/core/v1beta1"
+)
+
+const auditLogConditionType = "AuditlogServiceAvailability"
```
franpog859, line 13: Did you look at the Gardener packages for this constant? If there is no way to get it through a package then it's ok, but if we can make it not hardcoded it'd be much better

components/provisioner/internal/gardener/auditlogging_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+		//when
+		enable := auditLogConfigurator.CanEnableAuditLogsForShoot(seedName)
+
+		//then
+		assert.False(t, enable)
+	})
+}
+
+func TestAuditLogConfigurator_SetAuditLogAnnotation(t *testing.T) {
```
franpog859, line 51: There could be also a test when the shoot is already annotated so the function will return `false, nil`, right?

components/provisioner/internal/gardener/shoot_reconciler.go
```diff
@@ -98,17 +95,9 @@ func (r *Reconciler) shouldReconcileShoot(shoot gardener_types.Shoot) (bool, err
...
-		var refetchedShoot gardener_types.Shoot
-		err := r.client.Get(context.Background(), namespacedName, &refetchedShoot)
-		if err != nil {
-			return err
-		}
-
-		modifyShootFn(&refetchedShoot)
-
-		err = r.client.Update(context.Background(), &refetchedShoot)
+		err := r.client.Update(context.Background(), modifiedShoot)
```
franpog859, line 100:
```suggestion
		return r.client.Update(context.Background(), modifiedShoot)
```
The error checking below is redundant here

components/provisioner/internal/gardener/shoot_reconciler.go
```diff
@@ -72,11 +68,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-	seed := getSeed(shoot)
-	if seed != "" && r.auditLogTenantConfigPath != "" {
-		err := r.enableAuditLogs(log, req.NamespacedName, &shoot, seed)
+	seedName := getSeedName(shoot)
+
+	if r.auditLogConfigurator.CanEnableAuditLogsForShoot(seedName) {
+		err := r.enableAuditLogs(log, &shoot, seedName)
 		if err != nil {
-			log.Errorf("Failed to enable audit logs for %s shoot: %s", shoot.Name, err.Error())
+			log.Errorf("failed to enable audit logs for %s shoot: %s", shoot.Name, err.Error())
```
franpog859, line 76: Why though :D

components/provisioner/internal/gardener/auditlogging.go
```diff
@@ -0,0 +1,107 @@
...
+		auditLogTenantConfigPath: auditLogTenantConfigPath,
+	}
+}
+
+func (a *auditLogConfigurator) CanEnableAuditLogsForShoot(seedName string) bool {
+	return seedName != "" && a.auditLogTenantConfigPath != ""
+}
+
+func (a *auditLogConfigurator) SetAuditLogAnnotation(shoot *gardener_types.Shoot, seed gardener_types.Seed) (bool, error) {
+	data, err := a.getConfigFromFile()
```
franpog859, line 36: What do you think about naming the `data` -> `auditlogConfig` or something more descriptive?

components/provisioner/internal/gardener/auditlogging.go
Maladie: I did, but couldn't find it

components/provisioner/internal/gardener/shoot_reconciler.go
Maladie: No idea :D

components/provisioner/internal/gardener/auditlogging.go
franpog859: Oh, ok 😢

franpog859 (approved):
LGTM

Issue #622, opened 2021-03-29
Refactor initialisation step and don't rerun steps that were executed without failure


PK85:
**Description**

This ticket is a part of [Make KEB well-organized](https://github.com/kyma-project/control-plane/issues/621).

**Reasons**

See [reasons](https://github.com/kyma-project/control-plane/issues/621).

**AC**

- refactor initialisation step
- don't rerun steps that were executed without failure
- before merging that enhancement, [THIS](https://github.com/kyma-project/control-plane/issues/619) one must be already in

piotrmiskiewicz:
InputCreator must be persisted as other parts of the operation. Sensitive data must be encrypted (overrides).

piotrmiskiewicz:
Tasks:
1. Persist the state of an InputCreator (and load the data from the database)
2. Do not rerun steps which succeeded
3. reorganize and split initialization step

piotrmiskiewicz:
All steps are grouped into a stage. When all steps in a stage are succeeded - the stage is finished and will never be retried.  This feature allows us to define steps like:
 Stage "preparation": Initialisation (, AVS, LMS, CLS, Create_Runtime
 Stage "check provisioner": CheckProvisionerStatus (new step which only checks the provisioning status)
 Stage "posta actions": TagsForRuntime, AvS_External

piotrmiskiewicz:
Status:
- [x]  Manager with stages
- [x]  Manager does not rerun finished stage
- [ ]  Manager does not rerun the queue
- [ ] Simplify (split) Initialisation
- [ ] Redefine a list of steps
- [ ] In case of a retry manager finds all other steps with the same weight and run it

PR #639, opened 2021-04-06
Keb with staged steps

piotrmiskiewicz:
**Description**

Implementation of a replacement for provisioning.Manager - StagedManager. This new implementation will replace existing one at the final PR
What was done in the PR:
- Manager with stages
- Operation with persisted finished stages to not unnecessary stage execution 
- Manager does not rerun finished stages 
- Manager does not rerun finished steps (it reruns only if the Pod is restarted)

**Related issue(s)**
https://github.com/kyma-project/control-plane/issues/622

### DIFF
--- a/components/kyma-environment-broker/internal/process/provisioning/staged_manager.go
+++ b/components/kyma-environment-broker/internal/process/provisioning/staged_manager.go
@@ -2,6 +2,7 @@ package provisioning
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal"
 	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/event"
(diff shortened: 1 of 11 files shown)


components/kyma-environment-broker/internal/process/provisioning/staged_manager.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return 0, nil
+}
+
+func (m *StagedManager) saveFinishedStage(operation internal.ProvisioningOperation, s *stage, log logrus.FieldLogger) (internal.ProvisioningOperation, error) {
+	operation.FinishStage(s.name)
+	op, err := m.operationStorage.UpdateProvisioningOperation(operation)
+	if err != nil {
+		log.Infof("Unable to save operation with finished stage %s: %s", s.name, err.Error())
+		return *op, err
```
jasiu001, line 139: `op` can be nil

components/kyma-environment-broker/internal/process/provisioning/staged_manager.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	log.Infof("Finished stage %s", s.name)
+	return *op, nil
```
jasiu001, line 142: maybe it is better to try couple of times as [here](https://github.com/kyma-project/control-plane/blob/main/components/provisioner/internal/operations/executor.go#L161) or use `wait` function instead of repeat all stage steps

piotrmiskiewicz: If you think about retrying failed update operation - the storage is doing that

jasiu001 (approved):

PR #644, opened 2021-04-07
Using M5 machine types for AWS

tobiscr:
**Description**

Changing from M4 to M5 machines for cost reduction and gaining higher performance.

**Related issue(s)**
https://github.tools.sap/kyma/backlog/issues/1201

### DIFF
--- a/components/kyma-environment-broker/internal/provider/aws_provider.go
+++ b/components/kyma-environment-broker/internal/provider/aws_provider.go
@@ -37,7 +37,7 @@ func (p *AWSInput) Defaults() *gqlschema.ClusterConfigInput {
 		GardenerConfig: &gqlschema.GardenerConfigInput{
 			DiskType:       ptr.String("gp2"),
 			VolumeSizeGb:   ptr.Integer(50),
-			MachineType:    "m4.2xlarge",
+			MachineType:    "m5.2xlarge",
 			Region:         DefaultAWSRegion,
 			Provider:       "aws",
 			WorkerCidr:     "10.250.0.0/19",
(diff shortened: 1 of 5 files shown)


piotrmiskiewicz (approved):

alexandra-simeonova (approved):

components/kyma-environment-broker/internal/provider/aws_provider.go
```diff
@@ -97,7 +97,7 @@ func (p *AWSTrialInput) Defaults() *gqlschema.ClusterConfigInput {
 		GardenerConfig: &gqlschema.GardenerConfigInput{
 			DiskType:       ptr.String("gp2"),
 			VolumeSizeGb:   ptr.Integer(50),
-			MachineType:    "m4.xlarge",
+			MachineType:    "m5.xlarge",
```
tobiscr, line 100: @piotrmiskiewicz : QQ - I haven't found this machine type in the schema (wasn't there before, see https://github.com/kyma-project/control-plane/pull/644/files#diff-b878f19cc19576376ee78a3ae7f507952afaf2e339968a132ea4914301dfc470R169). Could that become an issue?

piotrmiskiewicz: yes, the line you mentioned, contains the machine types in the schema

tobiscr: Yes, but currently the machine type we use for trials is missing there (`m5.xlarge"`). Should I add this type to the schema too?

Issue #619, opened 2021-03-29
Create missing test for provisioning/deprovisioning process 


PK85:
**Description**

We have a lack of test for provisioning/deprovisioning scenarios. We already have such a test for upgrades.

**Reasons**

Missing test + that will allow us to start work on KEB refactoring

**AC**
- add missing test scenario
- define other missing test scenarios (for example suspension/unsuspension)

PR #645, opened 2021-04-08
Add e2e provisioning queue test

ksputo:
**Description**

- Added e2e test covering `ProvisioningOperation` processing in the provisioning queue

**Related issue(s)**
#619

### DIFF
--- a/components/kyma-environment-broker/cmd/broker/suite_test.go
+++ b/components/kyma-environment-broker/cmd/broker/suite_test.go
@@ -7,16 +7,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/auditlog"
-	"github.com/spf13/afero"
-
-	"github.com/kyma-project/control-plane/components/kyma-environment-broker/common/hyperscaler"
-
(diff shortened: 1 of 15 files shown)


components/kyma-environment-broker/cmd/broker/suite_test.go
```diff
@@ -343,3 +363,312 @@ func fixK8sResources(defaultKymaVersion string, additionalKymaVersions []string)
...
+		provisionerClient: provisionerClient,
+		provisioningQueue: provisioningQueue,
+		storage:           db,
+		directorClient:    directorClient,
+
+		t: t,
+	}
+}
+
+func (s *ProvisioningSuite) CreateProvisioning(options RuntimeOptions) string {
```
piotrmiskiewicz, line 460: what is the purpose of this method, which just calls createProvisioning

piotrmiskiewicz:
One general comment: please reduce timeouts. The test should be green after at most few seconds (1 or 2 seconds) so havine such code: ```wait.Poll(time.Second*1, 2*time.Minute, func() (bool, error) {``` is not a good idea.

piotrmiskiewicz (approved):

PR #666, opened 2021-04-15
Block unsuspension if the instance is being deprovisioned

piotrmiskiewicz:
**Description**

- do not start unsuspension if the instance is being deprovisioned (or the deprovisioning is failed)
- unsuspension starts Pending operation
- the provisioning initialisation process checks if there is ongoing deprovisioning - if not - then changes from pending to inprogress status

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/suspension/handler.go
+++ b/components/kyma-environment-broker/internal/suspension/handler.go
@@ -60,6 +60,12 @@ func (h *ContextUpdateHandler) handleContextChange(newCtx internal.ERSContext, i
 		isActivated = *instance.Parameters.ErsContext.Active
 	}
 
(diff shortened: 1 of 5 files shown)


components/kyma-environment-broker/internal/suspension/handler.go
```diff
@@ -84,6 +84,10 @@ func (h *ContextUpdateHandler) handleContextChange(newCtx internal.ERSContext, i
 	}
 
 	if *newCtx.Active {
+		if lastDeprovisioning != nil && !lastDeprovisioning.Temporary {
```
ebensom, line 87: I think skipping the unsuspension event is not correct. It is a valid scenario to receive unsuspension while suspension is still in progress. KEB should queue the unsuspension operation but keep it pending until the suspension is completed.

jasiu001 (approved):

PR #677, opened 2021-04-19
Adapt to Kyma clusters created in AWS

sayanh:
**Description**

- Adapted to Kyma clusters created in AWS
- Only M4 and M5 AWS machines are added as per requirements


k15r (approved):

components/kyma-metrics-collector/pkg/process/event_stream.go
```diff
@@ -116,6 +118,16 @@ func (inp Input) Parse(providers *Providers) (*edp.ConsumptionMetrics, error) {
...
+		case AWS:
+			decoder := serializer.NewCodecFactory(scheme.Scheme).UniversalDecoder()
+			infraConfig := &gardenerawsv1alpha1.InfrastructureConfig{}
+			err := runtime.DecodeInto(decoder, rawExtension.Raw, infraConfig)
+			if err != nil {
+				return nil, err
+			}
+			if infraConfig.Networks.VPC.CIDR != nil {
+				vnets += 1
+			}
```
marcobebway, line 130: Why in case of **vpc** we increment **vnets**?

sayanh: As the [schema](https://github.com/kyma-project/control-plane/tree/main/components/kyma-metrics-collector#data-collection) expects `vnets` which inturn is used by other systems. Hence we had to use `vnets` for AWS as well

marcobebway (approved):

PR #686, opened 2021-04-21
Add aws_ha plan to KEB

ksputo:
**Description**

- ...
- ...
- ...

**Related issue(s)**


components/kyma-environment-broker/Makefile
```diff
@@ -36,7 +36,7 @@ go-mod-check: go-mod-check-local
 go-mod-check-local:
 	@echo make go-mod-check
 	go mod tidy
-	@if [ -n "$$(git status -s go.*)" ]; then \
+	@if [ -z "$$(git status -s go.*)" ]; then \
```
wozniakjan, line 39: this would introduce the bug with pipelines back - https://github.com/kyma-project/control-plane/pull/774

components/kyma-environment-broker/go.mod
```diff
@@ -46,6 +46,7 @@ require (
 )
 
 replace (
+	github.com/kyma-project/control-plane/components/provisioner => github.com/ksputo/control-plane/components/provisioner v0.0.0-20210611065312-8a619ab5a5e2
```
wozniakjan, line 49: I think we should merge the provisioner PR before we merge this and then we can get rid of the replace

components/kyma-environment-broker/internal/provider/aws_provider.go
```diff
@@ -79,20 +88,114 @@ func ZoneForAWSRegion(region string) string {
 	if !found {
 		zones = "a"
 	}
+	rand.Seed(time.Now().UnixNano())
```
wozniakjan, line 91: nit: I think `rand.Seed` is usually called only once, mostly in `init()`. Although I'm not sure if it's for anything else but performance reasons and if it has any effect on entropy at all if called once vs. each time before getting a random number.

components/kyma-environment-broker/internal/provider/aws_provider.go
```diff
@@ -79,20 +88,114 @@ func ZoneForAWSRegion(region string) string {
...
 	return fmt.Sprintf("%s%s", region, zone)
 }
 
+func MultipleZonesForAWSRegion(region string, zonesCount int) []string {
+	zones, found := awsZones[region]
+	if !found {
+		zones = "a"
+		zonesCount = 1
+	}
+	rand.Seed(time.Now().UnixNano())
```
wozniakjan, line 102: nit: similarly here, if we `rand.Seed` in init, we shouldn't need to call it here anymore

components/kyma-environment-broker/internal/provider/aws_provider.go
```diff
@@ -79,20 +88,114 @@ func ZoneForAWSRegion(region string) string {
...
+
+	availableZones := strings.Split(zones, "")
+	rand.Shuffle(len(availableZones), func(i, j int) { availableZones[i], availableZones[j] = availableZones[j], availableZones[i] })
+	availableZones = availableZones[:zonesCount]
```
wozniakjan, line 106: do we want to ensure that `zonesCount` is less than `len(availableZones)`? If I understand this correctly, `zonesCount` may at some point in future be user input and this may panic if `zonesCount` end up being larger then capacity of `availableZones`.

wozniakjan (approved):
looks pretty good, few nits below

PR #699, opened 2021-04-23
Provisioning AWS HA runtimes

ksputo:
**Description**
Pull request introduce `aws_ha` plan, which spins AWS cluster with multiple availability zones.

- Added `aws_ha` plan with 2 availability zones by default `planID: aecef2e6-49f1-4094-8433-eba0e135eb6a`
- User can specify number of zones using `ZonesCount` parameter, limits are in the schema
- Changed KEB-Provisioner contract for AWS, from providing single zone to providing multiple zones (zone <String> -> awsZones []<AwsZone>)
- Implemented migration in provisioner `components/provisioner/internal/provider-config-migrator/migrator.go`

**Related issue(s)**
https://github.tools.sap/kyma/backlog/issues/1308


components/provisioner/pkg/gqlschema/schema.graphql
```diff
@@ -44,10 +44,8 @@ type AzureProviderConfig {
 }
 
 type AWSProviderConfig {
-    zone: String
+    zones: [AWSZone]!
```
ksputo, line 47: possible renaming needed here:
```
"errors": [
      {
        "message": "Fields \"zones\" conflict because they return conflicting types [String!]! and [AWSZone]!. Use different aliases on the fields to fetch both if this was intentional.",
        ```

File: components/kyma-environment-broker/internal/broker/testdata/aws-ha-schema.json [hunk 2]
```diff
@@ -0,0 +1,44 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
piotrmiskiewicz, line 2: the new parameter - zonesCount, is it specified in the schema?

components/kyma-environment-broker/internal/avs/model.go
```diff
@@ -130,7 +130,7 @@ func generateNameAndDescription(operation internal.ProvisioningOperation, beType
 
 func providerCodeByPlan(planID string) string {
 	switch planID {
-	case broker.AWSPlanID:
+	case broker.AWSPlanID, broker.AWSHAPlanID:
```
piotrmiskiewicz, line 133: please, create a ticket to use CloudProvider field instead of duplicating a logic

resources/kcp/values.yaml
```diff
@@ -9,10 +9,10 @@ global:
       version: "PR-769"
     provisioner:
       dir:
-      version: "PR-769"
+      version: "PR-699"
     kyma_environment_broker:
       dir:
-      version: "PR-556"
+      version: "PR-686"
```
piotrmiskiewicz, line 15: are yuo sure? why not 699?

components/kyma-environment-broker/Makefile
```diff
@@ -36,7 +36,7 @@ go-mod-check: go-mod-check-local
 go-mod-check-local:
 	@echo make go-mod-check
 	go mod tidy
-	@if [ -n "$$(git status -s go.*)" ]; then \
+	@if [ -z "$$(git status -s go.*)" ]; then \
```
wozniakjan, line 39: this would introduce the bug with pipelines back - #774

components/kyma-environment-broker/go.mod
```diff
@@ -46,6 +46,7 @@ require (
 )
 
 replace (
+	github.com/kyma-project/control-plane/components/provisioner => github.com/ksputo/control-plane/components/provisioner v0.0.0-20210616080800-d0d1cc0ca19e
```
wozniakjan, line 49: if I understand this correctly, the provisioner changes are part of this PR as well? Wondering if we should do this instead before merging
```
github.com/kyma-project/control-plane/components/provisioner => ../provisioner
```

I think it's similar use-case kubernetes has with API dependency for example
https://github.com/kubernetes/kubernetes/blob/513ae557a3231cde76ad155bbbd4e18f8182df25/go.mod#L490

components/kyma-environment-broker/internal/provider/aws_provider.go
```diff
@@ -79,20 +88,114 @@ func ZoneForAWSRegion(region string) string {
 	if !found {
 		zones = "a"
 	}
+	rand.Seed(time.Now().UnixNano())
```
wozniakjan, line 91: nit: I think `rand.Seed` is usually called only once, mostly in `init()`. Although I'm not sure if it's for anything else but performance reasons and if it has any effect on entropy at all if called once vs. each time before getting a random number.

components/kyma-environment-broker/internal/provider/aws_provider.go
```diff
@@ -79,20 +88,114 @@ func ZoneForAWSRegion(region string) string {
...
 	return fmt.Sprintf("%s%s", region, zone)
 }
 
+func MultipleZonesForAWSRegion(region string, zonesCount int) []string {
+	zones, found := awsZones[region]
+	if !found {
+		zones = "a"
+		zonesCount = 1
+	}
+	rand.Seed(time.Now().UnixNano())
```
wozniakjan, line 102: nit: similarly here, if we `rand.Seed` in init, we shouldn't need to call it here anymore

components/kyma-environment-broker/internal/provider/aws_provider.go
```diff
@@ -79,20 +88,114 @@ func ZoneForAWSRegion(region string) string {
...
+
+	availableZones := strings.Split(zones, "")
+	rand.Shuffle(len(availableZones), func(i, j int) { availableZones[i], availableZones[j] = availableZones[j], availableZones[i] })
+	availableZones = availableZones[:zonesCount]
```
wozniakjan, line 106: nit: do we want to ensure that `zonesCount` is less than `len(availableZones)`? If I understand this correctly, `zonesCount` may at some point in future be user input and this may panic if `zonesCount` end up being larger than the capacity of `availableZones`.

wozniakjan (dismissed):
looking good, porting my review from https://github.com/kyma-project/control-plane/pull/686#pullrequestreview-686299009, just couple of nits

rafalpotempa (dismissed):
I tested the Provisioner flow and it works. 
✅   from Framefrog.

rafalpotempa (approved):

piotrmiskiewicz (approved):

piotrmiskiewicz (approved):

rafalpotempa (approved):

PR #713, opened 2021-05-06
Set cloud provider type in AVS evaluation names correctly

ebensom:
**Description**

The AVS evaluation names have to reflect the correct cloud provider type correctly:

- for Azure it should be AZR
- for AWS it's AWS

Added shoot name to the AVS evaluation description.

### DIFF
--- a/components/kyma-environment-broker/internal/avs/model.go
+++ b/components/kyma-environment-broker/internal/avs/model.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/broker"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


components/kyma-environment-broker/internal/avs/model.go
```diff
@@ -116,14 +116,35 @@ func newBasicEvaluationCreateRequest(operation internal.ProvisioningOperation, e
...
 
+func providerCodeByPlan(planID string) string {
+	switch planID {
+	case broker.AWSPlanID:
+		return "AWS"
+	case broker.GCPPlanID:
+		return "GCP"
+	case broker.AzurePlanID, broker.AzureLitePlanID, broker.AzureHAPlanID:
+		return "AZR"
+	case broker.TrialPlanID, broker.FreemiumPlanID:
```
piotrmiskiewicz, line 139: freemium could be AWS or Azure

piotrmiskiewicz:
We are planning to have an explicit value in an instance or operation which defines the provider (calculated provider) and add it to an info in /runtimes endpoint.

piotrmiskiewicz (approved):

components/kyma-environment-broker/internal/avs/model.go
PK85: exactly, I think we should do first https://github.tools.sap/kyma/backlog/issues/1473 and then adjust and merge this PR.

PR #719, opened 2021-05-10
Retrigger Kyma upgrade if installation is in error state

ebensom:
**Description**

- provisioner triggers a new upgrade when the installation is in Error state. 
- In theory the installation could still be a non-permanent error state between installer retries, and the retrigger would wipe out the kyma-installer deployment and recreate the config and installation CR, but in practice this wouldn't be a problem.

**Related issue(s)**


franpog859 (commented):
Hey! We'll need to check that but I think I have an objection. The installer sometimes fails to install some components (ory, istio, and so on). It retries it and most of the time succeeds. The fail leads to InstallationError though so this implementation would cause the retriggering of the upgrade every small, recoverable installation error 

**EDIT:** Yup, it looks like Installer returns the InstallationError when it's still going to be recovered:
>`time="2021-05-11T08:23:14Z" level=warning msg="Installation error occurred: installation error occurred: install component application-connector" Component=Executor OperationId=123 OperationType=PROVISION RuntimeId=123 ShootName=qwe Stage=WaitingForInstallation`

I think we don't want to retrigger the upgrade immediately then. It regards mostly [this part of the implementation](https://github.com/kyma-project/control-plane/pull/719/files#diff-6a6a86d0012b950ef15d0a3b9aeedff5a74c8d601db33730aa70c01a67293dc0R53-R56)

ebensom:
> I think we don't want to retrigger the upgrade immediately then. It regards mostly [this part of the implementation](https://github.com/kyma-project/control-plane/pull/719/files#diff-6a6a86d0012b950ef15d0a3b9aeedff5a74c8d601db33730aa70c01a67293dc0R53-R56)

Yes, I was also considering this issue. What we could do is to introduce a new method `RetriggerUpgrade` in the Installation client and corresponding hydroform installation package which just adds the `action: install` label in case it's missing. This won't hurt ongoing installations while still allow to retrigger errored installations from the Kyma control plane, which is our goal.

The problem I would like to solve with this PR is to avoid having manually go to several Kyma runtimes one-by-one which has failed upgrades, and retrigger them. The retriggers should be possible from the central KCP.

franpog859 (commented):
Idea with a new `RetriggerUpgrade` method sounds much more safe than automatic retriggering. It’d need to be used safely anyway as it just ignores the Instalator errors. It’s quite big though. New Provisioner operation, changes to the database, GraphQL schema, Hydroform and so on. I’m afraid Frogs don’t have such capacity right now :disappointed: 

I see that you mainly want to fix the issue with the not existing `action=install` label. Some time ago I merged [a fix to the Provision operation](https://github.com/kyma-project/control-plane/pull/715) which may solve the issue (it’s not on PROD yet). Do you think it may solve your problem? It should work for Provisioning clusters but sadly may not for the Upgrades (Installer is already present on the cluster and probably won’t have the Empty state if failed to deploy some secrets)

PR #769, opened 2021-05-28
Self-service for configuring custom OIDC

rJankowski93:
**Description**

- ...
- ...
- ...

**Related issue(s)**
https://github.tools.sap/kyma/backlog/issues/1361


components/provisioner/assets/database/provisioner.sql
```diff
@@ -12,6 +12,26 @@ CREATE TABLE cluster
...
+    username_claim text NOT NULL,
+    username_prefix text NOT NULL
+);
+
+CREATE TABLE signing_algorithms
+(
+    id uuid PRIMARY KEY CHECK (id <> '00000000-0000-0000-0000-000000000000'),
+    cluster_id uuid NOT NULL,
+    algorithm text NOT NULL
+);
```
franpog859, line 33: There is no need to have `cluster_id` here as we keep the OIDC config ID in `gardener_config`. In the `signing_algorithms` should be the ID of the oidc_config as a foreign key as the algorithms are part of it. When the OIDC config is deleted so should be the signing_algorithms

I think it may cause little changes also in the implementation part

franpog859 (commented):
The code looks ok! Remember to rename the `202131051400_oidc_config.X.sql` migration files to more recent dates just to be sure there are no races

When you manage to pass the `pre-control-plane-components-provisioner` job, please, run the `/test pre-main-control-plane-gke-provisioner-integration` and connect to the Prow cluster. It fails most of the times but you can see if all the migrations and the schemas work fine

I'm proceeding to the manual tests now

components/provisioner/assets/database/provisioner.sql
rJankowski93: Done. 
I added to `oidc_config` FK  gardener_config_id
and to `signing_algorithms` FK oidc_config_id

franpog859 (approved):

franpog859 (approved):

PR #785, opened 2021-06-08
Enable OIDC authcode + PKCE flow in kubeconfig-service

ebensom:
**Description**

With OIDC authorization code grant + PKCE flow the OIDC client secret is not needed in the generated kubeconfig of the SKR.


strekm (approved):
/meow

PR #799, opened 2021-06-15
Self-service for overriding admin user

rJankowski93:
**Description**

- ...
- ...
- ...

**Related issue(s)**


components/provisioner/internal/provisioning/service_test.go
```diff
@@ -742,6 +742,7 @@ func TestService_UpgradeGardenerShoot(t *testing.T) {
 		sessionFactory.On("NewSessionWithinTransaction").Return(writeSession, nil)
 		writeSession.On("UpdateGardenerClusterConfig", upgradedConfig).Return(nil)
 		writeSession.On("RollbackUnlessCommitted").Return()
+		writeSession.On("InsertAdministrators", runtimeID, mock.Anything).Return(nil)
```
franpog859, line 745: I think we can be more specific here. At least `mock.AnythingOfType(...`

components/provisioner/pkg/gqlschema/schema.graphql
```diff
@@ -265,6 +265,7 @@ input UpgradeRuntimeInput {
 
 input UpgradeShootInput {
     gardenerConfig: GardenerUpgradeInput! # Gardener-specific configuration for the cluster to be upgraded
+    administrators: [String]                # List of administrators
```
franpog859, line 268: I'd suggest changing it to `administrators: [String!]`. We don't need to have `null` administrators

franpog859 (commented):
Please, bump the Provisioner in this PR

Add also a description to this PR as it could help in the future while troubleshooting or other cases. Describe that it adds the option to replace administrators while upgrading a shoot

franpog859 (approved):
LGTM

Please, update the PR description though

PR #801, opened 2021-06-17
Don't allow unavailable nodes during trial cluster upgrades

ebensom:
**Description**

The autoscaler maxUnavailable is set to 1. With this setting, Gardener starts to drain the only single worker node while the new node is being created, causing several minutes downtime for all workloads.


ksputo:
What about other plans? I see that `openstack_provider.go` also has this parameter set to `1` - could you change it too?

ksputo (approved):
lgtm

PR #821, opened 2021-06-29
Randomize maintenance window for non-porduction clusters

AdrianPei:
**Description**

- ...Randomize maintenance window for non-porduction clusters
- ...Add purpose attribute to provision unit test
- ...Update image version in kcp chart to PR-821

**Related issue(s)**

Resolves #1583

### DIFF
--- a/components/provisioner/internal/gardener/provisioner.go
+++ b/components/provisioner/internal/gardener/provisioner.go
@@ -64,8 +64,12 @@ func (g *GardenerProvisioner) ProvisionCluster(cluster model.Cluster, operationI
 	}
 
 	region := cluster.ClusterConfig.Region
+	purpose := ""
+	if cluster.ClusterConfig.Purpose != nil {
+		purpose = *cluster.ClusterConfig.Purpose
+	}
 
-	if g.shouldSetMaintenanceWindow() {
+	if g.shouldSetMaintenanceWindow(purpose) {
 		err := g.setMaintenanceWindow(shootTemplate, region)
 
 		if err != nil {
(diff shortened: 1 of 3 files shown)


components/provisioner/internal/gardener/provisioner.go
```diff
@@ -64,8 +64,9 @@ func (g *GardenerProvisioner) ProvisionCluster(cluster model.Cluster, operationI
 	}
 
 	region := cluster.ClusterConfig.Region
+	purpose := *cluster.ClusterConfig.Purpose
```
ebensom, line 67: Can be nil, in which case this would crash. 

```suggestion
	purpose := ""
	if cluster.ClusterConfig.Purpose != nil {
		purpose = *cluster.ClusterConfig.Purpose
	}
```

components/provisioner/internal/gardener/provisioner_test.go
```diff
@@ -35,6 +35,7 @@ const (
 	operationId       = "operationId"
 	clusterName       = "test-cluster"
 	region            = "westeurope"
+	purpose           = "proudction"
```
ebensom, line 38: minor typo
```suggestion
	purpose           = "production"
```

ebensom (approved):
Please also bump the image version in kcp chart to PR-821 here:
https://github.com/kyma-project/control-plane/blob/main/resources/kcp/values.yaml#L12

ebensom (dismissed):

franpog859 (commented):
I think we could also add a unit test that actually tests this functionality. So there will be cluster with the `production` purpose, with other one, and with `nil` one, and there will be different maintenance windows present as a result

There is also an option to change the purpose while calling the `UpgradeShoot` mutation. Do you plan to somehow change the maintenance window in the case that the purpose is changed from `production` to something else?

AdrianPei:
It's a good idea to have an unit test for production purpose clusters. I can add one .
As for purpose change,  there's no plan for maintenance window change.  Users will keep the configuration and change it manually if they need.

ebensom (approved):

franpog859 (approved):

PR #825, opened 2021-06-30
KEB service catalog migration

wozniakjan:
**Description**

This PR contains changes required to install `sc-migration` and `btp-operator` when ERS triggers the instance PATCH endpoint with btp-operator credentials and migration flag.

**Notable changes**
* the update queue designed to allow conditional step triggers, similar to provisioning and upgrade queues
* instance update endpoint trigger for sc-migration
  * install btp-operator with sc-migration components
  * wait for sc-migration to finish
    * determined by reconciler `status != ClusterStatusReady`
    * the `sc-migration` component reconciler will need to report to root reconciler state pending until migration is finished
    * this should ensure KEB won't remove svcat components prematurely
* reject with 422 request for migration on kyma 1.x instances
 

**Related PRs and issues:**
- [x] https://github.com/kyma-project/control-plane/pull/980
- [x] https://github.com/kyma-project/control-plane/pull/987
- [x] https://github.com/kyma-project/control-plane/pull/1009
- [ ] https://github.com/kyma-incubator/reconciler/issues/527
- [ ] https://github.com/kyma-incubator/reconciler/pull/389

**Summary of reviews:**
- [x] Introducing new `RequiresReconcilerUpdate` on the `UpdatingOperation` to only execute reconciler update when necessary https://github.com/kyma-project/control-plane/pull/825/commits/d0f0a677e114162cfa287521d6e4d61257b5d7e8.
- [x] introduced a new step for `runtime` stage responsible for filling out kyma version and kubeconfig - https://github.com/kyma-project/control-plane/pull/825/commits/8119b47f64171d83770ed391cd252b2772380ff5
- [x] hide behind a feature flag - https://github.com/kyma-project/control-plane/pull/825/commits/d955b78ae727992b2038d296caaef9fbb2eb4a8a
- [x] move the flag in the update operation to not store in db with `json:-` - https://github.com/kyma-project/control-plane/pull/825/commits/7318457fb32a8a393f6618765b1efc9d9c73567a
- [x] exclude duplicate reconciler calls and check when no migration flag only btp operator - https://github.com/kyma-project/control-plane/pull/825/commits/c8e296abdf273e73a9e883a086fb0e9b259947a3
- [x] rename feature flag - https://github.com/kyma-project/control-plane/pull/825/commits/8e1cc41c10fab5780c0648136af8913acb07b5c9
- [x] reimplement instance details migration flag - https://github.com/kyma-project/control-plane/pull/825/commits/1d6473243501a223af5ae71ef62910696c24e1f1
- [x] change name of check `update.NewCheckSCMigrationDone` - https://github.com/kyma-project/control-plane/pull/825/commits/9142ab5eb76a9e9a87e7ac68fcd178a355232980
- [x] check the status from reconciler again after svcat and migrator are no longer in list of components - https://github.com/kyma-project/control-plane/pull/825/commits/9142ab5eb76a9e9a87e7ac68fcd178a355232980
- [x] disable sc-migration component in upgrade queue - https://github.com/kyma-project/control-plane/pull/825/commits/5cc34c0dd7be23fe7bf90d2d6af1e95b696dbd65


components/kyma-environment-broker/internal/process/provisioning/btp_operator_overrides.go
```diff
@@ -0,0 +1,61 @@
...
+		},
+		{
+			Key:   "manager.secret.tokenurl",
+			Value: creds.TokenURL,
+		},
+		{
+			Key:   "cluster.id",
+			Value: creds.ClusterID,
+		},
+	}
```
wozniakjan, line 58: the mapping is very well described in
https://github.com/SAP/sap-btp-service-operator/blob/v0.1.9/README.md#setup

PK85:
General comment about payload:

```
"sm_platform_credentials":{
```
Above attribute doesn't exist in current ERS implementation, and all cluster has:
```
   "context": {
        "globalaccount_id": "28e92f57-5fb5-4de2-91af-4c24ca5feb20",
			  "subaccount_id": "abc",
			  "username": "abc",
			  "password": "abc",
			  "url": "url",
    }
```

wozniakjan:
> Above "sm_platform_credentials" attribute doesn't exist on ERS side, and all our clusters received:

ah, thanks for the clarification, don't know what I was looking at

> Btw what "cluster_id":"ab32866b-8d97-409c-a3c4-145340d7be8e" is for?

it is necessary for the btp-operator installation when considering migration, https://github.com/SAP/sap-btp-service-operator-migration/blob/5189ed2/README.md#setup

wozniakjan:
@PK85 I may need some pointers for the `credentials`. From my testing and reading the code, I can't figure out how would this be the right schema
```json
   "context": {
	"username": "abc",
	"password": "abc",
	"url": "url",
    }
```
as far as I can tell, the `PUT` `instance` payload is unmarshalled in two runs.
1) generic unmarshal `ProvisioningDetails` in an API handler, where `context` remains [`raw` slice of `[]byte`](
https://github.com/pivotal-cf/brokerapi/blob/6d02437ac0f3436e6d8a70fded4c8c1655650ccf/domain/service_broker.go#L88)
2) unmarshal `context` to `ERSContext` here in the control plane
https://github.com/kyma-project/control-plane/blob/54a32c868ce15b9dd202a3e3f82de6ed9a91bc7d/components/kyma-environment-broker/internal/broker/instance_create.go#L244-L246

wozniakjan:
for the record, @PK85 is ok with structuring the btp-operator credentials under `context.sm_platform_credentials.btp_operator` https://sap-btp.slack.com/archives/G01L562297B/p1625658497004000?thread_ts=1625656575.003400&cid=G01L562297B

Tested and I think it works ok in conjunction with https://github.tools.sap/kyma/management-plane-config/pull/1035.

components/kyma-environment-broker/internal/dto.go
```diff
@@ -108,8 +108,9 @@ type ERSContext struct {
 }
```
piotrmiskiewicz, line 109: add isMigrationFromSCtoOperator flag

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -738,11 +743,41 @@ func NewProvisioningProcessingQueue(ctx context.Context, provisionManager *provi
...
 	provisionerClient provisioner.Client, publisher event.Publisher, logs logrus.FieldLogger) *process.Queue {
 
-	manager.DefineStages([]string{"cluster", "check"})
-	manager.AddStep("cluster", update.NewInitialisationStep(db.Instances(), db.Operations(), inputFactory))
-	manager.AddStep("cluster", update.NewUpgradeShootStep(db.Operations(), db.RuntimeStates(), provisionerClient))
-	manager.AddStep("check", update.NewCheckStep(db.Operations(), provisionerClient, 40*time.Minute))
+	updateSteps := []struct {
+		stage     string
+		step      update.Step
+		condition update.StepCondition
```
wozniakjan, line 749: @piotrmiskiewicz there were few things missing in update queue so I modeled it after the provisioning queue

components/kyma-environment-broker/internal/process/update/conditions.go
```diff
@@ -0,0 +1,24 @@
+package update
+
+import (
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal"
+)
+
+func ForKyma2(op internal.UpdatingOperation) bool {
+	return op.InstanceDetails.ClusterConfigurationVersion == 2
```
wozniakjan, line 8: @piotrmiskiewicz for update operation queue, is it reasonable to take this from the `InstanceDetails`?

piotrmiskiewicz: We need to look for last update kyma or provisioning operation and take RuntimeVersionData from that object. The main difference between upgradeKyma operation and the updating operation is that in the updating operation we need to keep the same version but in the upogradeKyma - we always have to compute the required version (could differ than the last one)

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -687,13 +694,60 @@ func NewProvisioningProcessingQueue(ctx context.Context, provisionManager *provi
...
+		},
+		{
+			stage:     "runtime",
+			step:      update.NewApplyReconcilerConfigurationStep(db.Operations(), db.RuntimeStates(), reconcilerClient),
+			condition: update.ForKyma2,
+		},
+		{
+			stage: "check",
+			step:  update.NewCheckStep(db.Operations(), provisionerClient, 40*time.Minute),
+		},
```
wozniakjan, line 742: the new steps for update perform
- fetch lastest `runtime_state`
- if `btp-operator` credentials are provided instal `btp-operator` component
- if migration flag is provided, install `sc-migration` component
- call reconciler with modified payload (list of components and overrides contain `btp-operator` and/or `sc-migration`)
- keep checking the reconciler whether the migration is done, `sc-migration` should not claim `reconciled` unless the migration has not finished
- remove SVCAT related components
- call reconciler with modified payload (removed SVCAT components)

components/kyma-environment-broker/internal/component_input_helpers.go
```diff
@@ -0,0 +1,120 @@
...
+func DisableServiceManagementComponents(r ProvisionerInputCreator) {
+	r.DisableOptionalComponent(SCMigrationComponentName)
+	r.DisableOptionalComponent(HelmBrokerComponentName)
+	r.DisableOptionalComponent(ServiceCatalogComponentName)
+	r.DisableOptionalComponent(ServiceCatalogAddonsComponentName)
+	r.DisableOptionalComponent(ServiceManagerComponentName)
+	r.DisableOptionalComponent(BTPOperatorComponentName)
+}
+
+func getBTPOperatorProvisioningOverrides(creds *ServiceManagerOperatorCredentials) []*gqlschema.ConfigEntryInput {
```
wozniakjan, line 35: this might be controversial, I would be interested in your opinion @piotrmiskiewicz and @ksputo, I very much disliked to write the same code for provisioning, updating and upgrade queue, I would like to propose this to serve as a common place and steps in the queues can reuse the same implementation.

components/kyma-environment-broker/internal/process/update/initialisation_step.go
```diff
@@ -79,6 +84,31 @@ func (s *InitialisationStep) Run(operation internal.UpdatingOperation, log logru
...
+		if version != nil {
+			op.RuntimeVersion = *version
+		}
+		op.LastRuntimeState = lrs
+	}, log)
+	operation = op
+	if delay != 0 {
+		return operation, delay, nil
+	}
+
```
piotrmiskiewicz, line 111: this code block which sets last runtime state sets it as an empty (zero value) value for kyma 1. If someone will implement to use it for kyma one then we won't see any error directly. let's discuss it

piotrmiskiewicz:
You have added few steps to the update process. It is an OSB update action, which can be triggered because of other reasons, not only SC migration. For example OIDC parameters. In such case we have also 2 calls to the reconciler...

components/kyma-environment-broker/internal/storage/driver/postsql/instance.go
```diff
@@ -330,7 +330,10 @@ func (s *Instance) GetByID(instanceID string) (*internal.Instance, error) {
 		}
 		return nil, err
 	}
+	// note for migration, instance has more up-to-date version of migration state than a previous operation
+	migrationTriggered := instance.InstanceDetails.SCMigrationTriggered
 	instance.InstanceDetails = lastOp.InstanceDetails
+	instance.InstanceDetails.SCMigrationTriggered = migrationTriggered
```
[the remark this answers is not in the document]
piotrmiskiewicz: Instance Details is a struct which describes some created/computed/gathered data for given instance. Because of some historical reasons (now I would implement it a different way) those data are stored in the operations table, but logically it belongs to the instance. That's why if you call a storage to get the instance - you will get also instance details. I think you don't need to do such hack.

components/kyma-environment-broker/internal/process/update/initialisation_step.go
wozniakjan: Thanks! good to know, I will introduce a new step for the "runtime" stage, I think that is the only stage requiring this information.

wozniakjan:
> For example OIDC parameters. In such case we have also 2 calls to the reconciler...

Thanks! I will try to add some extra logic so it's triggered only for the svcat migration.

wozniakjan:
@piotrmiskiewicz Thanks again for your insight, I attempted to resolve both of your recent comments

> You have added few steps to the update process. It is an OSB update action, which can be triggered because of other reasons, not only SC migration. For example OIDC parameters. In such case we have also 2 calls to the reconciler...

tried to address in https://github.com/kyma-project/control-plane/pull/825/commits/d0f0a677e114162cfa287521d6e4d61257b5d7e8. Introducing new `RequiresReconcilerUpdate` on the `UpdatingOperation`. Steps that modify reconciler payload from `runtime_state` then set this flag when reconciler call should be required. I think this will make sure that those 2 calls for reconciler happen only when necessary and not for OIDC params.

> The idea of stages is that once the stage is finished, it won't be retried. So if we need a field (lastRuntimeState) in stage "runtime", we need to set it in the same stage, not in the initialisation step

https://github.com/kyma-project/control-plane/pull/825/commits/8119b47f64171d83770ed391cd252b2772380ff5 introduced new step for `runtime` stage responsible for filling out kyma version so it should be available in that the `runtime` stage even on retries.

> this code block which sets last runtime state sets it as an empty (zero value) value for kyma 1. If someone will implement to use it for kyma one then we won't see any error directly. let's discuss it

I have tested the block and it appears to set kyma 1 as well as kyma 2 correctly. My logic is that if `runtime_state` contains reconciler payload, I take kyma version from there. If it doesn't, I take kyma version from `kyma_config.version` instead.

components/kyma-environment-broker/internal/component_input_helpers.go
piotrmiskiewicz: the plan was to have common implementation (to be reused) but there was no time for refactoring

wozniakjan:
- [ ] hide behind feature flag
- [ ] move requres update in the update operation to not store in db with `json:-`
- [ ] exclude duplicate reconciler calls and check when no migration flag only btp operator

components/kyma-environment-broker/internal/process/update/initialisation_step.go
```diff
@@ -64,6 +64,10 @@ func (s *InitialisationStep) Run(operation internal.UpdatingOperation, log logru
 		if err != nil {
 			return operation, time.Second, nil
 		}
+		// because below lines would overwrite the trigger taking the instance details from
+		// previous operation while the update is in state Pending, this ensures that we
+		// overwrite the overwritten overwrites...
+		instance.InstanceDetails.SCMigrationTriggered = operation.InstanceDetails.SCMigrationTriggered
```
wozniakjan, line 70: @ksputo @piotrmiskiewicz why do we overwrite instance details in the update operation?
because I think this prevents any changes to instance details in the PATCH instance endpoint. What I observed to happen is:
* I set the migration flag in the instance details
* the update operation gets created with correct instance details
* the update/iinitialisation_step looks up the instance details from DB
* this means taking the lastOperation, that didn't have the migration flag obviously because it was an entirely different operation
https://github.com/kyma-project/control-plane/blob/5bb0ad9c1d451d83313964ba3cb76f9aa6817407/components/kyma-environment-broker/internal/storage/driver/postsql/instance.go#L326-L333
* overwrite instance details with migration flag for the update operation to always false unless it was present for provisioning request already (which wouldn't make a lot of sense imo)

components/kyma-environment-broker/internal/storage/postsql/read.go
```diff
@@ -429,6 +429,35 @@ func (r readSession) GetLatestRuntimeStateWithReconcilerInputByRuntimeID(runtime
...
+	))
+
+	count, err := r.session.
+		Select("*").
+		From(RuntimeStateTableName).
+		Where(condition).
+		OrderDesc(CreatedAtField).
+		// because both cluster_setup as well as kyma_config are encrypted
+		// we can't easily query the content of the json. This gets the last
+		// N operations for further processing
```
wozniakjan, line 446: I was trying to use https://www.postgresql.org/docs/8.3/pgcrypto.html instead and query on decrypted json object but I couldn't get it to work the same way the encryption works in KEB go code. The reason for this is that for each operation, we store some `kyma_config`. Frequently, after decoding, it is just `{ <nil> [] [] <nil>}` but it makes the query harder to implement because the content is encrypted data.

wozniakjan: TODO: we will populate `runtime_state.kymaVersion` in the database retrospectively for all relevant entries and use that in the future.

components/kyma-environment-broker/internal/process/update/initialisation_step.go
wozniakjan: populate instance details migration triggered from `operation.provisioningparameters.erscontext` instead

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -711,13 +713,88 @@ func NewProvisioningProcessingQueue(ctx context.Context, provisionManager *provi
...
 func NewUpdateProcessingQueue(ctx context.Context, manager *update.Manager, workersAmount int, db storage.BrokerStorage, inputFactory input.CreatorForPlan,
-	provisionerClient provisioner.Client, publisher event.Publisher, logs logrus.FieldLogger) *process.Queue {
+	provisionerClient provisioner.Client, publisher event.Publisher, runtimeVerConfigurator *runtimeversion.RuntimeVersionConfigurator, runtimeStatesDb storage.RuntimeStates,
+	runtimeProvider input.ComponentListProvider, reconcilerClient reconciler.Client, cfg Config, logs logrus.FieldLogger) *process.Queue {
 
-	manager.DefineStages([]string{"cluster", "check"})
-	manager.AddStep("cluster", update.NewInitialisationStep(db.Instances(), db.Operations(), inputFactory))
-	manager.AddStep("cluster", update.NewUpgradeShootStep(db.Operations(), db.RuntimeStates(), provisionerClient))
-	manager.AddStep("check", update.NewCheckStep(db.Operations(), provisionerClient, 40*time.Minute))
+	ifFeatureEnabled := func(c update.StepCondition) update.StepCondition {
```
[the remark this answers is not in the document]
wozniakjan: good idea, renamed in https://github.com/kyma-project/control-plane/commit/8e1cc41c10fab5780c0648136af8913acb07b5c9

components/kyma-environment-broker/internal/reconciler/fake.go
```diff
@@ -128,11 +128,11 @@ func (c *FakeClient) createOrUpdate(cluster Cluster) (*State, error) {
 	// cluster exists in db - add new configuration version
 	//TODO: implement comparision mechanism for configs (new config should not be added if nothing changes in request) - needed for upgrade testing
 	//TODO: implement clusterVersion bumping (happens when Kyma version is updated?)
-	latestConfigVersion := int64(len(c.inventoryClusters[cluster.Cluster].clusterStates))
+	latestConfigVersion := int64(len(c.inventoryClusters[cluster.Cluster].clusterStates)) + 1
```
ksputo, line 131:
```suggestion
	latestConfigVersion := int64(len(c.inventoryClusters[cluster.Cluster].clusterStates)) + 1
	
	c.inventoryClusters[cluster.Cluster].clusterConfigs[latestConfigVersion] = cluster
```

adding this line would allow us to keep track of configurations passed to fake-reconciler with `ApplyClusterConfiguration` calls for easier debugging

wozniakjan: thanks, I have rebased on top of https://github.com/kyma-project/control-plane/pull/1166, I think that also includes this change

piotrmiskiewicz (approved):

piotrmiskiewicz (approved):

components/kyma-environment-broker/internal/component_input_helpers.go
```diff
@@ -0,0 +1,120 @@
...
+
+type ClusterIDGetter func() (string, error)
+
+func DisableServiceManagementComponents(r ProvisionerInputCreator) {
+	r.DisableOptionalComponent(SCMigrationComponentName)
+	r.DisableOptionalComponent(HelmBrokerComponentName)
+	r.DisableOptionalComponent(ServiceCatalogComponentName)
+	r.DisableOptionalComponent(ServiceCatalogAddonsComponentName)
+	r.DisableOptionalComponent(ServiceManagerComponentName)
+	r.DisableOptionalComponent(BTPOperatorComponentName)
```
wozniakjan, line 32: explicitly disable all kyma1 and kyma2 components, some paths then enable SVCAT related for kyma1 or btp-operator for kyma2

PR #834, opened 2021-07-08
[US-933] Support "day" input for KEB maintenance windows

berkanercan:
**Description**

- Adding the logic to read maintenance policies from a configmap and update the maintenance windows accordingly.


components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -88,8 +107,21 @@ func (m *orchestrationManager) resolveOperations(o *internal.Orchestration) ([]o
 		for _, r := range runtimes {
 			windowBegin := time.Time{}
 			windowEnd := time.Time{}
+
+			maintenanceDays := r.MaintenanceDays
+			maintenanceWindowBegin := r.MaintenanceWindowBegin
+			maintenanceWindowEnd := r.MaintenanceWindowEnd
+
+			for _, p := range policies {
+				if p.GlobalAccountID == r.GlobalAccountID && p.Plan == r.Plan { // TODO: with what should I compare p.Plan?
```
ebensom, line 116: You should separate the matching of GA and Plan. Both of these match attributes are optional, but if one is specified, and if it's not matching the runtime's attribue, we should skip the rule entry. I think for GA we should support regexp instead of exact match only. 

So something like:
```
if p.Match.Plan != "" && p.Match.Plan != r.Plan {
	continue
}
if p.Match.GlobalAccountID != "" {
	matched, err := regexp.MatchString(p.Match.GlobalAccountID, r.GlobalAccountID)
	if err != nil || !matched {
		continue
	}
}

// We have a rule match here, either be one or all of the rule match options. Let's override maintenance attributes.
if len(p.Days) > 0 {
	maintenanceDays = p.Days
}
if !p.TimeBegin.IsZero() {
	maintenanceWindowBegin = p.TimeBegin
}
if !p.TimeEnd.IsZero() {
	maintenanceWindowEnd = p.TimeEnd
}
```

components/kyma-environment-broker/common/orchestration/automock/K8sClientProvider.go
```diff
@@ -0,0 +1,38 @@
+// Code generated by mockery 2.9.0. DO NOT EDIT.
```
ebensom, line 1: This doesn't seem to be imported anywhere. Where is it used? upgrade tests seem to use github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/orchestration/mock

components/kyma-environment-broker/common/orchestration/dto.go
```diff
@@ -153,3 +153,15 @@ type StatusResponseList struct {
 type UpgradeResponse struct {
 	OrchestrationID string `json:"orchestrationID"`
 }
+
+type MaintenancePolicyEntry struct {
+	GlobalAccountID string         `json:"globalAccountID"`
```
ebensom, line 158: Could you wrap GlobalAccountID and Plan into 
```
type MaintenancePolicyMatch struct {
	GlobalAccountID string
	Plan            string
}
```

and include it in the `MaintenancePolicyEntry` like this:
```
type MaintenancePolicyEntry struct {
	Match MaintenancePolicyMatch `json:"match"`
...
```

The reason is that we will enhance the policy later on with a NotMatch specification.

components/kyma-environment-broker/common/orchestration/dto.go
```diff
@@ -153,3 +153,15 @@ type StatusResponseList struct {
...
+	Plan            string         `json:"plan"`
+	Days            []time.Weekday `json:"days"`
+	TimeBegin       time.Time      `json:"timeBegin"`
+	TimeEnd         time.Time      `json:"timeEnd"`
+}
+
+type MaintenancePolicy struct {
+	Rules []MaintenancePolicyEntry `json:"rules"`
```
ebensom, line 166: Just when giving a second thought, we should support a configurable default rule here:
```
type MaintenancePolicy struct {
	...
	Default MaintenancePolicyEntry `json:"rules"`
```

And the logic in manager should be that after iterating over the rules and no rules matched, we propagate any optional days/begin/end attributes if specified.

components/kyma-environment-broker/common/orchestration/resolver.go
```diff
@@ -132,7 +132,7 @@ func (resolver *GardenerRuntimeResolver) getRuntime(runtimeID string) (runtime.R
 
 func (resolver *GardenerRuntimeResolver) resolveRuntimeTarget(rt RuntimeTarget, shoots []gardenerapi.Shoot) ([]Runtime, error) {
 	runtimes := []Runtime{}
-
+	// r.Plan = resolver.runtimes[r.RuntimeID].ServicePlanName
```
ebensom, line 135: Remove instead of uncomment

components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -39,7 +50,18 @@ func (m *orchestrationManager) Execute(orchestrationID string) (time.Duration, e
...
+	key := client.ObjectKey{Namespace: m.policyNamespace, Name: m.policyName}
+	if err := m.k8sClient.Get(m.ctx, key, config); err != nil {
+		m.log.Info("Orchestration Config is absent")
+	}
+	if config.Data[maintenancePolicyKeyName] == "" {
+		m.log.Info("Maintenance policy is set to Gardener defaults")
+	}
+
+	var policies []orchestration.MaintenancePolicyEntry
+	json.Unmarshal([]byte(config.String()), &policies)
```
ebensom, line 63: Can throw error, need to check

ebensom (approved):

PR #831, opened 2021-07-07
[WIP] Accept BTP Operator credentials on update endpoint

wozniakjan:
**Description**

Changes proposed in this pull request:

- allow ERS to send new btp-operator credentials on update

**Related issue(s)**

followup to: https://github.com/kyma-project/control-plane/pull/825

PR #836, opened 2021-07-09
Provisioner endpoint to install just additional runtime components

wozniakjan:
**Description**

- new provisioner GraphQL mutation query
- this will allow installing a runtime component without the need to specify configs for all other unrelated components

**Related issue(s)**

will be used by: https://github.com/kyma-project/control-plane/pull/831

### DIFF
--- a/components/provisioner/internal/api/resolver.go
+++ b/components/provisioner/internal/api/resolver.go
@@ -108,6 +108,25 @@ func (r *Resolver) UpgradeRuntime(ctx context.Context, runtimeId string, input g
 	return operationStatus, nil
 }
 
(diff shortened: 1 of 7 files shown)


ebensom:
Not sure what this PR is needed for, but for Kyma 2.0 all additional components (BTP operator, compass runtime agent, monitoring agent) will have to be managed by reconciler , and reconcile state back upon deviatipn from desired state. So provisioner should jot manage any components

wozniakjan:
> Not sure what this PR is needed for, but for Kyma 2.0 all additional components

I was informed we need a path to migrate even if kyma 2.0 is not ready by the time of btp-operator migration, perhaps @PK85 can explain more around the calendar and deadlines, it came from him as a request.

wozniakjan:
The timeline got updated and the provisioner will indeed no longer provision kyma when the migration is supposed to happen. Closing as no longer necessary

PR #850, opened 2021-07-19
Remove async execution of suspension context update in e2e test

ksputo:
**Description**
Suspension/unsuspension update calls are executed without `?async_required=true` parameter in real scenario. Asynchronous execution of the update is only required when passing `parameters` in update call payload - in case of supsension/unsuspension, we only pass `context` and this has to be synchronous to reflect the real scenario.

- Removed `async_required=true` parameter from URL when calling update (suspension and unsuspension, context update)
- nit: use `InstanceID(), ClusterName()` methods to access `client` non-exported fields

**Related issue(s)**

### DIFF
--- a/resources/kcp/values.yaml
+++ b/resources/kcp/values.yaml
@@ -31,7 +31,7 @@ global:
         version: "PR-699"
       e2e_provisioning:
         dir:
-        version: "PR-844"
+        version: "PR-850"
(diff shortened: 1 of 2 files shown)


wozniakjan (approved):
/lgtm

PK85 (approved):

PR #851, opened 2021-07-20
use plan_id as metric label instead of using plan_name in metric name

life0215:
…

**Description**

- consolidate plan_id usage in metrics, propose to use plan_id as metric label instead of using plan_name as part of metric name

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/metrics/operations_collector.go
+++ b/components/kyma-environment-broker/internal/metrics/operations_collector.go
@@ -60,34 +60,34 @@ func NewOperationsCollector(statsGetter OperationsStatsGetter) *OperationsCollec
 	for _, p := range supportedPlansIDs {
 		opStats[p] = OperationStat{
 			inProgressProvisioning: prometheus.NewDesc(
-				fqName(broker.PlanNamesMapping[p], internal.OperationTypeProvision, domain.InProgress),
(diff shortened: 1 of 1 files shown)


ebensom (approved):

klaudiagrz:
@life0215 Make sure to keep both plan_id and plan_name for not breaking compatibility

life0215:
> @life0215 Make sure to keep both plan_id and plan_name for not breaking compatibility


We have only one grafana [dashboard](https://github.com/kyma-project/control-plane/blob/main/resources/kcp/charts/kyma-environment-broker/files/dashboard.json#L915) which consumes these metrics and expecting KEB metrics to have consolidated labels and metric naming( i.e. changes in this PR).
So we are not expecting breaking compatibility. 

Unfortunately keeping the plan_name in metric names will make this PR meaningless. We still need to deal with two sets of variables plan_name, plan_id. which is time consuming and hard to maintain.

Do we have any other service which are integrated with these metrics ?

szwedm:
Looks good, I just wonder if image should be bumped?

szwedm:
One more question, was the dashboard adjusted to incorporate these changes ?

szwedm (approved):

life0215:
> @life0215 one more question, was the dashboard adjusted to incorporate these changes ?

Thanks, am working on the adjustment.

PR #866, opened 2021-07-26
 Change StatusResponse struct to have targets and details info

berkanercan:
**Description**

The kcp orchestrations (list) default table output should be extended with the TARGETS and DETAILS columns:

```
ORCHESTRATION ID                       TYPE              CREATED AT            STATE       DRY RUN  TARGETS      DETAILS
15d5be2b-6426-4788-a8f4-c1a8e5f44107   cluster upgrade   2021/06/08 07:53:51   failed      false    all          K8S: 1.18.18
de8850b1-a5c4-42a7-86fb-cc07f7c1202f   kyma upgrade      2021/07/20 06:25:29   succeeded   false    trial, free  Kyma: 1.24.0
```
The targets are an already existing attribute of the orchestration.StatusResponse struct. In the list colums, all values should be joined together with a ", ", and the result string should be truncated to 20 characters, including the leading "..."

The details should also be displayed by orchestration show (kcp orchestration {ID}) command.

This PR only contains the implementation partially. Please check https://github.com/kyma-project/control-plane/pull/908 to see the full implementation


components/kyma-environment-broker/common/orchestration/dto.go
```diff
@@ -104,15 +104,27 @@ type TargetSpec struct {
...
+	OrchestrationID string                 `json:"orchestrationID"`
+	Type            Type                   `json:"type"`
+	State           string                 `json:"state"`
+	Description     string                 `json:"description"`
+	CreatedAt       time.Time              `json:"createdAt"`
+	UpdatedAt       time.Time              `json:"updatedAt"`
+	Parameters      Parameters             `json:"parameters"`
+	OperationStats  map[string]int         `json:"operationStats,omitempty"`
+	KymaDetails     *KymaDetailResponse    `json:"kymaDetails,omitempty"`
+	ClusterDetails  *ClusterDetailResponse `json:"clusterDetails,omitempty"`
```
wozniakjan, line 127: is this sufficient? I would assume you need to populate those fields somewhere too
https://github.com/kyma-project/control-plane/blob/88243fac522464d9348c2e276fa1ddf614f7c1cb/components/kyma-environment-broker/internal/orchestration/handlers/converter.go#L13-L24

wozniakjan:
I just wanted to copy the KEB debugging workflow and testing here too, so you don't need to dig it from slack
the gist of it is:
* point your kubecontext to `gke_sap-ti-dx-kyma-mps-dev_europe-west1_kyma-mps-dev`
* in the `kcp-system` namespace, there is a deployment `kcp-kyma-environment-broker`
* you can get the deployment locally, edit, add a suffix `*-be` to the name, the convention is appending your initials but in case of a clash, feel free to add more letters to remain unique
* also make sure to edit the label selector and labels, ideally with the same suffix
* edit `APP_DISABLE_PROCESS_OPERATIONS_IN_PROGRESS` env variable from false to true
```
        - env:
          - name: APP_DISABLE_PROCESS_OPERATIONS_IN_PROGRESS
            value: "true"
```
* you can do `$ kubectl port-forward deployment/[your-keb-deployment-name] 8080:8080` and then directly call curl to localhost:8080. No need to pass any authentication primitives as that layer is outside of KEB.
* link to swagger UI for KEB https://kyma-env-broker.cp.dev.kyma.cloud.sap/
* example request to locally portforwarded endpoint
```
$ curl -XGET "localhost:8080/oauth/v2/service_instances/[skr-instance-name]" -i -H "X-Broker-API-Version: 2.14"
```

berkanercan:
Changes are tested on `kcp-mps-dev` cluster with my own deployment (`kcp-kyma-environment-broker-be`). Please see the result of an orchestration. It has kubernetes and kyma versions included. 
```
{
  "orchestrationID": "f0115d65-8ca2-4c64-9c86-8b67bf87ce5f",
  "type": "upgradeCluster",
  "state": "succeeded",
  "description": "Scheduled 0 operations",
  "createdAt": "2021-09-20T09:35:13.696232Z",
  "updatedAt": "2021-09-20T09:35:13.696232Z",
  "parameters": {
    "targets": {
      "include": [
        {
          "target": "all",
          "globalAccount": "054ac2c2-318f-45dd-855c-eee41513d40d",
          "subAccount": "054ac2c2-318f-45dd-855c-eee41513d40d",
          "region": "europe|eu-",
          "runtimeID": "054ac2c2-318f-45dd-855c-eee41513d40d",
          "planName": "azure",
          "shoot": "c-0ab3fe0"
        }
      ],
      "exclude": [
        {
          "target": "all",
          "globalAccount": "054ac2c2-318f-45dd-855c-eee41513d40d",
          "subAccount": "054ac2c2-318f-45dd-855c-eee41513d40d",
          "region": "europe|eu-",
          "runtimeID": "054ac2c2-318f-45dd-855c-eee41513d40d",
          "planName": "azure",
          "shoot": "c-0ab3fe0"
        }
      ]
    },
    "strategy": {
      "type": "parallel",
      "schedule": "immediate",
      "parallel": {
        "workers": 1
      }
    },
    "Kubernetes": {
      "kubernetesVersion": "1.19.13"
    },
    "Kyma": {
      "kymaVersion": "1.24.4"
    }
  }
}
```

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -12,6 +12,8 @@ import (
 	"sort"
 	"time"
 
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/auditlog"
```
wozniakjan, line 15: nit: can you please put this import back to among the other control-plane imports?

wozniakjan (commented):
good progress, there are two more details that you can either implement as part of this PR or create followup PRs
* bump the KEB image version (it is still manual, sorry for that), similar to https://github.com/kyma-project/control-plane/pull/920/commits/4638f46478e7b7fd655eb1ddf48e590947511cc8
* reflect the changes in swagger (also manual, also apologies for the inconvenience) similar to https://github.com/kyma-project/control-plane/pull/920/commits/95bcd3cf9ef84e4230b93459c304d7cd8724c2fb

@ebensom would you like to verify this implements the functionality you desired? If this is good for you we can merge.

components/kyma-environment-broker/common/orchestration/dto.go
```diff
@@ -8,13 +8,18 @@ import (
...
 type Parameters struct {
-	Targets  TargetSpec   `json:"targets"`
-	Strategy StrategySpec `json:"strategy,omitempty"`
-	DryRun   bool         `json:"dryRun,omitempty"`
+	Targets    TargetSpec           `json:"targets"`
+	Strategy   StrategySpec         `json:"strategy,omitempty"`
+	DryRun     bool                 `json:"dryRun,omitempty"`
+	Kubernetes KubernetesParameters `json:""`
 	// upgrade kyma specific parameters
 	Kyma KymaParameters `json:""`
```
ebensom, line 16: I think because there will be quite a few kyma & kubernetes parameters, have the block set in the JSON:

**BREAKING CHANGE, so CLI must be recompiled with the new common/orchestration package** (which you will anyway do as CLI changes are coming)

```suggestion
	Kubernetes *KubernetesParameters `json:"kubernetes,omitempty"`
	// upgrade kyma specific parameters
	Kyma *KymaParameters `json:"kyma,omitempty"`
```

components/kyma-environment-broker/common/orchestration/dto.go
```diff
@@ -104,15 +109,27 @@ type TargetSpec struct {
 	Exclude []RuntimeTarget `json:"exclude,omitempty"`
 }
 
+type KymaDetailResponse struct {
```
ebensom, line 112: I would reuse the `Paramaters.Kyma` and `Parameters.Kubernetes` for sending kyma and kubernetes info in the response, and wouldn't introduce new structs for the response

components/kyma-environment-broker/common/orchestration/dto.go
```diff
@@ -8,13 +8,18 @@ import (
...
 	Kyma KymaParameters `json:""`
 }
 
+type KubernetesParameters struct {
+	Version string `json:"kubernetesVersion,omitempty"`
+}
+
 // KymaParameters hold the attributes of kyma upgrade specific orchestration create requests.
 type KymaParameters struct {
 	Version string `json:"kymaVersion,omitempty"`
```
ebensom, line 25: Since we now don't embed Kyma into Parameters, we can shorten the JSON field name in Kyma:

```suggestion
	Version string `json:"version"`
```

components/kyma-environment-broker/common/orchestration/dto.go
```diff
@@ -8,13 +8,18 @@ import (
...
+	Targets    TargetSpec           `json:"targets"`
+	Strategy   StrategySpec         `json:"strategy,omitempty"`
+	DryRun     bool                 `json:"dryRun,omitempty"`
+	Kubernetes KubernetesParameters `json:""`
 	// upgrade kyma specific parameters
 	Kyma KymaParameters `json:""`
 }
 
+type KubernetesParameters struct {
+	Version string `json:"kubernetesVersion,omitempty"`
```
ebensom, line 20: Adding all the fields from `ClusterDetailResponse`
```suggestion
	KubernetesVersion   string `json:"kubernetesVersion"`
	MachineImage        string `json:"machineImage"`
	MachineImageVersion string `json:"machineImageVersion"`
```

components/kyma-environment-broker/internal/broker/config.go
```diff
@@ -0,0 +1,6 @@
+package broker
+
+type KEBConfig struct {
```
ebensom, line 3: I feel the purpose of this new config is a bit confusing. How about sending down the whole `cmd/broker.Config` struct from `main()` to the orchestration managers, and would just use that.

components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -60,6 +63,9 @@ func (m *orchestrationManager) Execute(orchestrationID string) (time.Duration, e
 		return m.failOrchestration(o, errors.Wrap(err, "while resolving operations"))
 	}
 
+	o.Parameters.Kyma.Version = m.cfg.KymaVersion
```
ebensom, line 66: This code should be moved to `resolveOperations()` in the branch which handles the `Pending` (to be started) orchestration. 
Teh reason is that if the orchestration is not pending (i.e. it is a resumed in progress after restart), it would overwrite parameters from config. Furthermore the parameters should only be set to the default config values, if they are empty, i.e. for each parameter it was not provided as input.

components/kyma-environment-broker/internal/broker/config.go
berkanercan: I previously moved the whole `cmd/broker.Config` under /internal/broker and it caused a import cycle. Standalone KEBConfig was my solution to avoid that. I will try to move the whole `cmd/broker.Config` under orchestration/manager

components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -121,6 +124,13 @@ func (m *orchestrationManager) resolveOperations(o *internal.Orchestration, poli
 			return result, errors.Wrap(err, "while resolving targets")
 		}
 
+		if o.Parameters.Kyma.Version == "" {
```
ebensom, line 127: One tiny thing, and the PR can be merged:

These paramaters sets should go after the `for _, r := range runtimes {` loop below, because in upgrade kyma factory's `NewOperation()` there is a code which checks if an input version was provided for the orchestration, and calculates oper runtime target version accordingly:
https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/internal/orchestration/manager/upgrade_kyma.go#L74

We must not trigger this code accidentally with the default kyma version.

ebensom (approved):

ebensom (approved):

PR #874, opened 2021-08-02
 Sync the node autoscale params during KEB cluster upgrade

vvxxvvxx:
**Description**

- Set/Sync the autoscale params (AutoScalerMin, AutoScalerMax, MaxSurge, MaxUnavailable) from the plan file in [plan provider](https://github.com/kyma-project/control-plane/tree/main/components/kyma-environment-broker/internal/provider) during the KEB cluster upgrade.
- Add the unit subtests related to this autoscaler settings change during the cluster upgrade.

**Related issue(s)**


piotrmiskiewicz:
what is the purpose of this PR? why do we need autoscalers parameters in the upgrade request?

components/kyma-environment-broker/internal/process/input/builder_test.go
```diff
@@ -215,6 +216,70 @@ func TestInputBuilderFactory_ForPlan(t *testing.T) {
 		assert.Equal(t, gqlschema.KymaProfileEvaluation, *result.upgradeRuntimeInput.KymaConfig.Profile)
 	})
 
+	t.Run("should build CreateUpgradeShootInput with proper plan", func(t *testing.T) {
```
piotrmiskiewicz, line 219: let's name it: should build CreateUpgradeShootInput with proper autoscaler parameters

piotrmiskiewicz: please, create a test similar to that, but provisioning parameters should contains autoscaler params. such test will show us that the user's values are used instead of defaults

components/kyma-environment-broker/internal/process/input/builder_test.go
```diff
@@ -215,6 +216,70 @@ func TestInputBuilderFactory_ForPlan(t *testing.T) {
...
+		maxSurge := *result.upgradeShootInput.GardenerConfig.MaxSurge
+		maxUnavailable := *result.upgradeShootInput.GardenerConfig.MaxUnavailable
+
+		assert.Equal(t, autoscalerMax, provider.Defaults().GardenerConfig.AutoScalerMax)
+		assert.Equal(t, autoscalerMin, provider.Defaults().GardenerConfig.AutoScalerMin)
+		assert.Equal(t, maxSurge, provider.Defaults().GardenerConfig.MaxSurge)
+		assert.Equal(t, maxUnavailable, provider.Defaults().GardenerConfig.MaxUnavailable)
+		t.Logf("%v, %v, %v, %v", autoscalerMax, autoscalerMin, maxSurge, maxUnavailable)
+
+		// given
```
piotrmiskiewicz, line 252: we don't need separate test for trial. One test (for one plan) is ok for me

piotrmiskiewicz:
@ebensom  - we have already discussed with @vvxxvvxx and agreed with such solution, thanks

ebensom (dismissed):

ebensom (approved):

PR #894, opened 2021-08-18
feature: VictoriaMetrics integration as component

AdrianPei:
**Description**
set up VictoriaMetrics as component during provisioning, deprovisioning and upgrade

- install VictoriaMetrics via helm chart during provisioning, overrides username and password in runtime-monitoring-agent component
- uninstall VictoriaMetrics via helm chart during deprovisioning if existence
- upgrade VictoriaMetrics via helm chart during upgrade if existence
- store username and password in InstanceDetails

**Related issue(s)**

https://github.tools.sap/kyma/backlog/issues/1450


components/kyma-environment-broker/internal/process/provisioning/vmuser_overrides.go
```diff
@@ -0,0 +1,125 @@
...
+		},
+		{
+			Key:   "vmuser.password",
+			Value: vmPassword,
+		},
+	}
+	operation.InputCreator.AppendOverrides(VMAuthComponentName, VMAuthOverrides)
+
+	retry := time.Duration(0)
+	operation, retry = s.operationManager.UpdateOperation(operation, func(operation *internal.ProvisioningOperation) {
```
ebensom, line 84: We should only save the monitoring data when a new password is generated, updating the record at every try could lead to infinite operation processing

components/kyma-environment-broker/internal/process/upgrade_kyma/vmuser_upgrade.go
```diff
@@ -0,0 +1,100 @@
...
+	for _, rel := range releases {
+		if rel.Name == releaseName {
+			isPresent = true
+		}
+	}
+	if err != nil {
+		return s.handleError(operation, err, "err while getting release", log)
+	}
+	if !isPresent {
+		log.Info("Release no longer exists. Skip unistallation.")
```
ebensom, line 57: If the monitoring doesn't exist yet for the runtime, we should install it. This is how we will roll out for existing runtimes.

components/kyma-environment-broker/internal/vmauthoverrides/mock/client_fake.go
```diff
@@ -0,0 +1,98 @@
+package mocks
```
ebensom, line 1: This mock code can be automatically be generated with e.g. `mockery`, it is widely used in KEB. See e.g. [this](https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/common/orchestration/resolver.go#L20-L23) as an example

components/kyma-environment-broker/internal/vmauthoverrides/vmauthoverrides.go
```diff
@@ -0,0 +1,190 @@
...
+	"k8s.io/klog"
+)
+
+const (
+	Driver     = "memory"
+	MaxHistory = 10
+)
+
+type Config struct {
+	Namespace       string `envconfig:"default=mornitoring-system"`
```
ebensom, line 28:
```suggestion
	Namespace       string `envconfig:"default=monitoring-system"`
```

components/kyma-environment-broker/internal/vmauthoverrides/vmauthoverrides.go
```diff
@@ -0,0 +1,190 @@
...
+	ChartUrl        string `envconfig:"default=https://storage.googleapis.com/kyma-mps-dev-artifacts/rmi-0.1.0.tgz"`
```
ebensom, line 29: This must not have default value to avoid accidents.

components/kyma-environment-broker/internal/vmauthoverrides/vmauthoverrides.go
```diff
@@ -0,0 +1,190 @@
...
+	RemoteWriteUrl  string `envconfig:"default=http://vminsert-skr.monitoring-system.svc:8480/insert/0/prometheus"`
+	RemoteWritePath string `envconfig:"default=/api/v1/write"`
+	Disabled        bool   `envconfig:"default=false"`
```
ebensom, line 32:
```suggestion
	Disabled        bool   `envconfig:"default=true"`
```

components/kyma-environment-broker/internal/vmauthoverrides/vmauthoverrides.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return cfg, nil
+}
+
+func GeneratePassword(isPresent bool, operation internal.ProvisioningOperation) string {
+	var VMPassword = ""
+	if isPresent {
+		VMPassword = operation.VMUser.Password
+	} else {
+		s1 := strings.Split(operation.ProvisioningParameters.ErsContext.GlobalAccountID, "-")
```
ebensom, line 170: We should not assume that the global account ID is always in UUIDv4 format. 
I'd implement here a simple random string generation instead and not rely on `internal.ProvisioningOperation` data at all.

components/kyma-environment-broker/internal/vmauthoverrides/vmauthoverrides.go
```diff
@@ -0,0 +1,190 @@
...
+		s2 := strings.Split(operation.InstanceDetails.SubAccountID, "-")
+		s3 := strings.Split(operation.InstanceID, "-")
+		VMPassword = s1[0] + s2[0] + s3[0]
+	}
+	return VMPassword
+}
+
+func getProfileValues(chart chart.Chart, profileName string) (map[string]interface{}, error) {
```
ebensom, line 178: This chart will never have profile preset values, it's only true for kyma component charts. We can remove this code as well as the MergeMaps utility.

ebensom (commented):
/lgtm
/ok-to-test
/test all

PK85 (changes requested):
Please see: https://github.tools.sap/kyma/backlog/issues/1450#issuecomment-331719

AdrianPei:
@PK85 as discussed, this PR will be used temporarily until the functionality is migrated to component reconciler.

components/kyma-environment-broker/internal/process/provisioning/monitoring_integration.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+func (s *MonitoringIntegrationStep) Name() string {
+	return "Monitoring_Integration"
+}
+
+func (s *MonitoringIntegrationStep) Run(operation internal.ProvisioningOperation, log logrus.FieldLogger) (internal.ProvisioningOperation, time.Duration, error) {
+	releaseName := operation.InstanceDetails.ShootName
+	isPresent, err := s.client.IsInstalled(releaseName)
```
ebensom, line 40: If the last revision of the release is not deployed (e.g. pending, failed) for some reason, this would still return true and the step wouldn't retry installing.
For uninstalling, checking all possible states (deployed, pending, failed, superseded) is fine. But when trying to install I think we only need to consider deployed status (maybe change the existing `IsInstalled()` to `IsPresent()`, and have a separate `IsDeployed()` method?), and (re)try installation with `upgrade` helm option set (as the release might be present).

resources/kcp/charts/kyma-environment-broker/values.yaml
```diff
@@ -193,6 +193,13 @@ ems:
 connectivity:
   disabled: true
 
+monitoring:
+  namespace: "kcp-system"
+  chartUrl: "https://storage.googleapis.com/kyma-mps-dev-artifacts/rmi-0.1.0.tgz"
```
ebensom, line 198: I think we should remove these `chartUrl` `remoteWriteUrl` default values from OSS KCP code

components/kyma-environment-broker/internal/process/provisioning/monitoring_integration.go
AdrianPei: change install process as suggested

ebensom (approved):

components/kyma-environment-broker/internal/process/deprovisioning/monitoring_uninstall.go
```diff
@@ -0,0 +1,59 @@
...
+	} else {
+		log.Info("Release not found. Skip unistallation.")
+	}
+	return operation, 0, nil
+}
+
+func (s *MonitoringUnistallStep) handleError(operation internal.DeprovisioningOperation, err error, msg string, log logrus.FieldLogger) (internal.DeprovisioningOperation, time.Duration, error) {
+	log.Errorf("%s: %s", msg, err)
+	switch {
+	case kebError.IsTemporaryError(err):
```
piotrmiskiewicz, line 54: the client does not returns Temporary error so in case of a network tempoary issue we are failing the deprovisioning operation

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -636,6 +642,11 @@ func NewProvisioningProcessingQueue(ctx context.Context, provisionManager *provi
 			stage: createRuntimeStageName,
 			step:  provisioning.NewBusolaMigratorOverridesStep(),
 		},
+		{
+			stage:    createRuntimeStageName,
```
piotrmiskiewicz, line 646: after the merg of the [PR](https://github.com/kyma-project/control-plane/pull/912) we need to add a condition which will enable the step only for kyma 2.0

components/kyma-environment-broker/internal/process/provisioning/monitoring_integration.go
```diff
@@ -0,0 +1,111 @@
...
+		}, log)
+		if retry > 0 {
+			return operation, time.Second, nil
+		}
+	} else {
+		log.Info("Release already deployed.")
+		vmPassword = operation.Monitoring.Password
+	}
+
+	log.Info("Override username and password")
```
piotrmiskiewicz, line 87: change to debug

components/kyma-environment-broker/internal/process/provisioning/monitoring_integration.go
```diff
@@ -0,0 +1,111 @@
...
+			log.Info("Release found but not deployed. Ready to upgrade.")
+			_, err = s.client.UpgradeRelease(params)
+		}
+		if err != nil {
+			return s.handleError(operation, err, "failed to deploy chart", log)
+		}
+		retry := time.Duration(0)
+		operation, retry = s.operationManager.UpdateOperation(operation, func(operation *internal.ProvisioningOperation) {
+			operation.Monitoring.Username = operation.InstanceID
+			operation.Monitoring.Password = vmPassword
```
piotrmiskiewicz, line 77: this value is stored in the database as a plain text. Is it a good candidate to be encrypted (some pieces of data are already being encrypted)? @ebensom ?

ebensom: I was also thinking of this, but we are going to move this step from KEB and implement it in component reconciler. So until then, it can stay as it is.

components/kyma-environment-broker/internal/process/deprovisioning/monitoring_uninstall.go
ebensom: https://github.com/kyma-project/control-plane/pull/934

components/kyma-environment-broker/internal/process/provisioning/monitoring_integration.go
ebensom: https://github.com/kyma-project/control-plane/pull/934

components/kyma-environment-broker/cmd/broker/main.go
ebensom: We would also like to install agent for central monitoring for existing 1.X SKRs in stage so that we can cross verify the new monitoring functionality with the existing one.

PR #898, opened 2021-08-20
Modify openstack provider configuration for CCEE

anishj0shi:
- To expose the APIServer over the public internet, We have to modify Exposure Class in the shoot spec
- FloatingIP Network for CCEE is `FloatingIP-external-cp-kyma`


voigt:
This is done (based on @anishj0shi's PR) in https://github.com/kyma-project/control-plane/pull/915

PR #917, opened 2021-09-13
Connectivity bind step applies configuration overrides

Maladie:
**Description**

- Connectivity bind step applies configuration overrides
- Tests added

**Related issue(s)**


components/kyma-environment-broker/internal/process/provisioning/connectivity_bind.go
```diff
@@ -106,82 +86,24 @@ func (s *ConnectivityBindStep) Run(operation internal.ProvisioningOperation, log
 		operation = op
 	} else {
 		// get the credentials from encrypted string in operation.Connectivity.Instance.
```
koala7659, line 88: I'm afraid we will never come into this place. Returning  `operation, 0, nil` in line 98 after first pass will make Executor to think this step is completed and the move to the next step of processing the operation. See line 94 from the code of `components/kyma-environment-broker/internal/process/upgrade_kyma/manager.go`

```
for _, weightStep := range m.sortWeight() {
		steps := m.steps[weightStep]
		for _, step := range steps {
			logStep := logOperation.WithField("step", step.Name())
			logStep.Infof("Start step")

			operation, when, err = m.runStep(step, operation, logStep)
			if err != nil {
				logStep.Errorf("Process operation failed: %s", err)
				return 0, err
			}
			if operation.IsFinished() {
				logStep.Infof("Operation %q got status %s. Process finished.", operation.Operation.ID, operation.State)
				return 0, nil
			}
// this condition will be true after first pass and we will move to the next step 
			if when == 0 {
				logStep.Info("Process operation successful")
				continue
			}

			logStep.Infof("Process operation will be repeated in %s ...", when)
			return when, nil
		}
	}
````

Comment (koala7659) [reply components/kyma-environment-broker/internal/process/provisioning/connectivity_bind.go:88]: The second thing is we could maybe simplify it and update operation in one step just after extracting connectivity overrides. This will allow us to do not Encrypt/Decrypt credentials between two runs which is strange.


So the whole solution could looks like following:

```		
		respBinding, err := smCli.Bind(operation.Connectivity.Instance.InstanceKey(), operation.Connectivity.BindingID, nil, false)
		if err != nil {
			return s.handleError(operation, err, log, fmt.Sprintf("Bind() call failed"))
		}
		// get overrides
		connectivityOverrides, err = connectivity_bind.GetConnectivityCredentials(respBinding.Binding)
		if err != nil {
			return s.handleError(operation, err, log, fmt.Sprintf("unable to load config"))
		}

                 connectivityOverrides, err = connectivity_bind.GetConnectivityCredentials(respBinding.Binding)
		if err != nil {
			return s.handleError(operation, err, log, fmt.Sprintf("unable to load config"))
		}

     	// save the status
		op, retry := s.operationManager.UpdateOperation(operation, func(operation *internal.ProvisioningOperation) {
			applyOverrides(connectivityOverrides, *operation)
			operation.Connectivity.Instance.Provisioned = true
			operation.Connectivity.Instance.ProvisioningTriggered = false
		}, log)
		if retry > 0 {
			log.Errorf("unable to update operation")
		}
		operation = op

	}

	log.Debugf("Got Connectivity Service credentials from the binding.")

	return operation, 0, nil
```

File: components/kyma-environment-broker/internal/process/provisioning/connectivity_bind_test.go [hunk 2]
```diff
@@ -1,99 +1,120 @@
...
+	"github.com/stretchr/testify/require"
 )
 
-const connectivityServiceKey = `
-	{
-		"clientid" : "clientid314159265359",
-		"clientsecret": "Y2xpZW50c2VjcmV0Cg==",
-		"connectivity_service":
+func TestConnectivityBind_Run(t *testing.T) {
+	t.Run("Should succesfully apply overrides", func(t *testing.T) {
```
koala7659, line 15: There are two logic execution paths in tested step, dependent on the value of  `operation.Connectivity.Instance.Provisioned` We should test them both, that means first and second run. 

Also we could test the behaviour of retry after UpdateOperation is failed.

components/kyma-environment-broker/internal/process/upgrade_kyma/connectivity_bind.go
```diff
@@ -82,28 +83,29 @@ func (s *ConnectivityUpgradeBindStep) Run(operation internal.UpgradeKymaOperatio
 			operation.Connectivity.Instance.ProvisioningTriggered = false
 		}, log)
 		if retry > 0 {
+			log.Errorf("unable to update operation")
 			return operation, time.Second, nil
 		}
 		operation = op
 	} else {
 		// get the credentials from encrypted string in operation.Connectivity.Instance.
```
koala7659, line 91: Please check my 2 comments to `provisioning/connectivity_bind `step. They apply here as well.

components/kyma-environment-broker/internal/process/upgrade_kyma/connectivity_bind_test.go
```diff
@@ -1,101 +1,122 @@
...
+	"github.com/kyma-project/control-plane/components/provisioner/pkg/gqlschema"
+	"github.com/stretchr/testify/require"
 )
 
-const connectivityServiceKey = `
-	{
-		"clientid" : "clientid314159265359",
-		"clientsecret": "Y2xpZW50c2VjcmV0Cg==",
-		"connectivity_service":
+func TestUpgradeConnectivityBind_Run(t *testing.T) {
```
koala7659, line 15: Please see my comments to `provisioning/connectivity_bind_test.go` I think we should test here more logic paths. First pass/second pass/ retry.

koala7659 (changes requested):

PR #921, opened 2021-09-14
Adjust HA plans and AWS supported zones

ebensom:
**Description**

- Azure and AWS HA plans create clusters spanning 3 zones (currently min and max are both fixed to 3)
- Azure and AWS HA plans create one node per zone with 8 vCPU / 32 GiB RAM by default (`autoscalerMin.mininum` = `autoscalerMin.default` = 1)
- AWS HA plan support provisioning clusters with machine types larger than 8 vCPU / 32 GiB RAM 
- HA plans should never create more than 2 surge nodes per zone during cluster maintenance
- Fix AWS supported zones mapping. In https://github.com/kyma-project/control-plane/pull/916 zones for `ca-central-1`, `sa-east-1`, `ap-northeast-2`, `ap-south-1` were removed, assuming we can't support them because they don't support M4 machine types, however our AWS plan only use M5 machines which are supported by the removed zones.

**Related issue(s)**
https://github.tools.sap/kyma/backlog/issues/1308

### DIFF
--- a/components/kyma-environment-broker/cmd/broker/provisioning_test.go
+++ b/components/kyma-environment-broker/cmd/broker/provisioning_test.go
@@ -154,7 +154,7 @@ func TestProvisioning_ClusterParameters(t *testing.T) {
 
 			expectedMinimalNumberOfNodes:       1,
 			expectedMaximumNumberOfNodes:       10,
-			expectedMachineType:                "Standard_D4_v3",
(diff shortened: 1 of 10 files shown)


wozniakjan (approved):
/lgtm
/hold

you may want to fix the unit tests as well

ksputo (approved):
lgtm

Issue #946, opened 2021-09-30
Propagate errors for failed operations


wozniakjan:
**Description**

Currently, the errors from gardener do not get propagated all the way to the end-user. When user triggers an invalid operation, provisioner keeps trying it for a long period of time despite receiving a failure message from gardener. Eventually, KEB times out the operation and the user receives not very helpful information much later that the operation timed out.

Example:
1) call `PATCH` on `/v2/service_instances/$instance` with syntactically valid payload but semantically wrong
```json
...
     "parameters":{
         "name":"$instance",
         "autoScalerMin": 17,
         "autoScalerMax": 10 
     }
...
```
2) provisioner keeps retrying the operation and calls gardener, but it does receive failure response with a self-explanatory error message
```
time="2021-09-30T09:57:35Z" level=error msg="Internal Server Error: Failed to upgrade Cluster: error executing update shoot configuration, All attempts fail:\n#1: Shoot.core.gardener.cloud \"c33a972\" is invalid: spec.provider.workers[0].maximum: Forbidden: maximum value must not be less or equal than minimum value\n#2: Shoot.core.gardener.cloud \"c33a972\" is invalid: spec.provider.workers[0].maximum: Forbidden: maximum value must not be less or equal than minimum value\n#3: Shoot.core.gardener.cloud \"c33a972\" is invalid: spec.provider.workers[0].maximum: Forbidden: maximum value must not be less or equal than minimum value\n#4: Shoot.core.gardener.cloud \"c33a972\" is invalid: spec.provider.workers[0].maximum: Forbidden: maximum value must not be less or equal than minimum value\n#5: Shoot.core.gardener.cloud \"c33a972\" is invalid: spec.provider.workers[0].maximum: Forbidden: maximum value must not be less or equal than minimum value"
```

**Reasons**
The operations take a long time to time out and block any other operations on the same SKR because they get executed serially. Users are left in limbo trying to figure out what is wrong. Instead, we can give them the information as soon as we receive it.

**Attachments**

PR #920, opened 2021-09-13
Update endpoint for autoscaler options for KEB

wozniakjan:
**Description**

Enhance KEB instance update endpoint to allow users to modify autoscaler parameters

Sample test using port-forwarded deployment:
```
instance=test-jw34
curl -XPATCH "localhost:8080/oauth/v2/service_instances/$instance?accepts_incomplete=true" -i \
    -H "X-Broker-API-Version: 2.14" \
    -H "Content-Type: application/json" \
    --data-binary @- << EOF
{
   "service_id":"47c9dcbf-ff30-448e-ab36-d3bad66ba281",
   "plan_id":"4deee563-e5ec-4731-b9b1-53b42d855f0c",
   "context":{
       "globalaccount_id":"***",
       "subaccount_id":"***",
       "user_id":"jan.wozniak@sap.com"
   },
   "parameters":{
       "name":"$instance",
       "autoScalerMin": 10,
       "autoScalerMax": 11,
       "maxSurge": 12,
       "maxUnavailable": 13
   }
}
EOF
```

and then connecting to the broker SQL database
```
broker=> select type, finished_stages, provisioning_parameters from operations where instance_id='test-jw34' order by created_at desc limit 5;
   type    |    finished_stages    |   provisioning_parameters
-----------+-----------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 update    |                       | {"plan_id":"4deee563-e5ec-4731-b9b1-53b42d855f0c","service_id":"47c9dcbf-ff30-448e-ab36-d3bad66ba281","ers_context":{"tenant_id":"","subaccount_id":"***","globalaccount_id":"***","active":true,"user_id":"jan.wozniak@sap.com"},"parameters":{"name":"test-jw34","targetSecret":null,"volumeSizeGb":null,"machineType":null,"region":null,"purpose":null,"licence_type":null,"zones":null,"zonesCount":null,"autoScalerMin":10,"autoScalerMax":11,"maxSurge":12,"maxUnavailable":13,"components":null,"kymaVersion":"","overridesVersion":"","administrators":null,"provider":null},"platform_region":"cf-eu10","platform_provider":"unknown"}
[...]

PR #958, opened 2021-10-04
Makeshift check for valid autoscaler parameters in KEB

wozniakjan:
**Description**

Provisioner swallows errors from gardener which results in KEB failing the operation only on timeout. This patch provides a simple sanity check for a common error that would otherwise take an hour to fail to provision.

This can be safely removed once the following issue is resolved https://github.com/kyma-project/control-plane/issues/946

This is a follow-up for https://github.com/kyma-project/control-plane/pull/920

Example:

```bash
$ curl -XPATCH "localhost:8080/oauth/v2/service_instances/$instance?accepts_incomplete=true" -i \
    -H "X-Broker-API-Version: 2.14" \
    -H "Content-Type: application/json" \
    --data-binary @- << EOF
{
   "service_id":"47c9dcbf-ff30-448e-ab36-d3bad66ba281",
   "plan_id":"4deee563-e5ec-4731-b9b1-53b42d855f0c",    # azure, with defaults 2/10
   "context":{
       "globalaccount_id":"***",
       "subaccount_id":"***",
       "user_id":"***"
   },
   "parameters":{
       "name":"$instance",
       "autoScalerMin": 17 
   }
}
EOF

HTTP/1.1 422 Unprocessable Entity
Content-Type: application/json
X-Broker-Api-Request-Identity: <nil>
Date: Wed, 06 Oct 2021 07:04:40 GMT
Content-Length: 143

{"description":"AutoScalerMax 10 should be larger than AutoScalerMin 17. User provided values min:17, max:<nil>; plan defaults min:2, max:10"}
```


components/kyma-environment-broker/internal/dto.go
```diff
@@ -72,7 +73,28 @@ const (
...
+		return nil
+	}
+	if *p.AutoScalerMin > *p.AutoScalerMax {
+		return fmt.Errorf("Auto Scaler Max %v should be larger than Auto Scaler Min %v", *p.AutoScalerMax, *p.AutoScalerMin)
+	}
+	return nil
+}
+
 type ProvisioningParametersDTO struct {
+	AutoScalerParameters `json:",inline"`
```
wozniakjan, line 96: inlining and embedding `AutoScalerParams`, these are shared between provisioning and updating.

components/kyma-environment-broker/internal/dto.go
```diff
@@ -72,7 +73,35 @@ const (
...
+func (p AutoScalerParameters) Validate() error {
+	if p.AutoScalerMin == nil && p.AutoScalerMax == nil {
+		return nil
+	}
+	if p.AutoScalerMin == nil || p.AutoScalerMax == nil {
+		missing := "Min"
+		if p.AutoScalerMax == nil {
+			missing = "Max"
+		}
+		return fmt.Errorf("When setting Auto Scaler Min/Max you need to provide both, missing: %v", missing)
```
wozniakjan, line 94: because it is technically possible at the moment to provide only min or max which makes it much harder to validate because even the operation in the db won't specify any value. At that point, KEB was relying on provisioner/gardener setting defaults.

PK85: we should allow to update only min, or only max. If wrong values we should say why. Keb should take values from db and send to provisioner, we should log that user has changed only e.g. min.

wozniakjan: thanks for the input, I will try to figure something out then, but it is unfortunate that some defaulting is done in KEB, some in provisioner and some in gardener, it makes the design a little more convoluted.

The problem here is there is nothing in the database for the following scenario. If you set only `"autoScalerMin": 11` during provisioning, the operation in db would contain these values
```
broker=> select id, type, state, finished_stages, provisioning_parameters from operations where instance_id='test-jw40';
... "autoScalerMin":11,"autoScalerMax":null,"maxSurge":null,"maxUnavailable":null ...
```
and be forever in pending state because provisioner would default `autoScalerMax` to 10.
My naive idea is to take the provisioner hardcoded max as default in KEB too and if the user doesn't set max, set it to 10. If user sets min higher than 10, make max == min.

wozniakjan: I was wrong, it was explained to me that it is a plan defaulting that is modifying the value in the operations queue. Implemented reuse of this logic in https://github.com/kyma-project/control-plane/pull/958/commits/094b8ccdf71f6a07da372246ac5e72a832dbb55a

components/kyma-environment-broker/internal/broker/broker.go
```diff
@@ -14,6 +16,8 @@ const (
 	KymaServiceName = "kymaruntime"
 )
 
+type PlanDefaults func(planID string, platformProvider internal.CloudProvider, parametersProvider *internal.CloudProvider) (*gqlschema.ClusterConfigInput, error)
```
wozniakjan, line 19: importing `inputFactory` to call `GetHyperscalerProviderForPlanID` directly results in import cycle

```
env CGO_ENABLED=0 go build -o kyma-environment-broker ./cmd/broker/main.go
package command-line-arguments
        imports github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/appinfo
        imports github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/broker
        imports github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/process/input
        imports github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/broker: import cycle not allowed
make: *** [/home/wozy/projects/go/src/github.com/kyma-project/control-plane/scripts/generic_make_go.mk:121: build-local] Error 1
```

so I decided to wrap the call in a function and pass that to `ProvisionEndpoint` and `UpdateEndpoint`

piotrmiskiewicz (approved):

PR #964, opened 2021-10-06
Fix monitoring steps

ebensom:
**Description**

- Fill in InstanceDetails in `UpgradeKymaOperation` properly, i.e. set `ShootName` attribute in case it's missing
- Use subaccount ID from Provisioning parameters (due to data inconsistency in DB it could be empty for some operation records)
- Avoid updating DB at every run of upgrade kyma initialization step. This prevents operation timeout to work properly.

**Related issue(s)**


components/kyma-environment-broker/internal/process/deprovisioning/monitoring_uninstall.go
```diff
@@ -31,6 +31,9 @@ func (s *MonitoringUnistallStep) Name() string {
 
 func (s *MonitoringUnistallStep) Run(operation internal.DeprovisioningOperation, log logrus.FieldLogger) (internal.DeprovisioningOperation, time.Duration, error) {
 	releaseName := operation.InstanceDetails.ShootName
+	if releaseName == "" {
+		return s.operationManager.OperationFailed(operation, "rmi release name cannot be empty", log)
+	}
```
piotrmiskiewicz, line 36: do not fail a step in the deprovisioning operation. If this logic cannot be executed, we have to log it, but go to next step, see EDP deprovisioning step

piotrmiskiewicz (approved):

PR #966, opened 2021-10-06
Update swagger to reflect implementation of KEB

wozniakjan:
**Description**

The swagger schema is very out of sync with the current implementation of KEB. This is an effort to synchronize the two again.

**Related issue(s)**


resources/kcp/charts/kyma-environment-broker/files/swagger.yaml
```diff
@@ -2102,6 +2238,22 @@ components:
...
+          type: string
+        password:
+          type: string
+
+    ServiceManagerCredentials:
+      basic:
+        $ref: '#/components/schemas/ServiceManagerCredentialsBasic'
+
+    SMPlatformCredentials:
+      credentials:
```
ksputo, line 2253: `SMPlatformCredentials` should be typ `object` to render it's children properly

```suggestion
    SMPlatformCredentials:
      type: object
      credentials:
```

ksputo (approved):

PR #973, opened 2021-10-12
Keb component source for reconciler

piotrmiskiewicz:
**Description**

- provide a source URL of a component to reconciler

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/reconciler/contract.go
+++ b/components/kyma-environment-broker/internal/reconciler/contract.go
@@ -28,6 +28,7 @@ type Configuration struct {
 }
 
 type Components struct {
+	URL           string          `json:"URL"`
 	Component     string          `json:"component"`
 	Namespace     string          `json:"namespace"`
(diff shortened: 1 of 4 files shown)


components/kyma-environment-broker/internal/reconciler/contract.go
```diff
@@ -28,6 +28,7 @@ type Configuration struct {
 }
 
 type Components struct {
+	URL           string          `json:"URL"`
```
wozniakjan, line 31: just for my understanding, what was the reason of implementing the contract here as opposed to [importing from reconciler directly](https://github.com/kyma-incubator/reconciler/blob/587b592eaa2c82d7d923ffdf0297f04fe00dfae7/pkg/keb/model_gen.go#L74-L79)?

piotrmiskiewicz: I don't remember....

PR #980, opened 2021-10-14
Install btp-operator when credentials provided

ksputo:
**Description**

- Switch from mock to fake implementation of `componentListProvider` in provisioning test to have better insight of the built input
- Added new object to provisioning context: `ServiceManagerOperatorCredentials` containing new set of credentials for SM operator flow
- Disable service management components `helm-broker`, `service-catalog`, `service-catalog-addons`, `service-manager-proxy`, `btp-operator` in `initialisation` step and enable them only in appropriate scenario:
  - when "old", `sm_platform_credentials` object provided in context, install `helm-broker`, `service-catalog`, `service-catalog-addons` and `service-manager-proxy`
  - when "new", `sm_operator_credentials` object provided in context, install `btp-operator` component

**Related issue(s)**

tools/#1829


components/kyma-environment-broker/cmd/broker/provisioning_test.go
```diff
@@ -110,6 +110,78 @@ func TestProvisioningWithReconciler_HappyPath(t *testing.T) {
...
+		InstanceID:      iid,
+	})
+
+	suite.AssertClusterKymaConfig(opID, reconciler.KymaConfig{
+		Version:        "2.0",
+		Profile:        "Production",
+		Administrators: []string{"john.smith@email.com"},
+		Components: []reconciler.Components{
+			{
+				Component: "service-catalog2",
```
wozniakjan, line 165: nit: probably my lack of understanding of how this works but shouldn't btp-operator credentials in the request payload and Kyma 2.0 configured make it so no service catalogue is installed?

wozniakjan (commented):
this looks very good, just a few comments below

components/kyma-environment-broker/internal/process/provisioning/btp_operator_overrides.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+const BTPOperatorComponentName = "btp-operator"
+
+type BTPOperatorOverridesStep struct {
+	operationManager *process.ProvisionOperationManager
+}
+
+func NewBTPOperatorOverridesStep(os storage.Operations) *BTPOperatorOverridesStep {
```
wozniakjan, line 22: nit: I think `storage.Operations` is not necessary for the provisioning BTP Operator overrides

wozniakjan (approved):
/lgtm

PR #987, opened 2021-10-15
Reconciler payload persistence (memory)

szwedm:
**Description**

- add reconciler payload to storage.

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/storage/driver/memory/runtime_state.go
+++ b/components/kyma-environment-broker/internal/storage/driver/memory/runtime_state.go
@@ -59,3 +59,15 @@ func (s *runtimeState) GetByOperationID(operationID string) (internal.RuntimeSta
 
 	return internal.RuntimeState{}, dberr.NotFound("runtime state with operation ID %s not found", operationID)
 }
+
+func (s *runtimeState) GetLastByRuntimeID(runtimeID string) (internal.RuntimeState, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	states, err := s.ListByRuntimeID(runtimeID)
+	if err != nil {
+		return internal.RuntimeState{}, err
+	}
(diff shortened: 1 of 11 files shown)


piotrmiskiewicz (dismissed):

components/kyma-environment-broker/internal/storage/driver/memory/runtime_state.go
```diff
@@ -59,3 +59,12 @@ func (s *runtimeState) GetByOperationID(operationID string) (internal.RuntimeSta
...
 	return internal.RuntimeState{}, dberr.NotFound("runtime state with operation ID %s not found", operationID)
 }
+
+func (s *runtimeState) GetLastByRuntimeID(runtimeID string) (internal.RuntimeState, error) {
+	states, err := s.ListByRuntimeID(runtimeID)
+	if err != nil {
+		return internal.RuntimeState{}, err
+	}
+
+	return states[0], nil
```
wozniakjan, line 69: nit: when integrating this with https://github.com/kyma-project/control-plane/pull/825, I found it useful to guard for slice boundaries and return error self-explanatory error if the `states` slice was empty. I realize that practically this should never occur in real life unless someone truncates `runtime_states` db, but in case something really strange happens and `states` is empty, this code will throw panic when it can instead return `error`.

wozniakjan (dismissed):
works very well for me in https://github.com/kyma-project/control-plane/pull/825. Feel free to unhold if @piotrmiskiewicz has no more comments and you think it's ok to merge
/lgtm
/hold

components/kyma-environment-broker/internal/storage/driver/memory/runtime_state.go
szwedm: Good point, I will add a db content existence check

PR #1053, opened 2021-11-05
Fix containerd security issue

szwedm:
**Description**

- update go.mod and go.sum to include secure version of containerd dependency

**Related issue(s)**


components/kyma-environment-broker/go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/Masterminds/sprig v2.22.0+incompatible
 	github.com/Peripli/service-manager v0.19.4
 	github.com/Peripli/service-manager-cli v1.11.8
+	github.com/containerd/containerd v1.5.7 // indirect
```
wozniakjan, line 15: it is possible that now as a result of this the unit tests are failing
```
[0085] | # github.com/deislabs/oras/pkg/oras
1388
[0085] | vendor/github.com/deislabs/oras/pkg/oras/push.go:55:31: not enough arguments in call to remotes.PushContent
1389
[0085] | have (context.Context, remotes.Pusher, v1.Descriptor, "github.com/containerd/containerd/content".Store, nil, func(images.Handler) images.Handler)
1390
[0085] | want (context.Context, remotes.Pusher, v1.Descriptor, "github.com/containerd/containerd/content".Store, *semaphore.Weighted, platforms.MatchComparer, func(images.Handler) images.Handler)
1391
[0097] | ?   	github.com/kyma-project/control-plane/components/kyma-environment-broker/cmd/accountcleanup	[no test files]
1392
[0098] | FAIL	github.com/kyma-project/control-plane/components/kyma-environment-broker/cmd/broker [build failed]
```

wozniakjan (commented):
I suppose if the security issue flags unit tests as vulnerable, we can claim this false alarm because it's not in a live code for production deployment

components/kyma-environment-broker/go.mod
szwedm: I managed to update the dependencies without changing the code. As you mentioned this security vulnerability exists only in the unit tests code. The dependency chain is:
`
github.com/containerd/containerd@v1.5.0-beta.4 -> github.com/Microsoft/hcsshim@v0.8.16 -> github.com/testcontainers/testcontainers-go@v0.11.0
`
Testcontainers is only used in unit tests. We should decide whether we merge this PR with all the dependencies update or treat it as a false-alarm.

wozniakjan: I see value in this PR even if there is no know attack surface addressed. This also bumps a few of our outdated dependencies so let's get this in.

wozniakjan (approved):
/lgtm

PR #1066, opened 2021-11-09
Add reconciliation status check to KEB's upgrade process

ksputo:
**Description**

- Added `CheckClusterConfigurationStep` step to upgrade queue which gets the cluster state from reconciler to verify the reconciliation state, similar to provisioning queue
- Added `CreateClusterConfigurationStep` step to upgrade queue as first one after `initialisation` - this step sets proper values in operation.InputCreator: runtimeID, instanceID, shootName, shootDomain, ProvisioningParameters and enabled SM-components similarly to provisioning process. It also fetches latest runtimeState and rewrites the overrides for all the components from here **(TODO: filter out duplicated components overrides when applying new from `Overrides_From_Secrets_And_Config_Step` upgrade step - ticket)**

**Related issue(s)**


components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -783,36 +783,46 @@ func NewKymaOrchestrationProcessingQueue(ctx context.Context, db storage.BrokerS
...
 	}{
+		{
+			weight: 1,
+			step:   upgrade_kyma.NewCreateClusterConfiguration(db.Operations(), db.RuntimeStates(), reconcilerClient),
+			cnd:    upgrade_kyma.ForKyma2,
+		},
 		{
 			weight: 2,
-			step:   upgrade_kyma.NewOverridesFromSecretsAndConfigStep(db.Operations(), runtimeOverrides, runtimeVerConfigurator),
+			step:   upgrade_kyma.NewCheckClusterConfigurationStep(db.Operations(), reconcilerClient, 15*time.Minute),
```
piotrmiskiewicz, line 793: why we need this step here?

components/kyma-environment-broker/cmd/broker/upgrade_kyma_test.go
```diff
@@ -0,0 +1,89 @@
...
+						"subaccount_id": "sub-id",
+						"user_id": "john.smith@email.com"
+					},
+					"parameters": {
+						"name": "testing-cluster"
+					}
+		}`)
+	opID := suite.DecodeOperationID(resp)
+	suite.processProvisioningByOperationID(opID)
+	time.Sleep(1 * time.Second)
```
piotrmiskiewicz, line 45: remove the sleep

components/kyma-environment-broker/internal/process/upgrade_kyma/create_cluster_configuration.go
```diff
@@ -0,0 +1,90 @@
...
+		SetInstanceID(operation.InstanceID).
+		SetShootName(operation.InstanceDetails.ShootName).
+		SetShootDomain(operation.ShootDomain).
+		SetProvisioningParameters(operation.ProvisioningParameters)
+
+	// enable service management components for upgrade 1.x -> 2.0
+	operation.InputCreator.EnableOptionalComponent(provisioning.HelmBrokerComponentName)
+	operation.InputCreator.EnableOptionalComponent(provisioning.ServiceCatalogComponentName)
+	operation.InputCreator.EnableOptionalComponent(provisioning.ServiceCatalogAddonsComponentName)
+	operation.InputCreator.EnableOptionalComponent(provisioning.ServiceManagerComponentName)
```
piotrmiskiewicz, line 54: do we ha ve a logic for disabling it? if not, please, create a ticket

components/kyma-environment-broker/cmd/broker/main.go
ksputo: upgradeQueue does not have `stages` implemented like we do have in provisioning process. This step is at the beginning of the process, because in the first iteration of the step executor loop it won't be executed at all:
```go
	if operation.ClusterConfigurationVersion == 0 {
		// upgrade was trigerred in reconciler, no need to call provisioner and create UpgradeRuntimeInput
		// TODO: deal with skipping steps in case of calling reconciler for Kyma 2.0 upgrade - introduce stages
		log.Debugf("Cluster configuration not yet created, skipping")
		return operation, 0, nil
	}
```

`operation.ClusterConfigurationVersion` is 0 by default, and changes after `apply_cluster_configuration.go` step is executed:
```go
	updatedOperation, repeat := s.operationManager.UpdateOperation(operation, func(operation *internal.UpgradeKymaOperation) {
		operation.ClusterConfigurationVersion = state.ConfigurationVersion
	}, log)
```

`apply_cluster_configuration` being the last step in the queue returns some retry value:
```go
	// return some retry value to get back to initialisation step
	return updatedOperation, 5 * time.Second, nil
```

so we get back to the beginning - initilisation step gets skipped because `operation.ClusterConfigurationVersion` is already not equal `0`, same for `Create_Cluster_Configuration` step which leads us to execution of `Check_Cluster_Configuration` step again which starts fetching for the reconcilitation status.

piotrmiskiewicz (dismissed):

piotrmiskiewicz (approved):

components/kyma-environment-broker/internal/process/upgrade_kyma/create_cluster_configuration.go
ksputo: https://github.com/kyma-project/control-plane/issues/1098

PR #1077, opened 2021-11-11
bugfix: monitoring causing OOM during kyma upgrade

AdrianPei:
**Description**
fix the bug that central monitoring causing OOM during kyma upgrade

change the way to detect rma component installation status

**Related issue(s)**


components/kyma-environment-broker/internal/monitoring/monitoring.go
```diff
@@ -102,25 +103,14 @@ func (c *client) IsPresent(releaseName string) (bool, error) {
...
+		return false, kebError.AsTemporaryError(err, "unable to get helm history")
 	}
-	for _, rel := range releases {
-		if rel.Name == releaseName {
-			return true, nil
-		}
-	}
-
-	return false, nil
+	return true, nil
```
ebensom, line 113: You should also reimplement `IsDeployed` with a similar logic. After getting the helm release with the history action `.Run()`, you can check the current state of the release.

AdrianPei: IsDeployed also fixed

ebensom (commented):
/lgtm

You should also bump the KEB image in the chart

AdrianPei:
> /lgtm
> 
> You should also bump the KEB image in the chart

KEB image bumped

PR #1086, opened 2021-11-15
Refactor the KEB orchestration operations queuing logic

vvxxvvxx:
**Description**

- Use workqueue DelayingQueue to schedule the KEB orchestration operations to make sure the operations in the waiting queue can be handled in time.
- ...
- ...

**Related issue(s)**


components/kyma-environment-broker/common/orchestration/strategies/parallel.go
```diff
@@ -43,80 +46,146 @@ func (p *ParallelOrchestrationStrategy) Execute(operations []orchestration.Runti
...
+		p.mux.RLock()
+		if p.scheduleNum[execID] <= 0 {
+			p.mux.RUnlock()
+			dq.ShutDown()
+			pq.ShutDown()
+			break
+		}
+		p.mux.RUnlock()
+
+		if dq.ShuttingDown() || pq.ShuttingDown() {
```
ebensom, line 105: I think this is not needed because of the shutdown check on dq.Get() below. Once dq is marked for shutdown,. it will always instantly report shutdown=true for all calls to Get() once the dq is drained.
And dq has to be drained correctly in the case of orchestration cancel (i.e. calling `ParallelOrchestrationStrategy.Cancel()`), so we must not break prematurely.

vvxxvvxx: yes.. this is not needed. also thought about it but still keep it.. yes, Get() will also return shutdown as true and break.. will remove it.

vvxxvvxx: removed this condition and together with the break when `p.scheduleNum[execID] <= 0`

ebensom (approved):

PR #1090, opened 2021-11-15
Fix Operation Stop when reconciler returnrs 404

dbadura:
**Description**

- add 404 http status handling

**Related issue(s)**

See https://github.com/kyma-incubator/reconciler/issues/309


m00g3n (approved):
LGTM

PR #1124, opened 2021-11-22
KEB runtimes state: set to succeeded by default when operation state doesn't match

ebensom:
**Description**

This changes is to capture cases when the runtime state cannot be reliably determined from the state of the last operation. One example is when the last operation was `upgradeKyma` type, which was `canceled`. In these cases fall back by default to assume the runtime is in an active "succeeded" state.


wozniakjan (approved):
/lgtm

Issue #1096, opened 2021-11-16
After upgrade to 2.0 operation KymaConfig should not be empty


PK85:
**Description**

After upgrade we need to rewrite new reconciler.ClusterState into operation.KymaConfig.

**AC**

- After upgrade we need to rewrite new reconciler.ClusterState into operation.KymaConfig.

PR #1135, opened 2021-11-23
Fill KymaConfig after upgrade

szwedm:
**Description**

- add code to get KymaConfig from reconciler.Cluster

**Related issue(s)**

#1096

### DIFF
--- a/components/kyma-environment-broker/internal/model.go
+++ b/components/kyma-environment-broker/internal/model.go
@@ -431,9 +431,9 @@ func (r *RuntimeState) KymaConfigFromClusterSetup() {
 
 	profile := gqlschema.KymaProfile(r.ClusterSetup.KymaConfig.Profile)
 	kymaConfig := gqlschema.KymaConfigInput{
-		Version:          r.ClusterSetup.KymaConfig.Version,
-		Profile:          &profile,
-		Components:       components,
+		Version:    r.ClusterSetup.KymaConfig.Version,
+		Profile:    &profile,
+		Components: components,
 	}
 
 	r.KymaConfig = kymaConfig
--- a/components/kyma-environment-broker/internal/model.go
+++ b/components/kyma-environment-broker/internal/model.go
@@ -407,7 +407,14 @@ type RuntimeState struct {
 	ClusterSetup  *reconciler.Cluster           `json:"clusterSetup,omitempty"`
 }
 
-func (r *RuntimeState) KymaConfigFromClusterSetup() {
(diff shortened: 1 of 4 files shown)


components/kyma-environment-broker/internal/orchestration/handlers/orchestration_handler.go
```diff
@@ -203,6 +203,10 @@ func (h *orchestrationHandler) getOperation(w http.ResponseWriter, r *http.Reque
 		h.log.Errorf("while getting runtime state for upgrade operation %s: %v", operationID, err)
 	}
 
+	if upgradeState.ClusterSetup != nil {
+		upgradeState.KymaConfigFromClusterSetup()
```
wozniakjan, line 207: do we want to run this for upgrades not connected to any "orchestration" as well maybe?
https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/internal/broker/instance_update.go

wozniakjan (dismissed):
/lgtm
/hold

components/kyma-environment-broker/internal/orchestration/handlers/orchestration_handler.go
szwedm: it depends on whether other endpoints are showing KymaConfig in a response, like this one: https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/common/orchestration/dto.go#L147

wozniakjan: good point, I think they don't

components/kyma-environment-broker/internal/model.go
```diff
@@ -407,6 +407,38 @@ type RuntimeState struct {
...
+	}
+
+	profile := gqlschema.KymaProfile(r.ClusterSetup.KymaConfig.Profile)
+	kymaConfig := gqlschema.KymaConfigInput{
+		Version:    r.ClusterSetup.KymaConfig.Version,
+		Profile:    &profile,
+		Components: components,
+	}
+
+	r.KymaConfig = kymaConfig
```
piotrmiskiewicz, line 439: I'd like to see fields to be immutable, let's discuss

szwedm: I will change the method to return KymaConfig without modifying RuntimeState as we discussed

piotrmiskiewicz (approved):

PR #1147, opened 2021-11-26
Create new deprovisioning operation instead of retrying failed one

piotrmiskiewicz:
**Description**

- the deprovisioning retry must not restart existing failed deprovisioning operation but create a new one

**Related issue(s)**


ebensom (approved):
🚀

PR #1157, opened 2021-11-30
wip

piotrmiskiewicz:
**Description**

 - created as a POC during a discussion. should be removed after the KEB allows changing global account

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/model.go
+++ b/components/kyma-environment-broker/internal/model.go
@@ -112,14 +112,15 @@ type EventHub struct {
 }
 
 type Instance struct {
-	InstanceID      string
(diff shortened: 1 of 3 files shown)


components/kyma-environment-broker/internal/model.go
```diff
@@ -262,6 +271,8 @@ type InstanceDetails struct {
 	Monitoring        MonitoringData            `json:"monitoring"`
 	EDPCreated        bool                      `json:"edp_created"`
 
+	SubscriptionGlobalAccountID string `json:xxx`
```
piotrmiskiewicz, line 274: not needed

PR #1166, opened 2021-12-02
Add persisting cluster config in fake reconciler

ksputo:
**Description**

- Added a persistence of `clusterConfig` containing for example components list for future assertions, when `ApplyClusterConfig` method is called on fake reconciler more than once

**Related issue(s)**


wozniakjan (approved):
/lgtm

PR #1168, opened 2021-12-03
Align GCP plan values with the published schema

ebensom:
**Description**

- default autoscaler min = 2
- default autoscaler max = 10
- do not drain any modes prematurely
- Use 50 GB volume

**Related issue(s)**


voigt (approved):
/lgtm

PR #1190, opened 2021-12-09
Expose kyma version on runtimes endpoint

ebensom:
**Description**

- Expose kyma version on KEB `/runtimes` endpoint from the operation `RuntimeVersion` data. Only applicable when all operations are queried.
- Adjust KCP CLI to optionally display kyva version in default table format.

**Related issue(s)**


components/kyma-environment-broker/internal/runtime/handler.go
```diff
@@ -172,6 +175,9 @@ func (h *Handler) setRuntimeAllOperations(instance internal.Instance, dto *pkg.R
 		if len(provOprs) > 1 {
 			h.converter.ApplyUnsuspensionOperations(dto, provOprs[:len(provOprs)-1])
 		}
+		// Set kyma version from the last provisioning operation
+		kymaVersion = lastProvOp.RuntimeVersion.Version
```
wozniakjan, line 179: don't upgrade operations also kyma runtime version? Would it make sense to take the last provisioning or upgrade operation to determine the current kyma version of the SKR?

wozniakjan (approved):
/lgtm
/hold

feel free to unhold if you think my comment doesn't apply

components/kyma-environment-broker/internal/runtime/handler.go
ebensom: Yes, it's exactly what I'm doing a few lines below this change. If there are upgrade kyma operations and the last upgrade operation happened after the last provision/unsuspension operation, take it's kyma version data instead.

ebensom: Or did I miss something?

wozniakjan: oh, my bad, I missed that. Ignore me :)

Issue #1163, opened 2021-12-02
KEB orchestration with upgrade queue should be aware of kyma version and kyma migration status


wozniakjan:
Calling the orchestration endpoint currently always includes service catalog and other deprecated components. KEB orchestration should lookup whether an instance is supposed to have SVCAT or BTP Operator.

this is a followup to https://github.com/kyma-project/control-plane/pull/825 and https://github.com/kyma-project/control-plane/pull/980

PR #1184, opened 2021-12-09
Encrypt BTP Operator ERS credentials

wozniakjan:
**Description**

https://github.com/kyma-project/control-plane/pull/980 has introduced `SMOperatorCredentials` in the `ERSContext`. The encryption implemented for each operation and instance would drop the `SMOperatorCredentials` and not store them in database.

This enhances the work by storing encrypted credentials in `SMOperatorCredentials` struct for each operation and instance.

**Related issue(s)**

https://github.com/kyma-project/control-plane/issues/1163

PR #1201, opened 2021-12-13
Make KEB upgrade kyma queue aware of BTP Operator credentials

wozniakjan:
**Description**
Upgrade kyma queue currently ignores the fact ERS call could have updated ERSContext with new BTP operator credentials and would always install service catalog. This ensures KEB picks up the up to date ERSContext.

**Related issue(s)**

closes: https://github.com/kyma-project/control-plane/issues/1163
depends on:
- [x] https://github.com/kyma-project/control-plane/pull/1184

### DIFF
--- a/components/kyma-environment-broker/cmd/broker/broker_suite_test.go
+++ b/components/kyma-environment-broker/cmd/broker/broker_suite_test.go
@@ -508,7 +508,7 @@ func (s *BrokerSuiteTest) FinishUpdatingOperationByReconcilerBoth(operationID st
 	})
 	assert.NoError(s.t, err)
 
-	var state *reconciler.State
+	var state *reconcilerApi.HTTPClusterResponse
 	for ccv := updatingOp.ClusterConfigurationVersion; ccv <= updatingOp.ClusterConfigurationVersion+1; ccv++ {
 		err = wait.Poll(pollingInterval, 2*time.Second, func() (bool, error) {
 			state, err = s.reconcilerClient.GetCluster(updatingOp.RuntimeID, ccv)
@@ -516,7 +516,7 @@ func (s *BrokerSuiteTest) FinishUpdatingOperationByReconcilerBoth(operationID st
 				return false, err
 			}
 			if state.Cluster != "" {
-				s.reconcilerClient.ChangeClusterState(updatingOp.RuntimeID, ccv, reconciler.ReadyStatus)
+				s.reconcilerClient.ChangeClusterState(updatingOp.RuntimeID, ccv, reconcilerApi.StatusReady)
 				return true, nil
 			}
 			return false, nil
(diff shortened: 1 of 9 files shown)


components/kyma-environment-broker/cmd/broker/broker_suite_test.go
```diff
@@ -483,6 +483,38 @@ func (s *BrokerSuiteTest) FinishUpdatingOperationByReconciler(operationID string
 	assert.NoError(s.t, err)
 }
 
+func (s *BrokerSuiteTest) FinishUpdatingOperationByReconcilerBoth(operationID string) {
```
wozniakjan, line 486: this is a shorthand helper func, because update queue calls reconciler twice.

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -922,6 +917,11 @@ func NewKymaOrchestrationProcessingQueue(ctx context.Context, db storage.BrokerS
 			step:   upgrade_kyma.NewApplyClusterConfigurationStep(db.Operations(), db.RuntimeStates(), reconcilerClient),
 			cnd:    upgrade_kyma.ForKyma2,
 		},
+		{
+			weight: 11,
+			step:   upgrade_kyma.NewCheckClusterConfigurationStep(db.Operations(), reconcilerClient, 15*time.Minute),
+			cnd:    upgrade_kyma.ForKyma2,
```
wozniakjan, line 923: for some reason, this step was before calling reconciler API. I think the check should be right after the reconciler call.

components/kyma-environment-broker/internal/process/update/initialisation_step.go
```diff
@@ -64,11 +64,20 @@ func (s *InitialisationStep) Run(operation internal.UpdatingOperation, log logru
 		if err != nil {
 			return operation, time.Second, nil
 		}
+		if operation.ProvisioningParameters.ErsContext.SMOperatorCredentials != nil {
+			instance.Parameters.ErsContext.SMOperatorCredentials = operation.ProvisioningParameters.ErsContext.SMOperatorCredentials
```
wozniakjan, line 68: both update and upgrade queues need to take the SMOperatorCreds from previous operations if the current operation doesn't have them.

components/kyma-environment-broker/internal/process/upgrade_kyma/get_kubeconfig_step.go
```diff
@@ -33,6 +33,7 @@ func (s *GetKubeconfigStep) Name() string {
 
 func (s *GetKubeconfigStep) Run(operation internal.UpgradeKymaOperation, log logrus.FieldLogger) (internal.UpgradeKymaOperation, time.Duration, error) {
 	if operation.Kubeconfig != "" {
+		operation.InputCreator.SetKubeconfig(operation.Kubeconfig)
```
wozniakjan, line 36: input creator is not stored in the database in case of requeue, need to take it from the operation's kubeconfig that is stored

components/kyma-environment-broker/cmd/broker/main.go
voigt: It sounds correct what you are saying.

However, I remember looking over this part together with @ksputo - wasn't there a specific reason for checking the cluster config earlier?

piotrmiskiewicz (dismissed):

piotrmiskiewicz (approved):

components/kyma-environment-broker/cmd/broker/main.go
wozniakjan: I guess we will find out soon, however, this was preventing running upgrade on kyma2 whether migrated or not. So we wouldn't be able to upgrade even with service catalog from 2.0.0 to 2.0.1.

wozniakjan: even after followup inspection, I couldn't figure out a reason why check should be the first step in the queue, especially if it is the only step that marks the operation as "finished".

Thank you for the heads up @voigt, I will make sure to discuss with @ksputo and also keep an eye for potential bug reports in this area.

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -878,11 +878,6 @@ func NewKymaOrchestrationProcessingQueue(ctx context.Context, db storage.BrokerS
 		step     upgrade_kyma.Step
 		cnd      upgrade_kyma.StepCondition
 	}{
-		{
-			weight: 2,
```
wozniakjan, line 882: note: this was not supposed to be moved. Explanation in https://github.com/kyma-project/control-plane/pull/1258.

PR #1232, opened 2021-12-23
Do not use dirty Secret Bindings

franpog859:
**Description**

Currently, HAP could provide a "dirty" Secret Binding for the new SKR when a tenant deprovisioned a runtime (Secret Binding is marked as "dirty") and provisioned a runtime before the clean up was performed by Subscription Cleanup Job ([HAP uses the existing, "dirty" Secret Binding](https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/common/hyperscaler/account_pool.go#L112)). In such a case, all the resources would be removed by Subscription Cleanup Job as it cleans up everything from all the Secret Bindings marked "dirty"

- HAP will provide only clean Secret Bindings. All the dirty ones will be cleaned up regularly as before
- Subscription Cleanup Job will run twice per day

There are possible drawbacks to such a solution. Subscription Cleanup Job cleans everything up once per day now. If a customer provisions and deprovisions their SKR multiple times, all of existing Secret Bindings could be used and marked as "dirty" so there could be a situation when there is no clean one left. In this case, KEB will return an error saying `failed to find unassigned secret binding` and there will be no option to provision a new SKR till the next cleanup. To minimize such possibility I configured the Subscription Cleanup Job to run twice per day instead of only once


koala7659 (approved):
LGTM

components/kyma-environment-broker/common/hyperscaler/account_pool_test.go
```diff
@@ -277,8 +280,36 @@ func newTestAccountPool() AccountPool {
...
+			Name:      "secret8",
+			Namespace: testNamespace,
+		},
+	}
+	secretBinding9 := &gardener_types.SecretBinding{
+		ObjectMeta: machineryv1.ObjectMeta{
+			Name:      "secretBinding9",
+			Namespace: testNamespace,
+			Labels: map[string]string{
+				"hyperscalerType": "azure",
```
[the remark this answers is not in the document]
franpog859: In the current implementation, a Secret Binding can have a `dirty` label set to `true` or just no `dirty` label. Subscription Cleanup Job [deletes the label](https://github.com/kyma-project/control-plane/blob/main/components/subscription-cleanup-job/internal/job/cleaner.go#L105) when it's not dirty anymore. In this case there is no option to have `dirty` label set to `false`

vvxxvvxx:
lgtm

resources/kcp/charts/kyma-environment-broker/templates/subscription-cleanup-job.yaml
```diff
@@ -30,5 +30,5 @@ spec:
             - name: gardener-kubeconfig
               secret:
                 secretName: {{ .Values.gardener.secretName }}
-  schedule: '0 2 * * *'
+  schedule: '0 2,14 * * *'
```
koala7659, line 33: It implies to run subscription-cleanup-job every 12 hours which is IMHO too rare. You should restore previous schedule for this job after problem is fixed.

koala7659 (approved):

components/kyma-environment-broker/common/hyperscaler/account_pool.go
```diff
@@ -109,7 +109,7 @@ func (p *secretBindingsAccountPool) IsSecretBindingUsed(hyperscalerType Type, te
 }
 
 func (p *secretBindingsAccountPool) CredentialsSecretBinding(hyperscalerType Type, tenantName string) (*v1beta1.SecretBinding, error) {
-	labelSelector := fmt.Sprintf("tenantName=%s,hyperscalerType=%s", tenantName, hyperscalerType)
+	labelSelector := fmt.Sprintf("tenantName=%s, hyperscalerType=%s, !dirty", tenantName, hyperscalerType)
```
koala7659, line 112: After additional review I think we should make in both cases (L112 and L 126) same label selector. `shared!=true, tenantName=%s, hyperscalerType=%s, !dirty` . I'm not sure if we can add `internal=true` label to selector query.

koala7659 (changes requested):

resources/kcp/charts/kyma-environment-broker/templates/subscription-cleanup-job.yaml
piotrmiskiewicz: just a question: I'm not sure how statistics looks like, how is it possible to have all subscriptions dirty, but why not run such job every hour?

piotrmiskiewicz (approved):

components/kyma-environment-broker/common/hyperscaler/account_pool.go
koala7659: Resolved!

koala7659 (approved):
LGTM

piotrmiskiewicz (approved):

PR #1249, opened 2022-01-04
Cluster deregistration must clean ClusterConfigurationVersion if done.

piotrmiskiewicz:
**Description**

- Check_Cluster_deregistration step is finished in few cases: cluster is "deleted" or cluster does not exists. In all cases it must clean the ClusterConfigurationVersion so the unsuspension process will start properly

**Related issue(s)**


wozniakjan (approved):
/lgtm

PR #1251, opened 2022-01-04
Add the `state` command

pPrecel:
**Description**

- add `state` command to the `reconciliations` parent.
- command allows users to display the Cluster State JSON

**Related issue(s)**

https://github.com/kyma-incubator/reconciler/issues/268


ebensom:
This command is more like for the reconciliation state of the runtime. The runtime state is completely different, and is controlled by and derived from KEB. 
Please move this command to the `reconciliations` command family:  `reconciliation state ...`

pPrecel:
> This command is more like for the reconciliation state of the runtime. The runtime state is completely different, and is controlled by and derived from KEB. There are already existing API in KEB to obtain the runtime state (and also corresponding kyma and cluster config) which corresponds to the runtime's lifecycle state machine.
> 
> _Please move this command to the `reconciliations` command family: `reconciliation state ...`_

Many thanks for the suggestion! This `State` represents the state of the cluster but it can be filtered by scheduling ID/runtime ID/correlation ID. I agree that `runtime` is not the best parent command but exactly as the `reconciliation` to be honest. What do you think about adding new `cluster` command? Then we could use it as `kcp cluster state` or `kcp cluster` to take the state of the cluster

pPrecel:
After quick sync with @ebensom we decided to move the `state` command under the `reconciliation` parent

jeremyharisch (approved):

m00g3n (approved):
LGTM

Issue #1254, opened 2022-01-06
KCP-CLI add flag to display Subscription Global Account ID


voigt:
Assignee: @voigt

voigt:
Added flag `--subscription-global-account-id`:

```bash
$ ./kcp-darwin runtimes --subscription-global-account-id
GLOBALACCOUNT ID                        Subscription Global Account ID   SUBACCOUNT ID                               SHOOT       REGION         PLAN         CREATED AT            STATE
3e64ebae-38b5-46a0-b1ed-9ccee153a0ae    empty-gaID                                                                   c-48452c7                               2020/03/30 08:05:11   failed (deprovision)
3e64ebae-38b5-46a0-b1ed-9ccee153a0ae                                                                                                                         2020/04/01 14:01:35   failed (deprovision)
3e64ebae-38b5-46a0-b1ed-9ccee153a0ae                                                                                                                         2020/04/15 16:36:06   failed (deprovision)
```

PR #1255, opened 2022-01-06
add flag to display subscription-global-account-id to kcp-cli

voigt:
**Description**

- adds flag `--subscription-global-account-id` to kcp-cli to display additional field

```bash
$ ./kcp-darwin runtimes --subscription-global-account-id
GLOBALACCOUNT ID                        Subscription Global Account ID   SUBACCOUNT ID                               SHOOT       REGION         PLAN         CREATED AT            STATE
3e64ebae-38b5-46a0-b1ed-9ccee153a0ae    empty-gaID                                                                   c-48452c7                               2020/03/30 08:05:11   failed (deprovision)
3e64ebae-38b5-46a0-b1ed-9ccee153a0ae                                                                                                                         2020/04/01 14:01:35   failed (deprovision)
3e64ebae-38b5-46a0-b1ed-9ccee153a0ae                                                                                                                         2020/04/15 16:36:06   failed (deprovision)
```

**Related issue(s)**

- #1254

### DIFF
--- a/tools/cli/pkg/command/runtime.go
+++ b/tools/cli/pkg/command/runtime.go
@@ -144,7 +144,6 @@ func (cmd *RuntimeCommand) Validate() error {
 func (cmd *RuntimeCommand) printRuntimes(runtimes runtime.RuntimesPage) error {
 	switch {
 	case cmd.output == tableOutput:
-
 		if cmd.display.SubscriptionGlobalAccountID {
 			tableColumns = append(tableColumns[:1+1], tableColumns[1:]...)
 			tableColumns[1] = printer.Column{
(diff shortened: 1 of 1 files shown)


ebensom:
I'm wondering why all this fuzz, i.e. adding a new flag for a specific attribute of the runtime? 
I don't see it a problem if the default table output does not show the subscription GA.
There is a custom output similar to `kubectl custom-columns` where specific non-default fields of the runtime objects could be displayed. Furthermore there is the full JSON output.

Take `kubectl` as an example. Every resource (called runtime.Object in K8S package) has a default metadata about which attributes to display (and how) in the default table output. Except for `--show-labels`, You can only control which attributes to display (and how) with changing the output type (-o wide / custom-columns=... / yaml / json / ...)

piotrmiskiewicz (approved):

PR #1258, opened 2022-01-11
Move check cluster configuration to the begining

piotrmiskiewicz:
**Description**

- because we do not have srtages in the upgrade kyma process, the check step must be the first one - to prevend unecessary running other steps after the reconciler is called

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/cmd/broker/main.go
+++ b/components/kyma-environment-broker/cmd/broker/main.go
@@ -878,6 +878,14 @@ func NewKymaOrchestrationProcessingQueue(ctx context.Context, db storage.BrokerS
 		step     upgrade_kyma.Step
 		cnd      upgrade_kyma.StepCondition
 	}{
+		// check cluster configuration is the first step - to not exeute other steps, when cluster configuration was applied
+		// this should be moved to the end when we introduce stages like in the provisioning process
+		// (also return operation, 0, nil at the end of apply_cluster_configuration)
+		{
+			weight: 1,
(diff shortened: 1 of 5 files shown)


components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -878,6 +878,14 @@ func NewKymaOrchestrationProcessingQueue(ctx context.Context, db storage.BrokerS
 		step     upgrade_kyma.Step
 		cnd      upgrade_kyma.StepCondition
 	}{
+		// check cluster configuration is the first step - to not execute other steps, when cluster configuration was applied
+		// this should be moved to the end when we introduce stages like in the provisioning process
+		// (also return operation, 0, nil at the end of apply_cluster_configuration)
+		{
+			weight: 1,
```
wozniakjan, line 885: fwiw, initially it had weight 2 - https://github.com/kyma-project/control-plane/pull/1201/files#r782155263

not sure if it's relevant or if 1 is ok as well. Feel free to unhold if it doesn't matter.

wozniakjan (approved):
For completion, the bug was introduced in https://github.com/kyma-project/control-plane/pull/1201
Thanks for fixing this!

/lgtm
/hold

szwedm (approved):

components/kyma-environment-broker/cmd/broker/main.go
piotrmiskiewicz: it is not a problem, init step has weight=0

PR #1262, opened 2022-01-13
Add check if cluster still exists before returning secret to the pool

Maladie:
**Description**

- Add check if cluster still exists before returning secret to the pool
- Test added

**Related issue(s)**


franpog859 (approved):
LGTM 🎉

PR #1269, opened 2022-01-14
KEB: Use configurable reconciler timeout in upgrade Check_Cluster_Configuration step

ebensom:
**Description**

- Use the same (configurable) reconciler `ProvisioningTimeout`  in Kyma 2 upgrade process when polling mothership-reconciler, which is used during the provisioning process.


wozniakjan (approved):
/lgtm

PR #1281, opened 2022-01-20
Reconciler Timeout Note

ralikio:
**Description**

- Description of Reconciler timeout when invoked from KEB.

**Related issue(s)**

### DIFF
--- a/docs/kyma-environment-broker/03-03-runtime-operations.md
+++ b/docs/kyma-environment-broker/03-03-runtime-operations.md
@@ -34,6 +34,9 @@ The provisioning process contains the following steps:
 
 >**NOTE:** The timeout for processing this operation is set to `24h`.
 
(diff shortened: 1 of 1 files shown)


docs/kyma-environment-broker/03-03-runtime-operations.md
```diff
@@ -34,6 +34,9 @@ The provisioning process contains the following steps:
 
 >**NOTE:** The timeout for processing this operation is set to `24h`.
```
klaudiagrz, line 35:
```suggestion
The timeout for processing the whole provisioning operation is set to `24h`. The provisioning steps delegate resource creation to Reconciler. Since Reconciler does not constrain a number of retries in case of failed reconciliation, KEB introduces a [provisioning timeout for Reconciler](../../resources/kcp/charts/kyma-environment-broker/values.yaml#L49) that is set to `2h`.
```

klaudiagrz (approved):

PR #1285, opened 2022-01-20
interval updated

iamps5:
**Description**

- avs interval updated from 180s to 60s

**Related issue(s)**


ebensom (commented):
/lgtm

You will also need to bump the KEB image in the kcp chart here:
https://github.com/kyma-project/control-plane/blob/main/resources/kcp/values.yaml#L17

ebensom (approved):

Issue #1286, opened 2022-01-21
Upgrade incorrectly installed BTP operator in the SKR


wozniakjan:
**Description**

A user has run into a scenario where an upgraded kyma cluster installed btp operator without credentials despite the fact it wasn't supposed to install one.

https://sap-btp.slack.com/archives/C0140PCSJ5Q/p1642686511076300

**Expected result**
Upgrade SKR installing only the components that are matching user expectations, do not install btp operator when it's not supposed to be installed

**Actual result**
btp operator was installed without explicit request for it

**Steps to reproduce**
n/a

**Troubleshooting**
n/a

PR #1287, opened 2022-01-21
Explicitly disable BTP Operator for upgrade queue

wozniakjan:
**Description**

For update and provisioning, we explicitly disable BTP operator and only enable it when appropriate creds are present. For upgrade, we relied on different logic and that is inheriting `InputCreator` from previous operations which may not be as robust as we thought. This is an effort to unify components enablement/disablement related to SVCAT migration.

**Related issue(s)**

mitigation of https://github.com/kyma-project/control-plane/issues/1286


ksputo (dismissed):

ksputo (dismissed):

ksputo (approved):
LGTM

PR #1289, opened 2022-01-21
Fix of the db persistence layer for saving cluster configuration during shoot upragde

koala7659:
Adding of storing two missing fields :

- machine_image
- machine_image_version 

Removing redundant fields: 

- region
- provider
- worker_cidr

Additional fix to preserve existing `ExposureClassName` in input converter in the case when new value is empty (nil)


franpog859 (approved):
LGTM

PR #1302, opened 2022-01-26
modify nodeSelector in component-reconcilers

lauraanddolay:
adjust `nodeSelector` for `specific` component (e.g. rma, ory ...) instead of the `generic` component


lauraanddolay:
cancel this PR.

PR #1308, opened 2022-01-27
Fix resetting autoscaler params in cluster upgrade operation

ebensom:
**Description**

Service Instance update API call can change a runtime's autoscaler min/max parameters. However during orchestrated cluster upgrade, the provisioner shoot upgrade input is built using the original provisioning parameters used during runtime creation, which case updated parameters to be reset back to their original value.

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/process/upgrade_cluster/initialisation.go
+++ b/components/kyma-environment-broker/internal/process/upgrade_cluster/initialisation.go
@@ -115,20 +115,6 @@ func (s *InitialisationStep) Run(operation internal.UpgradeClusterOperation, log
 }
 
 func (s *InitialisationStep) initializeUpgradeShootRequest(operation internal.UpgradeClusterOperation, log logrus.FieldLogger) (internal.UpgradeClusterOperation, time.Duration, error) {
-	// rewrite necessary data from ProvisioningOperation to operation internal.UpgradeOperation
-	provisioningOperation, err := s.operationStorage.GetProvisioningOperationByInstanceID(operation.InstanceID)
-	if err != nil {
-		log.Errorf("while getting provisioning operation from storage")
-		return operation, s.timeSchedule.Retry, nil
-	}
-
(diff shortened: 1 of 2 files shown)


components/kyma-environment-broker/internal/process/upgrade_cluster/initialisation.go
```diff
@@ -115,20 +115,6 @@ func (s *InitialisationStep) Run(operation internal.UpgradeClusterOperation, log
...
-		log.Errorf("while getting provisioning operation from storage")
-		return operation, s.timeSchedule.Retry, nil
-	}
-
-	operation, delay := s.operationManager.UpdateOperation(operation, func(op *internal.UpgradeClusterOperation) {
-		op.ProvisioningParameters = provisioningOperation.ProvisioningParameters
-	}, log)
-	if delay != 0 {
-		return operation, delay, nil
-	}
```
wozniakjan, line 130: I am not familiar with the logic to a sufficient extent, but I was wondering if we should get `ProvisioningParameters` from the last successful operation? In KEB, certain params are expected to be populated but sometimes these are added in the broker API endpoint, sometimes in the initialization step so not sure which workflow would fit best here.

Do we know if the rest of the upgrade queue expects these to be present?

wozniakjan: maybe @piotrmiskiewicz remembers?

ebensom: It is being done actually, when the upgrade cluster operation is created, the ProvisioningParameters is copies over from the Instance, which holds the latest state of the params:
https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/internal/orchestration/manager/upgrade_cluster.go#L60

wozniakjan (approved):
/lgtm

PR #1312, opened 2022-01-27
Improvement of checking Kyma version before starting cluster deprovision 

koala7659:
Previous implementation to recognise if the version of the cluster being deprovisioned was Kyma 2.x was based on checking  missing `activeKymeConfigId` entry in the database. This entry was present only for previously provisioned `Kyma 1.x `clusters.

It turned out there are some cases when `Kyma 2.x` cluster might still have assigned this data. It is in the case of Kyma 1.x clusters upgraded to version `Kyma 2.x` because the upgrade was performed by the Reconciler and Provisioner was not aware about it. As a result Kyma version installed on the cluster was not identified correctly and deprovisioning ended with a failure.  

The current mechanism is using `installationSDK` library to identify if deprovisioned cluster was installed as `Kyma 1.x` (with `kyma-installer` and `installation CR`). In the  other case we assume that it is `Kyma 2.x `cluster and we don't init uninstallation step - just delete the Kubernetes cluster and cleanup the environment.

PR #1334, opened 2022-02-04
Allow to deprovision Kyma runtime without kubeconfig in the database

koala7659:
Fix of the regression from #1312 with interrupted deprovisioning of SKR that does not have valid kubeconfig in the database. New helper function `shouldDeprovisionWithoutUninstall` added with separate logic to determine version of installed Kyma cluster. 

Logically we can have 2 cases : `kubeconfig` present and missing and each one must be evaluated for cases `Kyma 1.x` and `Kyma 2.x`: 

1. `Kubeconfig` is missing because of failed installation
   - Kyma 1.x - `ActiveKymaConfigId` exists
         Deprovisioning with Kyma uninstall 
   - Kyma  2.x - `ActiveKymaConfigId` is missing
         Deprovisioning with cluster delete

2. `Kubeconfig` exists. Kyma installed with no errors.
    - Kyma 1.x  - `ActiveKymaConfigId` exists
       Deprovisioning with Kyma uninstall 
     - Kyma 2.x `ActiveKymaConfigId` is missing or `ActiveKymaConfigId` exists  but missing installation CR (`Kyma1` upgraded to `Kyma2`)
       Deprovisioning with cluster delete


components/provisioner/internal/provisioning/service.go
```diff
@@ -475,6 +459,28 @@ func (r *service) RollBackLastUpgrade(runtimeID string) (*gqlschema.RuntimeStatu
...
 
+func (r *service) isKyma2Installed(cluster model.Cluster) bool {
+
+	if util.IsNilOrEmpty(cluster.ActiveKymaConfigId) {
+		return true
+	}
+
+	if cluster.Kubeconfig == nil {
+		log.Warnf("Kubeconfig for cluster %s is missing", cluster.ID)
+		return false
```
franpog859, line 470: When there is no kubeconfig, should we proceed with the "uninstall" flow? There is nothing to uninstall :D

components/provisioner/internal/provisioning/service.go
```diff
@@ -475,6 +459,28 @@ func (r *service) RollBackLastUpgrade(runtimeID string) (*gqlschema.RuntimeStatu
 	return r.RuntimeStatus(runtimeID)
 }
 
+func (r *service) isKyma2Installed(cluster model.Cluster) bool {
```
franpog859, line 462: I'm not sure about the name of this function. There is for example a condition with the existence of the kubeconfig which is not strictly related to the Kyma1.X/ Kyma2.X thing. Maybe we could go for something like `shouldDeprovisionWithoutUninstall`

franpog859 (commented):
I'd add a test where kubeconfig is not missing but there was an error with the installation so it's in some error state. In this case, it should proceed with the "with installation" operation

components/provisioner/internal/provisioning/service.go
koala7659: This will be processed correctly in trigger_kyma_uninstall.go step. Uninstall operation will be skipped if there is no kubeconfig. My intention was to restore original implementation.

franpog859 (approved):
lgtm

PR #1335, opened 2022-02-08
Keb add kyma version

piotrmiskiewicz:
**Description**

The runtime_state entity contains a field in the DB "kyma_version" which was updated by manual SQL script. The logic in the code executes a query, which filters by that field. But after that, it reads the kyma version from other field. In that case, the runtimestate contains the kyma_version (so the row was returned), but it was only gardener cluster update  - no kyma version in the data.


szwedm (approved):
/lgtm

szwedm (approved):
/lgtm

PR #1343, opened 2022-02-09
Bump schema-migrator base image to 3.15.0

wozniakjan:
**Description**

OpenSSL from edge has version [1.1.1m-r1](https://pkgs.alpinelinux.org/package/edge/main/x86/openssl#), this is to remediate CVE-2021-4160 present in current `1.1.1l-r7`.

**Related issue(s)**

### DIFF
--- a/components/schema-migrator/Dockerfile
+++ b/components/schema-migrator/Dockerfile
@@ -1,11 +1,11 @@
-FROM alpine:3.14.3
+FROM alpine:3.15.0
 LABEL source=git@github.com:kyma-project/control-plane.git
 
 ARG MIGRATE_VER=4.15.1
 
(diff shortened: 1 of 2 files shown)


ralikio (approved):

wozniakjan:
> Do we need to add openssl to the image?

we are not adding it, it's transitive dep of curl and postgress client. We are just using edge from alpine to install explicitly the newest available version [1.1.1m-r1](https://pkgs.alpinelinux.org/package/edge/main/x86/openssl#) because the older ones show in our security scanners as vulnerable to https://github.com/advisories/GHSA-ph2x-8239-7xc7

ksputo (approved):

PR #1351, opened 2022-02-11
Fix shoot filter response

kwiatekus:
**Description**

- IF shoot filter is provided it has to correspond to a valid runtime, otherwise no call to mothership is made

**Related issue(s)**
https://github.com/kyma-incubator/reconciler/issues/693


m00g3n (approved):
LGTM

PR #1352, opened 2022-02-11
Deprovisioning operations should be created as Pending

piotrmiskiewicz:
**Description**

- Deprovisioning operation must be created as pending

**Related issue(s)**


wozniakjan (approved):
/lgtm
although it looks like many unit tests expect this to be the case so you may need to amend a patches for the tests

wozniakjan (approved):

PR #1367, opened 2022-02-17
Ers cli commands

piotrmiskiewicz:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/tools/cli/pkg/ers/command/instances.go
+++ b/tools/cli/pkg/ers/command/instances.go
@@ -1,8 +1,6 @@
 package command
 
 import (
-	"encoding/json"
-	"fmt"
 	"github.com/kyma-project/control-plane/tools/cli/pkg/ers"
 	"github.com/kyma-project/control-plane/tools/cli/pkg/ers/client"
 	"github.com/kyma-project/control-plane/tools/cli/pkg/printer"
@@ -17,6 +15,10 @@ var tableColumns = []printer.Column{
 	{
 		Header:    "GLOBALACCOUNT ID",
 		FieldSpec: "{.GlobalAccountId}",
+	},
+	{
+		Header:    "BROKER ID",
+		FieldSpec: "{.BrokerId}",
 	},
 	{
 		Header:    "MIGRATED",
@@ -46,7 +48,6 @@ type InstancesCommand struct {
 }
 
 func (c *InstancesCommand) Run() error {
-	// TODO: this is a dummy implementation
 	if c.source != "" {
 		c.instanceFetcher = client.NewFileClient(c.source)
 	} else {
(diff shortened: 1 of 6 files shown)


tools/cli/pkg/ers/command/instances.go
```diff
@@ -1,43 +1,141 @@
...
+		// todo: use real client to ers
+		c.instanceFetcher = &emptyFetcher{}
+	}
+
+	tp, _ := printer.NewTablePrinter(tableColumns, false)
+
+	if c.filters.InstanceID != "" {
+		instance, err := c.instanceFetcher.GetInstanceById(c.filters.InstanceID)
+		tp.PrintObj(instance)
+		return err
```
tillknuesting, line 63: Would be nice to wrap this returned error with some more information e.g.` return fmt.Errorf("GetInstanceById: %w", err)`

tools/cli/pkg/ers/command/instances.go
```diff
@@ -1,43 +1,141 @@
...
+			continue
+		}
+		if c.filters.GlobalAccountID != "" && item.GlobalAccountId != c.filters.GlobalAccountID {
+			continue
+		}
+		result = append(result, item)
+	}
+
+	tp.PrintObj(result)
+	return err
```
tillknuesting, line 81: Returned error is supposed to be `nil`?

tools/cli/pkg/ers/instance.go
```diff
@@ -0,0 +1,13 @@
+package ers
+
+// Instance defines a struct for data stored by the ERS
+type Instance struct {
+	Id              string `json:"id"`
```
tillknuesting, line 5: I would propose to use ID instead of Id. See https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tools/cli/pkg/ers/command/instances.go
```diff
@@ -1,43 +1,141 @@
...
+	}
+
+	tp, _ := printer.NewTablePrinter(tableColumns, false)
+
+	if c.filters.InstanceID != "" {
+		instance, err := c.instanceFetcher.GetInstanceById(c.filters.InstanceID)
+		tp.PrintObj(instance)
+		return err
+	}
+	var result []ers.Instance
```
tillknuesting, line 65: Consider to preallocate slice for performance benefits e.g. `result := make([]ers.Instance, 0, len(instances))`

tools/cli/pkg/ers/command/instances.go
ralikio: To be fixed in the following PRs.

tools/cli/pkg/ers/command/instances.go
ralikio: To be fixed in the following PRs.

tools/cli/pkg/ers/instance.go
ralikio: To be fixed in the following PRs

tools/cli/pkg/ers/command/instances.go
ralikio: To be fixed in the following PRs.

ralikio (approved):

PR #1369, opened 2022-02-17
Errors categorization - meaningful alerting for provisioning functionality

vvxxvvxx:
**Description**

- Make the original caused error returned from each Run step of Provisioning process
- Categorize the errors returned
- Expose the details of the error returned including error component(category), error reason, error detailed/original message from the codes to the provisioning metric `compass_keb_provisioning_result` so as to help quickly locate the error/problem and help statistic of problem.

The **tests** performed are summarized [here](https://github.tools.sap/kyma/backlog/issues/1933#issuecomment-755314)

**Related issue(s)**
See backlog #1933


components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
```diff
@@ -88,7 +89,7 @@ func (s *InitialisationStep) Run(operation internal.DeprovisioningOperation, log
 func (s *InitialisationStep) run(operation internal.DeprovisioningOperation, log logrus.FieldLogger) (internal.DeprovisioningOperation, time.Duration, error) {
 	if time.Since(operation.CreatedAt) > s.operationTimeout {
 		log.Infof("operation has reached the time limit: operation was created at: %s", operation.CreatedAt)
-		return s.operationManager.OperationFailed(operation, fmt.Sprintf("operation has reached the time limit: %s", s.operationTimeout), log)
+		return s.operationManager.OperationFailed(operation, fmt.Sprintf("operation has reached the time limit: %s", s.operationTimeout), errors.New(""), log)
```
ebensom, line 92: Shouldn't there be a message/reason in the various places where the error created as `errors.New("")`?
I know it's wrapped with the original error description, but wouldn't it cause awkward displaying of the error?

vvxxvvxx: yea.. will be an extra colon in the text end... I just changed this argument to nil when there is no exact err passed in. if there is a specific err to pass in, it will be wrapped with description to keep the original err object for categorizing..

components/kyma-environment-broker/internal/process/upgrade_kyma_operation_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pkg/errors"
```
tillknuesting, line 8: why use an archived package? Can you check if you can use "errors" from standard library as it offers similar functionality.

components/kyma-environment-broker/go.mod
```diff
@@ -149,6 +154,7 @@ require (
 )
 
 replace (
+	github.com/kyma-project/control-plane/components/provisioner => github.com/vvxxvvxx/control-plane/components/provisioner v0.0.0-20220302055653-40b7f2b30892
```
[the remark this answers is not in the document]
vvxxvvxx: Hi, this is only to make the KEB image pointing to the right provisioner which has my changes. The provisioner code changes has been split to another PR https://github.com/kyma-project/control-plane/pull/1405. After the PR#1405 for provisioner is merged, I will remove this replace for provisioner in keb go.mod..

components/provisioner/internal/apperrors/apperrors.go
```diff
@@ -1,10 +1,45 @@
...
+	ErrTriggerKymaInstallation    ErrReason = "err_trigger_kyma_installation"
+)
+
+type ErrCode int
+type CauseCode int
 
 const (
 	CodeBadGateway ErrCode = 502
 	CodeInternal   ErrCode = 500
+	CodeExternal   ErrCode = 501
```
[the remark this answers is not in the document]
vvxxvvxx: Hi, this External code is for the error from error component/category ErrMpsOAuth2("mps oauth2"), ErrKymaInstaller("kyma installer" hydroform) or ErrGardener("gardener").. It will be filtered out by KEB as https://github.com/kyma-project/control-plane/blob/66880df6700eb3fa10e67274b691b389af929d79/components/kyma-environment-broker/internal/provisioner/client.go#L189.

Do you think I could change it to InternalCode (I just wanted to differentiate this kind of error as the error from something external)?

wozniakjan:
After looking at the changes a little closer, I realized this might be actually doing something not entirely in alignment with prometheus best practices. Usage of high cardinality or even worse unbounded sets as labels can be a cause of instability and exponential cost growth on prometheus, it's generally not advised - https://prometheus.io/docs/practices/naming/#labels

Number of time-series in prometheus db is a Cartesian product of all labels. The fact we have `InstanceID` as a label can already be problematic when the adoption of kyma grows, using something as unpredictable as error message in a label can start with poor performance but quickly escalate to unmaintainable prometheus db. It will grow exponentially.

Are we confident that we can have an upper bound on the cardinality of `errors` and `reasons`? Looking at some examples here, they seem very verbose such as this one and due to stacktrace presence, growth might be unpredictable

> [task "Waiting until shoot infrastructure has been reconciled" failed: Error while waiting for Infrastructure shoot--kyma-dev--b2608b8/b2608b8 to become ready: error during reconciliation: Error reconciling infrastructure: failed to apply the terraform config: Terraform execution for command 'apply' could not be completed: \* Error creating EIP: AddressLimitExceeded: The maximum number of addresses has been reached. status code: 400, request id: <omitted> on tf/main.tf line 221, in resource "aws_eip" "eip_natgw_z0": 221: resource "aws_eip" "eip_natgw_z0" { \* Error creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached. status code: 400, request id: <omitted> on tf/main.tf line 21, in resource "aws_vpc" "vpc": 21: resource "aws_vpc" "vpc" {]

components/kyma-environment-broker/internal/process/upgrade_kyma_operation_test.go
vvxxvvxx: Thanks. changed to "errors".

wozniakjan:
correcting my previous comment for other reviewers, I was educated by @vvxxvvxx and @ebensom a bit on prometheus storage.

> Number of time-series in prometheus db is a Cartesian product of all labels.

this statement I made is incorrect. Thanks to @ebensom for correcting me, sharing his much more accurate quote here

> Every unique key-value labelset creates a new time series, but not a cartesian product result.

Another thing I said that was incorrect

> It will grow exponentially.

The above-mentioned quote from @ebensom makes a big difference because it's unlikely that every instance will face every error combination, the growth will most likely be polynomial, close to linear.

But the discussion had an outcome that we will likely still want to remove `description` as a label with full error message. @ebensom provided good analysis so copying it here too

> I know it could aid us in including the error message in the alerting rule (alert message description), however in practice it could result in high churn rate and sparse TSDB which could affect Prometheus storage performance (both inserts and queries).

vvxxvvxx:
The label `error_msg` has been removed and the label name `error_component` was changed to error_category (may be better to understand..).

wozniakjan:
I guess this PR is currently waiting for https://github.com/kyma-project/control-plane/pull/1405, is that right? Because that one is in provisioner, I can't help much there but after #1405 merges, I will be happy to approve this too and see how it performs.

vvxxvvxx:
> I guess this PR is currently waiting for #1405, is that right? Because that one is in provisioner, I can't help much there but after #1405 merges, I will be happy to approve this too and see how it performs.

Yes, this PR is waiting for #1405 to be reviewed and merged first...

wozniakjan (approved):
/lgtm

This was a large PR that was rather difficult to review but I don't have any factual objections.

I think with our current state, we should rather strive for smaller PRs implementing a single change. Alternatively, keeping a commit per change scope and then rebasing the PR branch rather than merging main to it could make reviews in the future easier, limiting to a reasonable number of commits per PR. 57 is way too many, I would rather keep the commit count in single digit numbers.

PR #1370, opened 2022-02-18
Fix setting/restoring AVS maintenance during Kyma 2 upgrade

ebensom:
**Description**

- Extract Setting & restoring AVS maintenance status to functions outside of Initialization step
- Set AVS statuses in maintenance correctly in Initialization step for both Kyma 1 and Kyma 2 runtimes
- Restore AVS statuses at the end of processing in Initialization step for Kyma 1 runtimes
- Restore AVS statuses at the end of processing in Check_Cluster_Configuration step for Kyma 2 runtimes


wozniakjan (approved):
blind lgtm

PR #1390, opened 2022-02-24
Ers tool

ralikio:
**Description**

- HTTP Client for ERS,
- 

**Related issue(s)**


tools/cli/pkg/ers/client/client.go
```diff
@@ -0,0 +1,54 @@
...
+	return &ersClient{
+		url,
+		client,
+	}
+}
+
+func (c *ersClient) GetOne(instanceId string) (*ers.Instance, error) {
+	instances, err := c.client.get(fmt.Sprintf(environmentsPath+"?"+idParam, c.url, instanceId))
+	if err != nil {
+		return nil, errors.Wrap(err, "Error while sending request")
```
tillknuesting, line 30: %w is missing

tools/cli/pkg/ers/client/http.go
```diff
@@ -0,0 +1,78 @@
...
+
+func NewHttpClient(logger logger.Logger, client *http.Client) *HttpClient {
+	return &HttpClient{
+		logger,
+		client,
+	}
+}
+
+func (c *HttpClient) put(url string) error {
+	c.do(nil, func() (resp *http.Response, err error) {
```
tillknuesting, line 26: error value of `c.do` is not used that could be return by put

tools/cli/pkg/ers/client/http.go
```diff
@@ -0,0 +1,78 @@
...
+	c.logger.Debugf("Sending request to %s", resp.Request.URL)
+
+	if err != nil {
+		return errors.Wrap(err, "Error while sending request")
+	}
+
+	c.logger.Debugf("Response status: %s", resp.Status)
+
+	defer func() {
+		resp.Body.Close()
```
tillknuesting, line 59: `defer resp.Body.Close() ` would be shorter

tools/cli/pkg/ers/client/http.go
```diff
@@ -0,0 +1,78 @@
...
+		logger,
+		client,
+	}
+}
+
+func (c *HttpClient) put(url string) error {
+	c.do(nil, func() (resp *http.Response, err error) {
+		req, err := http.NewRequest("PUT", url, nil)
+		if err != nil {
+			return nil, err
```
tillknuesting, line 29: consider to wrap the error with additional information

tools/cli/pkg/ers/fetcher/file.go
```diff
@@ -12,25 +12,25 @@ type FileClient struct {
...
 		filename: filename,
 	}
 }
 
-func (c *FileClient) GetInstanceById(id string) (ers.Instance, error) {
+func (c *FileClient) GetInstanceById(id string) (*ers.Instance, error) {
 	instances, err := c.GetAllInstances()
 	if err != nil {
-		return ers.Instance{}, err
+		return &ers.Instance{}, err
```
tillknuesting, line 24: consider to wrap error with more information

tools/cli/pkg/ers/client/api.go
```diff
@@ -0,0 +1,13 @@
+package client
+
+import "github.com/kyma-project/control-plane/tools/cli/pkg/ers"
+
+type Client interface {
+	GetOne(id string) (*ers.Instance, error)
```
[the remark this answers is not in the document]
ralikio: Isn't `id` correct here? The mentioned section from [](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) states `Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url"`. So as long it is the same case it seems correct.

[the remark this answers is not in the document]
tillknuesting: "This rule also applies to "ID" when it is short for "identifier" (which is pretty much all cases when it's not the "id" as in "ego", "superego"), so write "appID" instead of "appId"."

tools/cli/pkg/ers/client/http.go
```diff
@@ -0,0 +1,86 @@
...
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/kyma-project/control-plane/tools/cli/pkg/ers"
+	"github.com/kyma-project/control-plane/tools/cli/pkg/logger"
+	"github.com/pkg/errors"
+)
+
+const timeoutInMilli = 400
```
tillknuesting, line 15: 400 Millisecond might be a bit optimistic as larger data might need longer to transmit

tools/cli/pkg/ers/client/client.go
```diff
@@ -0,0 +1,54 @@
...
+
+func NewErsClient(url string, client *HTTPClient) Client {
+	return &ersClient{
+		url,
+		client,
+	}
+}
+
+func (c *ersClient) GetOne(instanceID string) (*ers.Instance, error) {
+	instances, err := c.client.get(fmt.Sprintf(environmentsPath+"?"+idParam, c.url, instanceID))
```
piotrmiskiewicz, line 28: idParam - one parameter, but you are providing two

tools/cli/pkg/ers/client/http.go
piotrmiskiewicz: I would say even more than 3 seconds

tools/cli/pkg/ers/command/instances.go
```diff
@@ -42,17 +45,37 @@ type Filters struct {
...
+		TokenURL:     GlobalOpts.OauthUrl(),
+	}
+	configClient := config.Client(context.Background())
+
 	if c.source != "" {
-		c.instanceFetcher = client.NewFileClient(c.source)
+		c.instanceFetcher = fetcher.NewFileClient(c.source)
 	} else {
+		// create a shared ERS HTTP client which does the oauth flow
+		httpClient := client.NewHTTPClient(logger.New(), configClient)
```
piotrmiskiewicz, line 74: as I see, the http client is not shared. for instances operation it does not matter, but for migration - yes

tools/cli/pkg/ers/client/client.go
ralikio: It just a format string. Params include URL and id.

tools/cli/pkg/ers/command/broker.go
```diff
@@ -1,15 +1,50 @@
...
+	cobraCmd *cobra.Command
+	brokerId string
+}
 
 func NewSwitchBrokerCommand() *cobra.Command {
-	cmd := &cobra.Command{
-		Use:   "switch-broker",
+	cmd := &SwitchCommand{}
+	cobraCmd := &cobra.Command{
+		Use:   "switch [id]",
```
piotrmiskiewicz, line 18: fix also an example

PR #1406, opened 2022-03-02
Remove RuntimeID from KEB metrics

wozniakjan:
**Description**
There are duplicate time-series for the same instance, once with `RuntimeID` and once without. For example:
```
compass_keb_provisioning_result{global_account_id="e449f875-b5b2-4485-b7c0-98725c0571bf",instance_id="test-jw148",operation_id="1d02bba0-20cc-4203-8d1d-84d2d2e21eb5",plan_id="361c511f-f939-4621-b228-d0fb79a1fe15",runtime_id=""} 2
compass_keb_provisioning_result{global_account_id="e449f875-b5b2-4485-b7c0-98725c0571bf",instance_id="test-jw148",operation_id="1d02bba0-20cc-4203-8d1d-84d2d2e21eb5",plan_id="361c511f-f939-4621-b228-d0fb79a1fe15",runtime_id="588a6fc4-ea11-407b-9906-25fcb2119a4e"} 2
```
The reason for duplicated `compass_keb_provisioning_result` time-series for each `InstanceID` is that KEB updates the value after each step in the provisioning queue. The problem is, steps before `NewCreateRuntimeStep` for kyma1 and `NewCreateRuntimeWithoutKymaStep` for kyma2 don't have `RuntimeID` yet, so that is why there is one `compass_keb_provisioning_result` without `RuntimeID` tracking progress of steps until the creation of runtime step and then second `compass_keb_provisioning_result` with `RuntimeID` tracking progress of steps after the create runtime step.

It was decided that we should be fine removing `runtime_id` from all KEB metrics as a similar uniqueness guarantee is provided by `InstanceID` x `GlobalAccountID`.

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/metrics/operation_duration.go
+++ b/components/kyma-environment-broker/internal/metrics/operation_duration.go
@@ -25,14 +25,14 @@ func NewOperationDurationCollector() *OperationDurationCollector {
 			Name:      "provisioning_duration_minutes",
 			Help:      "The time of the provisioning process",
 			Buckets:   prometheus.LinearBuckets(20, 2, 40),
-		}, []string{"operation_id", "runtime_id", "instance_id", "global_account_id", "plan_id"}),
+		}, []string{"operation_id", "instance_id", "global_account_id", "plan_id"}),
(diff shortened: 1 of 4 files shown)


tillknuesting (approved):
lgtm

PR #1408, opened 2022-03-02
Change the level of logs regarding "cluster kubeconfig is nil" error

mvshao:
**Description**
Change the text in the log to indicate that empty Kubeconfig is a warning, not an error.

**Changes proposed in this pull request**

- change the log message from error to warning

### DIFF
--- a/components/provisioner/internal/operations/executor.go
+++ b/components/provisioner/internal/operations/executor.go
@@ -115,6 +115,10 @@ func (e *Executor) process(operation model.Operation, cluster model.Cluster, log
 
 		result, err := step.Run(cluster, operation, log)
 		if err != nil {
+			if err.Error() == "cluster kubeconfig is nil" {
+				log.Warnf("Warning, the %s", err.Error())
+				return false, 0, nil
+			}
 			log.Warnf("error while processing operation, stage failed: %s", err.Error())
 			return false, 0, err
 		}
--- a/components/provisioner/internal/operations/executor.go
+++ b/components/provisioner/internal/operations/executor.go
@@ -19,6 +19,8 @@ const (
 	backOffDirectorDelay = 1 * time.Second
 )
 
+var ErrKubeconfigNil = errors.New("cluster kubeconfig is nil")
+
 func NewExecutor(
 	session dbsession.ReadWriteSession,
 	operation model.OperationType,
@@ -115,7 +117,7 @@ func (e *Executor) process(operation model.Operation, cluster model.Cluster, log
 
 		result, err := step.Run(cluster, operation, log)
 		if err != nil {
-			if errors.Is(err, errors.New("cluster kubeconfig is nil")) {
+			if errors.Is(err, ErrKubeconfigNil) {
 				log.Warnf("Warning, the %s", err.Error())
 				break
 			}
--- a/components/provisioner/internal/operations/executor.go
+++ b/components/provisioner/internal/operations/executor.go
@@ -115,7 +115,7 @@ func (e *Executor) process(operation model.Operation, cluster model.Cluster, log
 
 		result, err := step.Run(cluster, operation, log)
 		if err != nil {
-			if err.Error() == "cluster kubeconfig is nil" {
+			if errors.Is(err, errors.New("cluster kubeconfig is nil")) {
 				log.Warnf("Warning, the %s", err.Error())
 				break
 			}
--- a/components/provisioner/internal/operations/executor.go
+++ b/components/provisioner/internal/operations/executor.go
@@ -117,7 +117,7 @@ func (e *Executor) process(operation model.Operation, cluster model.Cluster, log
 		if err != nil {
 			if err.Error() == "cluster kubeconfig is nil" {
 				log.Warnf("Warning, the %s", err.Error())
-				return false, 0, nil
+				break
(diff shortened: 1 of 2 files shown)


components/provisioner/internal/operations/executor.go
```diff
@@ -115,6 +115,10 @@ func (e *Executor) process(operation model.Operation, cluster model.Cluster, log
 
 		result, err := step.Run(cluster, operation, log)
 		if err != nil {
+			if err.Error() == "cluster kubeconfig is nil" {
```
VOID404, line 118: I don't think this is the recommended way to differentiate errors since Go 1.13
Maybe this code should use `errors.Is`/`errors.As`?
https://pkg.go.dev/errors

components/provisioner/internal/operations/executor.go
```diff
@@ -115,6 +115,10 @@ func (e *Executor) process(operation model.Operation, cluster model.Cluster, log
 
 		result, err := step.Run(cluster, operation, log)
 		if err != nil {
+			if errors.Is(err, errors.New("cluster kubeconfig is nil")) {
```
VOID404, line 118: If this is a known error with a const message, wouldn't it make sense to create it once, export it and reuse from variable?

Example from Go's std:
```go
var (
	ErrInvalid    = errors.New("invalid argument")
	ErrPermission = errors.New("permission denied")
	ErrExist      = errors.New("file already exists")
	ErrNotExist   = errors.New("file does not exist")
	ErrClosed     = errors.New("file already closed")
)

Comment (VOID404) [reply components/provisioner/internal/operations/executor.go:118]: Benefits of this solution would be visible with any changes around this error in the future. From changing anything about the error (message, structure), to finding all places where given error is mentioned with refactoring tools

Comment (VOID404) [reply components/provisioner/internal/operations/executor.go:118]: So to summarise, it should look something like this:
```go
var ErrKubeconfigNil = errors.New("cluster kubeconfig is nil")

func (s *CreateBindingsForOperatorsStep) Run(cluster model.Cluster, _ model.Operation, log logrus.FieldLogger) (operations.StageResult, error) {
	if cluster.Kubeconfig == nil {
		return operations.StageResult{}, ErrKubeconfigNil
	}
	// ...
}

func (e *Executor) process(operation model.Operation, cluster model.Cluster, logger logrus.FieldLogger) (bool, time.Duration, error) {
  //...
  if err != nil {
	  if errors.Is(err, ErrKubeconfigNil) {
		  log.Warnf("Warning, the %s", err.Error())
		  break
	  }
          log.Warnf("error while processing operation, stage failed: %s", err)
          return false, 0, err
  }
  //...
}
```

components/provisioner/internal/operations/executor.go
```diff
@@ -115,6 +115,10 @@ func (e *Executor) process(operation model.Operation, cluster model.Cluster, log
...
+				log.Warnf("Warning, the %s", err.Error())
```
VOID404, line 119:
```suggestion
				log.Warnf("Warning, the %s", err)
```

Calling `.Error()` here is pointless

components/provisioner/internal/operations/executor.go
mvshao: This is the one time error in code. In my opinion in this fix, it is not necessary to export a single variable. It might be a great idea to refactor the way of logging in the whole component.

VOID404: IMO it might make sense to refactor the area being worked on instead of one big refactor task

akgalwas: I agree with @VOID404 . We can apply this change as this is a good practice.

VOID404 (changes requested):
Other than [this](https://github.com/kyma-project/control-plane/pull/1408#discussion_r817734796), looks good

VOID404 (approved):
/lgtm

akgalwas (approved):

components/provisioner/internal/operations/executor.go
vvxxvvxx: it's now still err.Error()..
btw, may I ask why `break` when `errors.Is(err, ErrKubeconfigNil)` in executor? After break, it will go to `e.updateOperationStatus(logger, operation.ID, "Operation succeeded", model.Succeeded, time.Now())`, then if `ErrKubeconfigNil`, the provision operation will be considered as succeeded?

vvxxvvxx: it's now still err.Error()..
btw, may I ask why `break` when `errors.Is(err, ErrKubeconfigNil)` in executor? After break, it will go to `e.updateOperationStatus(logger, operation.ID, "Operation succeeded", model.Succeeded, time.Now())`, then if `ErrKubeconfigNil`, the provision operation will be considered as succeeded?

components/provisioner/internal/operations/executor.go
vvxxvvxx: May I ask why break when errors.Is(err, ErrKubeconfigNil) in executor? After break, it will go to e.updateOperationStatus(logger, operation.ID, "Operation succeeded", model.Succeeded, time.Now()), then if ErrKubeconfigNil, the provision operation will be considered as succeeded?

PR #1417, opened 2022-03-03
KEB bump gardener dependency to v1.41.1

szwedm:
**Description**

- update github.com/gardener/gardener to v1.41.1 with all its deps to address security issues

**Related issue(s)**


szwedm:
Hold reason: problem with the newest buildpack-golang image

piotrmiskiewicz (approved):

components/kyma-environment-broker/go.mod
```diff
@@ -154,9 +156,9 @@ replace (
 
 	// NOTE: some dependencies require old style client-go version k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible
 	// github.com/gardener/hvpa-controller, github.com/kyma-project/kyma/components/compass-runtime-agent, github.com/kyma-project/control-plane/components/provisioner, github.com/gardener/gardener
-	k8s.io/api => k8s.io/api v0.22.3
+	k8s.io/api => k8s.io/api v0.23.4
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.19.12
```
wozniakjan, line 160: nit: curious if the version skew between `apiextensions-apiserver` and other `k8s.io` modules can cause issues

I guess it's not very likely as it might be just transitive dependency whose code is not executed in KEB
```
$ go mod why k8s.io/apiextensions-apiserver
# k8s.io/apiextensions-apiserver
(main module does not need package k8s.io/apiextensions-apiserver)
```
but probably worth bumping some day too

wozniakjan (approved):

PR #1441, opened 2022-03-11
Remove obsoleted avs-bridge related code from KEB

ebensom:
**Description**

avs-bridge is no longer deployed with Kyma 2. All related code (override setting, API Key configuration) is obsolete, no longer used (dead code), and so is removed.


wozniakjan (approved):
/lgtm

PR #1443, opened 2022-03-11
Set provisioning parameters when processing deprovisioning operation

wozniakjan:
**Description**

When the deprovisioning operation is triggered while the provisioning operation is still in progress, KEB reports confusing metrics and duplicate time-series for each `instance_id`.

```
compass_keb_deprovisioning_result{global_account_id="",instance_id="test-jw160",operation_id="782f1126-fc42-44c9-98a5-b4b8686f1a44",plan_id=""} 0
compass_keb_deprovisioning_result{global_account_id="e449f875-b5b2-4485-b7c0-98725c0571bf",instance_id="test-jw160",operation_id="782f1126-fc42-44c9-98a5-b4b8686f1a44",plan_id="361c511f-f939-4621-b228-d0fb79a1fe15"} 2
```

This is because `initialization_step` in the provisioning queue detects the provisioning operation is not finished and tells the manager to put the processing of this back in the queue before getting `ProvisioningParameters`.

The second bug was that the switch-case statement translating operation status to operation status metrics didn't contain `pending` at all, falling through the statement and keeping the default value for float64, which is 0.

This PR adds the information to the runtime object for metrics processing as well as `pending` case and it seems to have resolved the issue, correctly reporting deprovisioning operation in pending status.

```
compass_keb_deprovisioning_result{global_account_id="e449f875-b5b2-4485-b7c0-98725c0571bf",instance_id="test-jw166",operation_id="cbb9157f-edb8-4b40-875a-59d2fefd46cd",plan_id="361c511f-f939-4621-b228-d0fb79a1fe15"} 3
```

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
+++ b/components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
@@ -99,6 +99,9 @@ func (s *InitialisationStep) run(operation internal.DeprovisioningOperation, log
 	}
 	if op.State == domain.InProgress {
 		log.Info("waiting for provisioning operation to finish")
+		// This is only in memory copy because metrics depend on provisioning parameters being available, this doesn't persist them in KEB database
+		operation.SubAccountID = operation.ProvisioningParameters.ErsContext.SubAccountID
+		operation.ProvisioningParameters = op.ProvisioningParameters
 		return operation, time.Minute, nil
 	}
 	operation, repeat := s.operationManager.UpdateOperation(operation, func(operation *internal.DeprovisioningOperation) {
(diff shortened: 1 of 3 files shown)


components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go
```diff
@@ -99,6 +99,8 @@ func (s *InitialisationStep) run(operation internal.DeprovisioningOperation, log
 	}
 	if op.State == domain.InProgress {
 		log.Info("waiting for provisioning operation to finish")
+		operation.SubAccountID = operation.ProvisioningParameters.ErsContext.SubAccountID
+		operation.ProvisioningParameters = op.ProvisioningParameters
```
piotrmiskiewicz, line 103: please, add a comment that we are copying those values because of metrics

piotrmiskiewicz (approved):

PR #1462, opened 2022-03-17
Fix bug in last error "converting NULL to string is unsupported"

vvxxvvxx:
**Description**

- Fixed the bug in PR#1405 to set default value of last error fields to ''.

**Related issue(s)**


components/schema-migrator/migrations/provisioner/202203171600_operation_alter_last_error.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE operation
```
koala7659, line 1: If you change default values in migration script - you should also update them in provisioner/assets/database/provisioner.sql file

koala7659 (changes requested):

components/schema-migrator/migrations/provisioner/202203171600_operation_alter_last_error.up.sql
vvxxvvxx: added `NOT NULL` in provisioner.sql for last error fields..

koala7659 (changes requested):
There should be a respective step down added in database migration

vvxxvvxx:
> There should be a respective step down added in database migration

this is the addition to`202202241500_operation_last_error.up.sql`.
Is `202202241500_operation_last_error.down.sql` enough for the down? this will remove the columns added.

vvxxvvxx:
> There should be a respective step down added in database migration

Added the down step.

vvxxvvxx:
huskies are looking into the error in `pre-main-control-plane-reconciler-upgrade-kyma2-latest-to-main-gardener`

Maladie (approved):

koala7659 (approved):
LGTM

PR #1480, opened 2022-03-22
update dashboard

ruanxin:
**Description**

The equivalent dashboard can be found here: 
https://grafana.cp.dev.kyma.cloud.sap/d/efa86fd1d0c121a26444b636a3f509a9344/reconciler-for-testing?orgId=1&refresh=10s

**Related issue(s)**


jakobmoellersap (changes requested):
I summarized my findings in the review summary since its hard to document this in the json, please reach out if you need to ask for details.

Could we do an average aggregation on the Successful Processing 95th percentile for all instances or is it planned to monitor instances individually?

Also for mothership, the deleted status will rise in the future heavily, disturbing the other metrics for Cluster Status. Maybe we should exclude the deleted ones and have them count up in an individual KPI. I think we should either way find a better visualization.

Reconciler Memory / CPU as well as Rate of Packets dropped /Rate of transmitted Packets dropped shows no data, not sure if this is intended.

ruanxin:
> Also for mothership, the deleted status will rise in the future heavily, disturbing the other metrics for Cluster Status. Maybe we should exclude the deleted ones and have them count up in an individual KPI. I think we should either way find a better visualization.

Yes, make sense, deleted state has been excluded

ruanxin:
> Reconciler Memory / CPU as well as Rate of Packets dropped /Rate of transmitted Packets dropped shows no data, not sure if this is intended.

Yes, that's correct, there is no package dropped during transmitting, a good sign:)

ruanxin:
follow up refactoring issue: https://github.com/kyma-incubator/reconciler/issues/969

jakobmoellersap (approved):
Approved under condition of later refactoring in PR mentioned above.

jakobmoellersap (approved):

adityabhatia (approved):

PR #1481, opened 2022-03-23
Fix DNS config in DB

### LINKED ISSUE #1290: Provisioner does not store shoot DNS configuration in the database
There was an extension of an input Gardener cluster configuration merged in [PR 970](https://github.com/kyma-project/control-plane/pull/970/)  where DNS configuration was added to GraphQL GardenerConfig input struct to add support for DNS providers and custom domain names for provisioned shoots.

While working on fixing other problem I found that Provisioner does not preserve this DNS configuration used to create shoot cluster in its database.  

For details please check the code that inserts cluster configuration into database:

https://github.com/kyma-project/control-plane/blob/main/components/provisioner/internal/provisioning/persistence/dbsession/write.go#L97

You can see that we do not store DNS config.

We should make sure DNS config is stored in the database.

VOID404:
**Description**

Provisioner doesn't store shoot's DNS config in database.
DNS config is not returned with shoot's info when querying provisioner.

- Add DNS config to database
- Add DNS config to graphql response

**Related issue(s)**

Resolves #1290


components/provisioner/internal/provisioning/persistence/dbsession/read.go
```diff
@@ -511,3 +519,46 @@ func (r readSession) getOidcConfig(gardenerConfigID string) (model.OIDCConfig, d
...
+		return nil, nil
+	}
+	if err != nil {
+		return nil, dberrors.Internal("Failed to get DNS config: %s", err)
+	}
+
+	dnsConfig := dnsConfigWithID.DNSConfig
+
+	_, err = r.session.
+		Select("*").
```
franpog859, line 549: In both of these `Select`s (this one and the one above), I'd suggest explicitly listing the table columns instead of getting all of them with the `*` sign. Provisioner is not aware of the real schema with potential changes that are provided by Schema Migrator so its behavior could change if we leave it like this

components/provisioner/internal/provisioning/persistence/dbsession/read.go
```diff
@@ -131,6 +133,12 @@ func (r readSession) GetCluster(runtimeID string) (model.Cluster, dberrors.Error
 	}
 	cluster.ClusterConfig.OIDCConfig = &oidcConfig
 
+	dnsConfig, dberr := r.getDNSConfig(providerConfig.ID)
+	if dberr != nil {
+		return model.Cluster{}, dberr.Append("Cannot get Oidc config for runtimeID: %s", runtimeID)
```
franpog859, line 138: It's the DNS config, not the OIDC one :D

components/provisioner/assets/database/provisioner.sql
```diff
@@ -181,4 +181,26 @@ CREATE TABLE signing_algorithms
...
+CREATE TABLE dns_config
+(
+    id uuid PRIMARY KEY CHECK (id <> '00000000-0000-0000-0000-000000000000'),
+    domain varchar(256) NOT NULL,
+    gardener_config_id uuid NOT NULL,
+    unique(gardener_config_id),
+    foreign key (gardener_config_id) REFERENCES gardener_config (id) ON DELETE CASCADE
+);
+
+-- DNS config
```
franpog859, line 196: It's the DNS providers table

franpog859 (commented):
I added some minor comments. Proceeding to the manual tests

franpog859 (commented):
I tested it manually and it looks like the `domainsInclude` and `primary` fields didn't match as they should do. While provisioning I specified `domainsInclude: ["domainsInclude"]` and `primary: true` but in the Runtime Status query I got `"domainsInclude": [""]` and `"primary": false`. I sent you the screenshots in a private message

components/provisioner/internal/provisioning/persistence/dbsession/read.go
VOID404: 🤦

franpog859 (approved):
LGTM. All the manual tests passed

franpog859 (approved):

PR #1491, opened 2022-03-25
Do not use empty region

piotrmiskiewicz:
**Description**

- if user provides an empty region in provisioning parameters, KEB should do the same as the region is nil

### DIFF
--- a/components/kyma-environment-broker/cmd/broker/provisioning_test.go
+++ b/components/kyma-environment-broker/cmd/broker/provisioning_test.go
@@ -42,6 +42,38 @@ func TestProvisioning_HappyPath(t *testing.T) {
 	suite.AssertProvisioningRequest()
 }
 
+func TestProvisioning_TrialWithEmptyRegion(t *testing.T) {
+	// given
+	suite := NewBrokerSuiteTest(t)
+	defer suite.TearDown()
+	iid := uuid.New().String()
+
+	// when
+	resp := suite.CallAPI("PUT", fmt.Sprintf("oauth/v2/service_instances/%s?accepts_incomplete=true", iid),
+		`{
+					"service_id": "47c9dcbf-ff30-448e-ab36-d3bad66ba281",
+					"plan_id": "7d55d31d-35ae-4438-bf13-6ffdfa107d9f",
+					"context": {
+						"sm_platform_credentials": {
+							  "url": "https://sm.url",
+							  "credentials": {}
+					    },
+						"globalaccount_id": "g-account-id",
+						"subaccount_id": "sub-id",
+						"user_id": "john.smith@email.com"
+					},
+					"parameters": {
+						"name": "testing-cluster",
+                        "region":""
+					}
+		}`)
+	opID := suite.DecodeOperationID(resp)
+	suite.processProvisioningByOperationID(opID)
+
+	// then
+	suite.AssertAWSRegionAndZone("eu-west-1")
+}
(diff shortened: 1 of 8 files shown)


components/kyma-environment-broker/internal/process/input/input.go
```diff
@@ -428,7 +428,9 @@ func (r *RuntimeInput) applyProvisioningParametersForProvisionRuntime() error {
 	updateInt(&r.provisionRuntimeInput.ClusterConfig.GardenerConfig.AutoScalerMax, params.AutoScalerMax)
 	updateInt(r.provisionRuntimeInput.ClusterConfig.GardenerConfig.VolumeSizeGb, params.VolumeSizeGb)
 	updateString(&r.provisionRuntimeInput.ClusterConfig.GardenerConfig.Name, r.shootName)
-	updateString(&r.provisionRuntimeInput.ClusterConfig.GardenerConfig.Region, params.Region)
+	if params.Region != nil && *params.Region != "" {
```
piotrmiskiewicz, line 431: I'm thinking about having this "if" statement inside the updateString function, but not sure if it breaks something

components/kyma-environment-broker/cmd/broker/provisioning_test.go
```diff
@@ -42,6 +42,38 @@ func TestProvisioning_HappyPath(t *testing.T) {
...
+							  "url": "https://sm.url",
+							  "credentials": {}
+					    },
+						"globalaccount_id": "g-account-id",
+						"subaccount_id": "sub-id",
+						"user_id": "john.smith@email.com"
+					},
+					"parameters": {
+						"name": "testing-cluster",
+                        "region":""
```
wozniakjan, line 67: nit: there is a mix of tabs and spaces here

components/kyma-environment-broker/internal/process/input/input.go
wozniakjan: yeah, I agree that for some arguments there could be a difference between providing empty value or not providing value at all, probably safer to keep it separate.

components/kyma-environment-broker/internal/provider/azure_provider.go
```diff
@@ -144,8 +146,11 @@ func (p *AzureTrialInput) ApplyParameters(input *gqlschema.ClusterConfigInput, p
 		}
 	}
 
-	if params.Region != nil {
+	if params.Region != nil && *params.Region != "" {
+		fmt.Println(*params.Region)
 		updateString(&input.GardenerConfig.Region, toAzureSpecific[*params.Region])
+		fmt.Println(input.GardenerConfig.Region)
+		fmt.Println(toAzureSpecific[*params.Region])
```
wozniakjan, line 153: nit: are these `fmt.Println` statements intended to stay?

wozniakjan (commented):
implementation looks solid, just take a look at the nits provided below please

components/kyma-environment-broker/internal/provider/azure_provider.go
piotrmiskiewicz: removed

components/kyma-environment-broker/cmd/broker/provisioning_test.go
piotrmiskiewicz: fixed

wozniakjan (approved):

PR #1507, opened 2022-03-29
Simple Stats

ralikio:
**Description**

- Simple stats for displaying `err`, `done` and `all` counters.

**Related issue(s)**

### DIFF
--- a/tools/cli/pkg/ers/command/migrate_all.go
+++ b/tools/cli/pkg/ers/command/migrate_all.go
@@ -40,6 +40,7 @@ func NewMigrationAllCommand(log logger.Logger) *cobra.Command {
 	cobraCmd.Flags().BoolVarP(&cmd.dryRun, "mock-ers", "", true, "Use fake ERS client to test")
 
 	cmd.corbaCmd = cobraCmd
+	cmd.stats = NewStats()
 
 	return cobraCmd
 }
@@ -54,6 +55,7 @@ type MigrationAllCommand struct {
 	log       logger.Logger
 	ersClient client.Client
 	dryRun    bool
+	stats     *Stats
 }
 
 func (c *MigrationAllCommand) Run() error {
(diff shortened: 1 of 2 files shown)


tools/cli/pkg/ers/command/migrate_all.go
```diff
@@ -115,18 +120,23 @@ func (c *MigrationAllCommand) simpleWorker(workerId int, workChannel chan ers.Wo
...
 			c.wg.Done()
+			c.stats.Done()
 			continue
 		}
 		refreshed, err := c.ersClient.GetOne(instance.Id)
 		if err != nil {
 			c.log.Warnf("[Worker %d] GetOne error: %s", workerId, err.Error())
 			c.wg.Done()
+			c.stats.Err(instance.Id, err)
+			// TODO: add retries
```
tillknuesting, line 131: In case you need a snipped for a retry mechanism:

```
package main

import (
	"context"
	"errors"
	"fmt"
	"log"
	"time"
)


var count int

// Effector type defines a function signature.
type Effector func(context.Context) (string, error)

// Retry accepts an Effector and returns an anonymous function with the same signature as the Effector
// It essentially wraps the received function with the retry logic.
func Retry(effector Effector, retries int, delay time.Duration) Effector {
	return func(ctx context.Context) (string, error) {
		for r := 0; ; r++ {
			response, err := effector(ctx)
			if err == nil || r >= retries {
				return response, err
			}

			log.Printf("attempt %d failed; retrying in %v", r+1, delay)

			select {
			case <-time.After(delay):
			case <-ctx.Done():
				if ctx.Err() != nil {
					return "", fmt.Errorf("anonymous function from retry: %w", ctx.Err())
				}

				return "", nil
			}
		}
	}
}

func EmulateTransientError(ctx context.Context) (string, error) {
	count++

	if count <= 3 {
		return "intentional fail", errors.New("error")
	}

	return "success", nil
}

func main() {
	r := Retry(EmulateTransientError, 3, 1*time.Second)
	res, err := r(context.TODO())

	fmt.Println(res, err)
}



```

piotrmiskiewicz (approved):

Issue #1471, opened 2022-03-18
Remove unnecessary dependency to Gardener in KEB


PK85:
**Description**

We are getting more and more security issues regarding our dependencies. We need to avoid them if we could achieve what we want with 10 lines of code. One of the example is our dependency to gardener where we do a simple call to API.

**Reasons**

Less dependencies less issues. We need to take into account that importing new stuff makes our KEB application unnecessary big and open to security vulnerabilities.

**AC**

- remove dependency to gardener - copy required code to our repo. Keep it small, use only what we need without all of those transitive dependencies

wozniakjan:
this task turned into a slightly larger problem.

KEB depends on 4 packages
```
$ grep . --exclude-dir=vendor --include=*.go -nre 'gardener/gardener' | awk -F '"' '{print($2)}' | sort | uniq
github.com/gardener/gardener/pkg/apis/core/v1beta1
github.com/gardener/gardener/pkg/client/core/clientset/versioned/fake
github.com/gardener/gardener/pkg/client/core/clientset/versioned/typed/core/v1beta1
github.com/gardener/gardener/pkg/client/core/clientset/versioned/typed/core/v1beta1/fake
```

From `github.com/gardener/gardener/pkg/apis/core/v1beta1` it needs `Shoot` CRD which depends on pretty much the whole gardener API package and a few extra outside APIs.

Our tests also need `clientset` and that is huge by itself. Copying the necessary code from the gardener turns into 241 files copy/pasted. At this point, how about we use unstructured for this gardener feature? Might be better than copying a large number of files.

wozniakjan:
> * remove dependency to gardener - copy required code to our repo. Keep it small, use only what we need without all of those transitive dependencies

just sharing an outcome from a discussion, we have a consensus about using unstructured instead of Gardener API given what KEB needs. Should be the easiest path forward and KEB shouldn't have that many false positives from security scanners.

PR #1514, opened 2022-03-31
[WIP] Remove dep on gardener

wozniakjan:
**Description**

Gardener is a rich resource of reported vulnerabilities by security scanners. Given KEB needs a very small part of Gardener API, it was considered to try copy/paste instead of vendoring the whole Gardened module.

Unfortunately, this looks like it will be rather difficult to maintain because even the small subset of Gardener API appears to include many other packages that transitively require different packages.

**Related issue(s)**

https://github.com/kyma-project/control-plane/issues/1471


wozniakjan:
closing in favour of https://github.com/kyma-project/control-plane/pull/1527

PR #1542, opened 2022-04-07
fix security vulnerability in kmc

FriedrichWilken:
This PR bumps some dependencies to fix security vulnerabilities:
`golang.org/x/crypto/ssh before 0.0.0-20220314234659-1baeb1ce4c0b in Go through 1.16.15 and 1.17.x through 1.17.8 allows an attacker to crash a server in certain circumstances involving AddHostKey.
`


k15r (approved):
Moar approvals, moar good ;)

PR #1561, opened 2022-04-12
Enable NAT Gateway in Provisioner

Maladie:
**Description**

- Adds optional NAT gateway parameters to Provisioner's schema

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/13553


janmedrek:
@Maladie, a comment regarding the default behavior - if we perform an upgrade of the cluster we need to make sure that NAT Gateway does not magically pop up if it was not used before and it was not passed as an input. We need to rule out that possibility as it may break the cluster connectivity to other services and create a serious downtime.

From what I see it won't be the case, just want to put some attention to this topic as it can get quite serious.

components/provisioner/internal/model/infrastructure.go
```diff
@@ -58,6 +58,16 @@ func NewAzureInfrastructure(workerCIDR string, azConfig AzureGardenerConfig) *az
 		},
 		Zoned: isZoned,
 	}
+
+	if isZoned && azConfig.input.EnableNatGateway != nil {
+		natGateway := azure.NatGateway{
+			Enabled:                      *azConfig.input.EnableNatGateway,
+			IdleConnectionTimeoutMinutes: util.UnwrapIntOrDefault(azConfig.input.IdleConnectionTimeoutMinutes, 4),
```
franpog859, line 65: Maybe we could specify some constant named `defaultNatGatewayIdleConnectionTimeoutMinutes` instead of some mysterious `4`. What do you think?

franpog859 (commented):
Looks ok. Proceeding to the manual tests

franpog859 (commented):
While testing it manually I encountered a problem with the shoot upgrade. The new values are updated in the database so they are visible in the `runtimeStatus` query:
```graphql
query {
  runtimeStatus(id: "some-id") {
    runtimeConfiguration {
      clusterConfig {
        providerSpecificConfig {
          ... on AzureProviderConfig {
            vnetCidr
            zones
            enableNatGateway
            idleConnectionTimeoutMinutes
          }
        }
      }
    }
  }
}
```
But they are not updated in the Shoot as they should be after the upgrade 🔴 

There is also one thing, maybe not related to this particular PR. We can specify the `zones` field in the `upgradeShoot` mutation even if it's not a mutable one. If you try so, you get a `\"validation.azure.provider.extensions.gardener.cloud\" denied the request: spec.provider.workers[0].zones: Invalid value: []string{\"some-zone\"}: field is immutable` error

franpog859 (commented):
1. I couldn't pass the `enableNatGateway: false` provisioning parameter due to error: `"Failed to start provisioning, error creating Shoot for 4a381f2a-6e32-4169-bc57-70b8923594db cluster: %!!(MISSING)!(MISSING)s(MISSING), admission webhook \"validation.azure.provider.extensions.gardener.cloud\" denied the request: spec.provider.infrastructureConfig.networks.natGateway: Invalid value: azure.NatGatewayConfig{Enabled:false, IdleConnectionTimeoutMinutes:(*int32)(0xc00bc87758), Zone:(*int32)(nil), IPAddresses:[]azure.PublicIPReference(nil)}: NatGateway is disabled but additional NatGateway config is passed"` 🔴
2. I couldn't upgrade the `enableNatGateway` parameter from `true` to `false` due to error: `Failed to upgrade Cluster: error executing update shoot configuration, All attempts fail:\n#1: admission webhook \"validation.azure.provider.extensions.gardener.cloud\" denied the request: spec.provider.infrastructureConfig.networks.natGateway: Invalid value: azure.NatGatewayConfig{Enabled:false, IdleConnectionTimeoutMinutes:(*int32)(0xc003c8bcf8), Zone:(*int32)(nil), IPAddresses:[]azure.PublicIPReference(nil)}: NatGateway is disabled but additional NatGateway config is passed` 🔴
3. The only parameters that I managed to go through the upgrade mutation are `vnetCidr` and `idleConnectionTimeoutMinutes`. They are changed in the database but still not in the Shoot 🔴

franpog859 (approved):
LGTM

PR #1576, opened 2022-04-13
Inject env vars to component reconciler from chart values

wozniakjan:
**Description**
https://github.com/kyma-incubator/reconciler/pull/1019 introduced enable/disable flag for a scmigration feature. Adding a way to the chart control this via helm values.

### DIFF
--- a/resources/kcp/charts/component-reconcilers/templates/deployment.yaml
+++ b/resources/kcp/charts/component-reconcilers/templates/deployment.yaml
@@ -50,6 +50,13 @@ spec:
         {{- ($.Values.global.mothership_reconciler.features) | nindent 8 }}
         {{- end }}
         {{- end }}
+        {{- if $.Values.component_reconcilers }}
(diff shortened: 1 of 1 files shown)


resources/kcp/charts/component-reconcilers/templates/deployment.yaml
```diff
@@ -50,6 +50,13 @@ spec:
         {{- ($.Values.global.mothership_reconciler.features) | nindent 8 }}
         {{- end }}
         {{- end }}
+        {{- if $.Values.component_reconcilers }}
```
Tomasz-Smelcerz-SAP, line 53: We are going to introduce per-component configuration in another [PR](https://github.com/kyma-project/control-plane/pull/1601).
Nevertheless I will aprove that to bring that into the code, so that we can unify our approaches later - it would be good if all the configuration for given component is located in a single place in the .Values/overrides structure.

Tomasz-Smelcerz-SAP (approved):

PR #1579, opened 2022-04-14
Errors categorization - meaningful alerting for deprovisioning functionality

vvxxvvxx:
**Description**

- Categorize the errors returned for deprovisioning steps (and also keep the original error)
- Expose the details of the error returned including error component(category), error reason, error detailed/original message from the codes to the deprovisioning metric `compass_keb_deprovisioning_result` so as to help quickly locate the error/problem and help statistic of problem.
- Add provisioner timeout error reason

**Related issue(s)**

[backlog #1933](https://github.tools.sap/kyma/backlog/issues/1933)


vvxxvvxx:
will update the go.mod and go.sum for `github.com/kyma-project/control-plane/components/provisioner` after https://github.com/kyma-project/control-plane/pull/1581 is merged.

ebensom (commented):
LGTM

components/kyma-environment-broker/internal/process/deprovisioning/manager.go
```diff
@@ -72,13 +80,26 @@ func (m *Manager) Execute(operationID string) (time.Duration, error) {
 	}
 	operation := *op
 
+	logOperation := m.log.WithFields(logrus.Fields{"operation": operationID, "instanceID": operation.InstanceID})
+
 	provisioningOp, err := m.operationStorage.GetProvisioningOperationByInstanceID(op.InstanceID)
 	if err != nil {
 		m.log.Errorf("Cannot fetch ProvisioningOperation for instanceID %s from storage: %s", op.InstanceID, err)
+
```
vvxxvvxx, line 88: when err!=nil here when getting provisioningOperation, we also need to assign operation.State=domain.Failed, right? as here it gives up the deprovisioning. @wozniakjan

wozniakjan: does it give up deprovisioning or just errors and requeues? I think unless you mark the operation as failed explicitly, it will continue to requeue it until it times out.

vvxxvvxx: It returns `return 0, err` here which means it will not be requeued (code https://github.com/kyma-project/control-plane/blob/fc5da85e9872f49a5e73910733503ee549112029/components/kyma-environment-broker/internal/process/queue.go#L87). so in this failure, I think the operation.State needs to be set to Failed, or to consider to retry (requeue) it? (need to r`eturn when(non-zero), nil` to requeue/retry)

components/kyma-environment-broker/internal/process/deprovisioning/manager.go
```diff
@@ -72,13 +80,26 @@ func (m *Manager) Execute(operationID string) (time.Duration, error) {
...
+		operation.LastError = kebError.ReasonForError(err)
+		log := logOperation.WithFields(logrus.Fields{"error_component": operation.LastError.Component(), "error_reason": operation.LastError.Reason()})
+		log.Errorf("Last error: %s", operation.LastError.Error())
+		m.publisher.Publish(context.TODO(), process.DeprovisioningStepProcessed{
```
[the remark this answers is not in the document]
vvxxvvxx: as here it returns `0, err` which terminates this deprovisioning, so I consider this is a failure which need to send the notification. As suggested above, if this err is changed to make a retry, then no need to introduce the event...

components/kyma-environment-broker/internal/process/deprovisioning/manager.go
wozniakjan: I discussed this with the authors and it looks like you may have discovered a bug.

@piotrmiskiewicz mentioned that this place should probably check if the error from `GetProvisioningOperationByInstanceID()` is recoverable or not. If it's for example network issue, operation should be requeued, but if DB returns not found, then not much we can do but fail the operation explicitly like it's done in

https://github.com/kyma-project/control-plane/blob/fc5da85e9872f49a5e73910733503ee549112029/components/kyma-environment-broker/internal/process/provisioning/check_cluster_configuration.go#L68

Thanks for pointing this out, feel free to leave it as is for now, it's probably outside of the scope of your PR to fix all bugs :) I will create an issue to go through the requeueing logic, I have a feeling there might be more places like this in KEB.

vvxxvvxx: yes, this is better to raise a separated PR for this fix :) (it's out of this PR's purpose, just wanted to point it out:)). I will leave this publish event here, but just this event here will not have the failed state result.

components/kyma-environment-broker/internal/process/deprovisioning/manager.go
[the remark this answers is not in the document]
wozniakjan: I was more pondering about the type of the event, whether to use `process.DeprovisioningStepProcessed{}` despite the fact the failure is not in any step yet, or introduce a new event, for example `process.DeprovisioningProcessed{}`  and publish that event in this case.

Currently, KEB registers events mainly for steps, one exception being `process.ProvisioningSucceeded{}`.

https://github.com/kyma-project/control-plane/blob/fc5da85e9872f49a5e73910733503ee549112029/components/kyma-environment-broker/internal/metrics/metrics.go#L17-L25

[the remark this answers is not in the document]
vvxxvvxx: for `OperationDurationCollector`, as there is no OldOperation here, so no data will be collected.
`StepResultCollector` is the one who uses step name as a label, in this case, the step name will be an empty "".
`OperationResultCollector` doesn't have step name label..

wozniakjan (approved):
/lgtm
/approve
/hold

I'm ok merging this, feel free to unhold at your convenience. It could make sense to get input from @ebensom or @piotrmiskiewicz regarding `DeprovisioningStepProcessed{}` vs `DeprovisioningProcessed{}` but should be fine even as is.

Thanks for explaining the details to me :)

PR #1527, opened 2022-04-05
Remove Gardener dependency from KEB

wozniakjan:
**Description**

Gardener is a rich resource of reported vulnerabilities by security scanners. Given KEB needs a very small part of Gardener API, it was considered to try copy/paste instead of vendoring the whole Gardened module. That spiralled into a large number of files https://github.com/kyma-project/control-plane/pull/1514. Instead, it was decided to change the code to use unstructured instead.

**Related issue(s)**

closes https://github.com/kyma-project/control-plane/issues/1471

PR #1600, opened 2022-04-20
Remove gardener for subscriptioncleanup job

wozniakjan:
**Description**

followup to https://github.com/kyma-project/control-plane/pull/1527, where KEB is stripping its dependency on gardener. Gardener is a rich source of CVEs There is only a very small subset of gardener API that KEB needs and it doesn't depend on the implementation at all.

**Related issue(s)**

https://github.com/kyma-project/control-plane/issues/1471

### DIFF
--- a/components/kyma-environment-broker/common/gardener/client.go
+++ b/components/kyma-environment-broker/common/gardener/client.go
@@ -8,8 +8,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-
-	gardener_apis "github.com/gardener/gardener/pkg/client/core/clientset/versioned/typed/core/v1beta1"
 )
 
 type SecretBinding struct {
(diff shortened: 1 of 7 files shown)


components/kyma-environment-broker/common/gardener/client.go
```diff
@@ -25,6 +23,15 @@ func (b SecretBinding) GetSecretRefName() string {
 	return str
 }
 
+func (b SecretBinding) GetSecretRefNamespace() string {
+	str, _, err := unstructured.NestedString(b.Unstructured.Object, "secretRef", "namespace")
+	if err != nil {
+		// NOTE this is a safety net, gardener v1beta1 API would need to break the contract for this to panic
+		panic(fmt.Sprintf("SecretBinding missing field '.secretRef.namespace': %v", err))
```
tillknuesting, line 30: Could you elaborate why a panic needs here to take place? Why not return an error or use log.Fatal?

components/kyma-environment-broker/go.mod
```diff
@@ -78,6 +77,7 @@ require (
 	github.com/evanphx/json-patch v5.6.0+incompatible // indirect
 	github.com/felixge/httpsnoop v1.0.2 // indirect
 	github.com/fsnotify/fsnotify v1.5.1 // indirect
+	github.com/gardener/gardener v1.44.4 // indirect
```
tillknuesting, line 80: indirect but still present :-)

wozniakjan: yeah, next step is provisioner, that should be the last one
```
$ go mod graph | grep ' github.com/gardener/gardener'
github.com/kyma-project/control-plane/components/kyma-environment-broker github.com/gardener/gardener@v1.44.4
github.com/kyma-project/control-plane/components/provisioner@v0.0.0-20220420073630-1208b41756a8 github.com/gardener/gardener@v1.24.0
```
or even better, we convince gardener to split their API into it's own module, which would drastically improve depending on it.

components/kyma-environment-broker/common/gardener/client.go
wozniakjan: returning an error would mean we will need to refactor in many places for a condition that can happen only if gardener changes their v1beta1 API to be incompatible with the current implementation, it was very briefly discussed outside of github and summarized here https://github.com/kyma-project/control-plane/pull/1527#discussion_r845069652

szwedm (approved):
/lgtm

szwedm (approved):
/lgtm

PR #1611, opened 2022-04-22
ERS CLI changes

piotrmiskiewicz:
**Description**

 - introduce new local directory with files per Kyma Service Instance that indicate that migration was called by us and it was successsful, if not keep false value. When CREATION_FAILED or DELETION_FAILED state then add kymaSkipped=true
- metadata command which joins data from the json file (ERS data) and calls KEB for instance info

**Related issue(s)**


tools/cli/pkg/ers/client/fake.go
```diff
@@ -46,6 +50,8 @@ func (f *Fake) GetPaged(pageStart, pageSize int) ([]ers.Instance, error) {
 }
 
 func (f *Fake) Migrate(instanceID string) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
```
piotrmiskiewicz, line 54: mutex was missing and running it was making unexpected errors

tools/cli/pkg/ers/command/migrate_all.go
```diff
@@ -98,17 +109,38 @@ func (c *MigrationAllCommand) Run() error {
...
-			c.stats.Add()
-		} else {
-			c.log.Infof("%sInstance %s already migrated%s",
-				Green, instance.Name, Reset)
+		meta, err := c.metadataStorage.Get(instance.Id)
+		if err != nil {
+			c.log.Warnf("Get %s Metadata error: %s", instance.Id, err.Error())
+			continue
+		}
+		if meta.KymaMigrated {
```
piotrmiskiewicz, line 117: do not run migration, if it was already done (information stored in local storage)

tools/cli/pkg/ers/metadata/storage.go
```diff
@@ -0,0 +1,66 @@
...
+func (s *Storage) ListAll() ([]ers.MigrationMetadata, error) {
+	result := []ers.MigrationMetadata{}
+	files, err := ioutil.ReadDir(metadataFolderName)
+	if err != nil {
+		return []ers.MigrationMetadata{}, nil
+	}
+	for _, file := range files {
+		data, err := ioutil.ReadFile(file.Name())
+		if err != nil {
+			continue
```
wozniakjan, line 54: nit: might be good to at least log an error to stderr? it should be pretty unlikely that here is ever any error but if it happens, it would be probably good to know about it

wozniakjan (commented):
minor nits about error propagation, I will defer to @ralikio for approval, he is much more familiar with this codebase

tools/cli/pkg/ers/metadata/storage.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	err = json.Unmarshal(data, &metadata)
+	return metadata, err
+}
+
+func (s *Storage) ListAll() ([]ers.MigrationMetadata, error) {
+	result := []ers.MigrationMetadata{}
+	files, err := ioutil.ReadDir(metadataFolderName)
+	if err != nil {
+		return []ers.MigrationMetadata{}, nil
```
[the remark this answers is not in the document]
piotrmiskiewicz: If the file does not exists - it means it was not created. It is not a corner case. It means , that such instance was not processed by the CLI and everything is ok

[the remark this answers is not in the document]
wozniakjan: > If the file does not exists - it means it was not created.

I agree, that was my intention by saying:

> down the stack, maybe just skip `os.IsNotExist()`

but there are other filesystem errors that may occur and in that case, it might be safer to abort the operation

szwedm (dismissed):
/lgtm

szwedm (approved):
/lgtm

PR #1633, opened 2022-04-26
Expose Detailed Metrics for Reconciler Audit Log Sidecar

jakobmoellersap:
**Description**

- Expose Fluent-Bit Sidecar Metrics of Reconciler in ServiceMonitor for Prometheus

**Related issue(s)**

- https://github.com/kyma-incubator/reconciler/issues/959


resources/kcp/charts/mothership-reconciler/templates/service.yaml
```diff
@@ -16,5 +16,9 @@ spec:
       targetPort: {{ .Values.service.istioPort }}
       protocol: TCP
       name: istio-http
+    - port: {{ .Values.fluentbit.port }}
+      targetPort: fluentbit
```
rakesh-garimella, line 20: I would prefer to call http-metrics so that it would make the purpose of the port clear

jakobmoellersap: I think http-metrics is misleading since reconciler also opens up metrics

jakobmoellersap: eg through http, we also expose the endpoint `/metrics`

rakesh-garimella (approved):

adityabhatia (approved):

PR #1685, opened 2022-05-12
Handle migration for suspended SKRs

piotrmiskiewicz:
**Description**

- When the migration update is done for suspended SKR, the unsuspension process must create a Kyma with btp operator

**Related issue(s)**


components/kyma-environment-broker/internal/broker/instance_update.go
```diff
@@ -293,15 +296,6 @@ func (b *UpdateEndpoint) processContext(instance *internal.Instance, details dom
 	}
 
 	if ersContext.IsMigration {
-		if k2, version, err := b.isKyma2(instance); err != nil {
```
piotrmiskiewicz, line 296: currently we are always working with kyma 2

PK85 (approved):

Issue #1556, opened 2022-04-11
Refactor KEB to not support 1.x Kyma related steps


PK85:
**Description**

Refactor KEB to not support 1.x Kyma related steps.

**Reasons**

Those steps are obsolete and not used in the KEB provisioning flows.

**AC**

- remove obsolete 1.x steps and if needed refactor existing ones to make provisioning, deprovisioning, update and upgrade flows clean
- reduce complexity of KEB flows and code, I expected steps code will be reduced by half

szwedm:
**Additional components collection step**

To enable the setting of additional components per Kyma version and plan we need a new approach to collect them for SKR. One solution might be declarative creation of ConfigMaps and kustomize. We already have similar working functionality for collecting runtime components overrides.

The idea
Add `additional-components.yaml` to `base` folder where overrides for specific runtime version are stored ([e.g. 2.1 base](https://github.tools.sap/kyma/management-plane-config/tree/master/resources/runtimes/2.1/base)).
Additional-components.yaml should contain ConfigMaps declarations with additional components set for plans.  
Example:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: additional-component-service-manager-proxy
  namespace: kcp-system
  labels:
    component: service-manager-proxy
    add-cmp-plan-azure: "true"
    add-cmp-plan-aws: "true"
    add-cmp-plan-openstack: "true"
    add-cmp-plan-gcp: "true"
    add-cmp-plan-azure_ha: "true"
    add-cmp-plan-aws_ha: "true"
    add-cmp-plan-preview: "true"
data:
  component.name: "service-manager-proxy"
  component.namespace: "kyma-system"

---
apiVersion: v1
kind: ConfigMap
metadata:
  name: additional-component-sc-migration
  namespace: kcp-system
  labels:
    component: sc-migration
    add-cmp-plan-azure: "true"
    add-cmp-plan-aws: "true"
    add-cmp-plan-gcp: "true"
data:
  component.name: "service-manager-proxy"
  component.namespace: "kyma-system"
  component.source: "https://github.com/kyma-incubator/sc-removal/releases/download/0.5.1/sap-btp-operator-migration-v0.5.1.tgz"
```
`kustomization.yaml` contains runtime versions for which prepared ConfigMaps should be relevant ([kustomization.yaml in base folder for 2.1](https://github.tools.sap/kyma/management-plane-config/blob/master/resources/runtimes/2.1/base/kustomization.yaml)).

[...]

PR #1686, opened 2022-05-12
New additional components provider

szwedm:
**Description**

- implement new components provider,
- deprecate v1alpha1.KymaComponent.

**Related issue(s)**
See #1556


wozniakjan (approved):
/lgtm
/approve

components/kyma-environment-broker/internal/runtime/components_provider.go
```diff
@@ -0,0 +1,346 @@
...
+	yamlContents, err := p.readYaml(p.defaultAdditionalComponentsYamlPath)
+	if err != nil {
+		return nil, err
+	}
+	return p.getComponentsFromYaml(yamlContents)
+}
+
+func (p *defaultAdditionalComponentsProvider) fetchAdditionalComponentsForGivenVersionAndPlan(kymaVersion internal.RuntimeVersionData, plan string) ([]KymaComponent, error) {
+	if plan == "" {
+		return nil, nil
```
wozniakjan, line 240: nit: how about logging this case?

components/kyma-environment-broker/internal/runtime/components_provider.go
```diff
@@ -0,0 +1,346 @@
...
+		return kebError.NewTemporaryError(msg)
+	case resp.StatusCode >= http.StatusInternalServerError:
+		return kebError.NewTemporaryError(msg)
+	default:
+		return errors.New(msg)
+	}
+}
+
+func (p *defaultAdditionalComponentsProvider) AdditionalComponents(kymaVersion internal.RuntimeVersionData, plan string) ([]KymaComponent, error) {
+	if kymaVersion.MajorVersion >= 2 {
```
piotrmiskiewicz, line 208: not necessary

components/kyma-environment-broker/internal/runtime/components_provider.go
szwedm: I like the idea, however we will need to pass a logger to the component provider or place it in the input builder factory. We will need to switch component providers sometime in the future, let's try to add logger then.

components/kyma-environment-broker/internal/runtime/components_provider.go
szwedm: Let's remove that condition when we switch component providers. It doesn't interfere with the flow for Kyma 2 as it is now and in case someone tries provisioning with Kyma 1 the error will be thrown at the beginning of the process.

piotrmiskiewicz (approved):

PR #1691, opened 2022-05-13
Create deprovision package

berkanercan:
**Description**

Authentication with oauth2 and access to the already existing delete runtime endpoint has been established.


components/kyma-environment-broker/common/deprovision/client.go
```diff
@@ -0,0 +1,63 @@
...
+	log    logrus.FieldLogger
+	URL    string
+	client *http.Client
+}
+
+func NewDeprovisionClient(parameters DeprovisionParameters) *DeprovisionClient {
+	cfg := clientcredentials.Config{
+		ClientID:     parameters.ClientID,
+		ClientSecret: parameters.ClientSecret,
+		TokenURL:     parameters.TokenURL,
```
vvxxvvxx, line 28: should this tokenURL a absolute url?

vvxxvvxx: just checking:), what is the difference between clientcredentials.Config.Client and oauth2.NewClient? i saw others using oauth2.NewClient

components/kyma-environment-broker/common/deprovision/client.go
```diff
@@ -0,0 +1,63 @@
...
+	request, err := http.NewRequest("DELETE", url, nil)
+	if err != nil {
+		return errors.Wrapf(err, "while creating the HTTP Delete request for deprovisioning")
+	}
+	request.Header.Set("X-Broker-API-Version", "2.14")
+
+	response, err := c.client.Do(request)
+	if err != nil {
+		return errors.Wrapf(err, "while calling %s", request.URL.String())
+	}
```
vvxxvvxx, line 56: After http client Do(request), need to Body.Close(). could refer to https://github.com/kyma-project/control-plane/blob/1cb304ac62839e6ae4b081c6eac94723e89ffd0b/components/kyma-environment-broker/common/orchestration/client.go#L293

components/kyma-environment-broker/common/deprovision/client.go
berkanercan: I wanted to use `clientcredentials` package, because I wanted to mimic the following command https://github.tools.sap/kyma/runtime-ops/blob/0d1ebb45b4fcb54f81367a31d6edd4176ac58da4/util.sh#L47. If I can do the same with `oauth2`, it makes sense to switch to that.

berkanercan: tokenURL should be absolute URL yes. It should be provided by the kcp user in the config file ```"$HOME/.kcp/config-dev.yaml"```

PR #1700, opened 2022-05-16
Comparing modifiedAt before and after migration and finish earlier wh…

PK85:
…en CIS has error and not called KEB at all

**Description**

- Comparing modifiedAt before and after migration and finish earlier when CIS has error and not called KEB at all
- ...
- ...

**Related issue(s)**

### DIFF
--- a/tools/cli/pkg/ers/command/migrate_all.go
+++ b/tools/cli/pkg/ers/command/migrate_all.go
@@ -24,11 +24,11 @@ func NewMigrationAllCommand(log logger.Logger) *cobra.Command {
 	}
 
 	cobraCmd := &cobra.Command{
-		Use:     `migrate-all`,
-		Short:   `Triggers full SC migration accepting json objects as input.`,
-		Long:    `Migrates all instances that are feed through stdin in the form of json objects`,
+		Use:   `migrate-all`,
+		Short: `Triggers full SC migration accepting json objects as input.`,
+		Long:  `Migrates all instances that are feed through stdin in the form of json objects`,
 		Example: `  ers migrate -w2 -d	Triggers migration starting two workers`,
-		Args:    cobra.MaximumNArgs(1),
+		Args: cobra.MaximumNArgs(1),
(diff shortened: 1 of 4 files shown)


ralikio (approved):

szwedm (dismissed):
/lgtm

tools/cli/pkg/ers/model.go
```diff
@@ -26,7 +26,7 @@ type Instance struct {
 	Labels          string
 	LandscapeLabel  string
 	// TODO: Date comes in different format than specified
-	// ModifiedDate      date.Date //2021-10-04T112314.137Z
+	ModifiedDate   int //2021-10-04T112314.137Z
```
tillknuesting, line 29:
```suggestion
	ModifiedDate   int64 //2021-10-04T112314.137Z
```
However I would suggest in general to use [time.Time](https://pkg.go.dev/time#Time) type that we have in the standard library for timestamps. However `2021-10-04T112314.137Z` looks malformed to me as to `RFC3339` it should be "2021-10-04T11:23:14.137Z"?

Example how to parse timestamp string:

```
package main

import (
	"fmt"
	"time"
)

type Instance struct {
	ModifiedDate time.Time
}

func main() {
	timeStampString := "2021-10-04T11:23:14.137Z"

	timeStamp, err := time.Parse(time.RFC3339, timeStampString)
	if err != nil{
		fmt.Println(err)
		return
	}

	instance := &Instance{}

	instance.ModifiedDate = timeStamp

	fmt.Println(instance.ModifiedDate)
} 
```

tools/cli/pkg/ers/command/migrate_all.go
```diff
@@ -107,6 +107,7 @@ func (c *MigrationAllCommand) Run() error {
 
 	for _, instance := range instances {
```
tillknuesting, line 109:
```suggestion
```

szwedm (dismissed):
/lgtm

tillknuesting (dismissed):

szwedm (approved):
/lgtm

PR #1729, opened 2022-05-25
Fix GetKubeconfig step in deprovisioning queue

szwedm:
**Description**

- add not found error check,
- add tests for step.

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/process/deprovisioning/get_kubeconfig_step.go
+++ b/components/kyma-environment-broker/internal/process/deprovisioning/get_kubeconfig_step.go
@@ -1,6 +1,7 @@
 package deprovisioning
 
 import (
+	"strings"
 	"time"
 
 	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal"
@@ -53,6 +54,11 @@ func (s *GetKubeconfigStep) Run(operation internal.DeprovisioningOperation, log 
 
 	status, err := s.provisionerClient.RuntimeStatus(operation.ProvisioningParameters.ErsContext.GlobalAccountID, operation.RuntimeID)
 	if err != nil {
+		if s.isNotFoundErr(err) {
+			log.Infof("shoot not found, skipping step")
+			operation.IsServiceInstanceDeleted = true
+			return operation, 0, nil
+		}
 		return handleError(s.Name(), operation, err, log, "call to provisioner RuntimeStatus failed")
 	}
 
@@ -72,4 +78,8 @@ func (s *GetKubeconfigStep) Run(operation internal.DeprovisioningOperation, log 
 	operation.K8sClient = cli
 
 	return operation, 0, nil
+}
+
+func (s *GetKubeconfigStep) isNotFoundErr(err error) bool {
+	return strings.Contains(err.Error(), "not found")
 }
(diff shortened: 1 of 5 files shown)


components/kyma-environment-broker/internal/process/deprovisioning/get_kubeconfig_step.go
```diff
@@ -73,3 +79,7 @@ func (s *GetKubeconfigStep) Run(operation internal.DeprovisioningOperation, log
 
 	return operation, 0, nil
 }
+
+func (s *GetKubeconfigStep) isNotFoundErr(err error) bool {
+	return strings.Contains(err.Error(), "not found")
```
wozniakjan, line 84: nit: is there by any chance an underlying type for the error in provisioner that we can cast on like it's done by apimachinery for kube-api errors?
https://github.com/kubernetes/apimachinery/blob/e3c6631c282ae532562559ff7b2ae93b1687a1ab/pkg/api/errors/errors.go#L525-L536

if there isn't then we probably keep it as is, it's just that I find the type check to be a little bit more robust against changes, e.g. someone can decide the message to become "Not found" or "isn't found" in the future which may break this code.

wozniakjan (dismissed):
/lgtm
/approve
/hold

feel free to unhold at your convenience if there is no way to type check on the error from provisioner.

components/kyma-environment-broker/internal/process/deprovisioning/get_kubeconfig_step.go
szwedm: at first I wanted to cast the error returned by provisioner to something identifiable, exactly like it is with k8s client and apimachinery errors, but it is not possible in this case. Provisioner transforms apimachinery errors to its own kind of error: 
https://github.com/kyma-project/control-plane/blob/main/components/provisioner/internal/gardener/provisioner.go#L206
https://github.com/kyma-project/control-plane/blob/main/components/provisioner/internal/util/errors.go#L26
All we get is error message in graphql.

tillknuesting: Would suggest to use https://pkg.go.dev/errors#Is to check for certain type of errors to support wrapped errors.

szwedm:
The deprovisioning operation was reaching 7h timeout, because this condition was always `true`:
https://github.com/kyma-project/control-plane/pull/1729/files#diff-db56336d92be215aaa1d4b69d9cd338aedd82c2f2cd5e623e6565142851fee0aL16

We modify `UpdatedAt` field right after deprovisioning begins:
https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/internal/process/deprovisioning/initialisation.go#L107
https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/internal/storage/driver/postsql/operation.go#L145

components/kyma-environment-broker/internal/process/deprovisioning/get_kubeconfig_step.go
tillknuesting: Since is out of scope to refactor the all other places I think its fine to it this way. But just in general in the codebase the errors are mostly handled unidiomatic.

szwedm: I changed the condition for repeating the step, it shouldn't run for more than 30 mins from the operation creation time, even when we miss "not found" error.

szwedm:
Deprovisioning will continue even when k8s client is not provided - case when Provisioner returns error when fetching Runtime Status. This way KEB will try to remove as much as it can during deprovisioning operation without failing the whole process.

wozniakjan (approved):
/lgtm
/approve

PR #1736, opened 2022-05-30
[testing]feature: kcp runtime functions for kubeconfig-serivce

AdrianPei:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/components/kubeconfig-service/pkg/endpoints/endpoint.go
+++ b/components/kubeconfig-service/pkg/endpoints/endpoint.go
@@ -3,6 +3,7 @@ package endpoints
 import (
 	"errors"
 	"net/http"
+	"sync"
 
 	"github.com/gorilla/mux"
 	authn "github.com/kyma-project/control-plane/components/kubeconfig-service/pkg/authn"
@@ -16,6 +17,9 @@ const (
 	mimeTypeYaml = "application/x-yaml"
 	mimeTypeText = "text/plain"
 )
+
+//mutex to avoid critical section in config map deployment
+var mu sync.Mutex
 
 //EndpointClient Wrpper for Endpoints
 type EndpointClient struct {
@@ -87,7 +91,7 @@ func (ec EndpointClient) generateKubeConfig(tenant, runtime string, userInfo aut
 		return nil, err
 	}
 
-	runtimeClient, err := run.NewRuntimeClient([]byte(rawConfig), userInfo.ID, userInfo.Role)
+	runtimeClient, err := run.NewRuntimeClient([]byte(rawConfig), userInfo.ID, userInfo.Role, tenant)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


components/kubeconfig-service/pkg/runtime/runtime_kcp.go
```diff
@@ -0,0 +1,227 @@
...
+		return errors.Errorf("Failed to marshal patch, %s", err.Error())
+	}
+	_, err = rtc.KcpK8s.CoreV1().ConfigMaps(KcpNamespace).Patch(context.Background(), userID, types.JSONPatchType, payload, metav1.PatchOptions{})
+	if err != nil {
+		return errors.Errorf("Failed to update config map, %s", err.Error())
+	}
+	return nil
+}
+
+func (rtc *RuntimeClient) DeployConfigMap(runtimeID string, L2L3OperatiorRole string) error {
```
ebensom, line 156:
```suggestion
func (rtc *RuntimeClient) DeployConfigMap(runtimeID string, L2L3OperatorRole string) error {
```

components/kubeconfig-service/pkg/runtime/runtime_kcp.go
```diff
@@ -0,0 +1,227 @@
...
+	userID := rtc.User.ServiceAccountName
+	tenantID := rtc.User.TenantID
+
+	log.Info("Checking if the user exists")
+	startTime := time.Now().String()
+	cm, err := rtc.KcpK8s.CoreV1().ConfigMaps(KcpNamespace).Get(context.Background(), userID, metav1.GetOptions{})
+	if k8serrors.IsNotFound(err) {
```
ebensom, line 163: I think the (get + create) is a critical section where multiple concurrent goroutines (HTTP handler routines calling the endpoint handler) could in theory race for reading this configmap of the same user and creating the CM when it doesn't exist. In practice however it should not cause issues as it's doubdtful if there will be multiple concurrent kubeconfig requests in flight for the same user. Even `kcp taskrun` command queries kubeconfigs sequentially within a loop.

All further operations (addition, deletion during expiry) are executed as individual PATCH K8S API requests, which is ensured to be serialized by the K8S API so won't cause problems .

AdrianPei: add a global channel to avoid critical section

components/kubeconfig-service/pkg/endpoints/endpoint.go
```diff
@@ -102,5 +102,15 @@ func (ec EndpointClient) generateKubeConfig(tenant, runtime string, userInfo aut
 		return nil, err
 	}
 
+	run.CmChannel <- true
```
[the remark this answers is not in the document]
AdrianPei: replaced channel with mutex as suggested

ebensom (approved):

PR #1738, opened 2022-05-30
Exclude canceled and retrying operation states when treating last operation

ebensom:
**Description**

Issue: [KCP CLI runtime state flag not filtering all Runtimes in Succeeded state.](https://github.tools.sap/kyma/backlog/issues/2570)

[LastOperation](https://github.com/kyma-project/control-plane/blob/0f7ca11deae009a120ac1f5ca583779b31c03d81/components/kyma-environment-broker/internal/storage/postsql/read.go#L105) doesn't exclude the canceled state which is valid for orchestrated operations . So as a result theses instances will be excluded during inner join as part of the state query, but the runtimes endpoint will set succeeded state for them


ebensom:
> We cannot exclude pending operation, for example, if the update comes from a platform (ERS), the operations is created with a status Pending. Then, if any other operation for that instance is "in progress", then that update is still in pending. From OSB perspective, that operation is in progress!

Hm, he pending operation was excluded already when searching for the last operation. I only added canceled and retrying states. 
So you are saying the original code is wrong in the sense that `GetLastOperation` should also include pending ones?

piotrmiskiewicz:
The main problem is that the GetLastOperation is executed to get the lastest instance details:
https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/internal/storage/driver/postsql/instance.go#L326

this is to copy instance details to new pending operation.

PR #1746, opened 2022-06-02
Disable ShootNetworkingFilter by default

franpog859:
**Description**

- Disable ShootNetworkingFilter by default so if there is no value or `nil` provided while provisioning, it'll be interpreted as `true`. The upgrade process does not change - no value or `nil` provided -> nothing happens, `true` provided -> the extension is being disabled, `false` -> the extension is being enabled

How to disable ShootNetworkingFilter on the existing shoot? Run the `upgradeShoot` mutation for this runtime with the `shootNetworkingFilterDisabled: true` provided:
```graphql
mutation {
  upgradeShoot(
    id: "SOME_RUNTIME_ID"
    config: {
      gardenerConfig: {
        shootNetworkingFilterDisabled: true
...
```

You can always check the actual `ShootNetworkingFilter` status by running the `runtimeStatus` query:
```graphql
query {
  runtimeStatus(id: "SOME_RUNTIME_ID") {
    runtimeConfiguration {
      clusterConfig {
        shootNetworkingFilterDisabled
...
```
You can also change it by manually modifying the Gardener Shoot. The value will be updated in the Provisioner's database after the next upgrade operation


wozniakjan (commented):
> * Disable ShootNetworkingFilter by default so if there is no value or `nil` provided while provisioning, it'll be interpreted as `true`.

this feels a little counterintuitive to me that users need to pass `true` in order to disable something, do I understand this correctly that
* `DisableShootNetworkingFilter == *true` - disabled
* `DisableShootNetworkingFilter == *false` - enabled
* `DisableShootNetworkingFilter == nil` - enabled

I think you don't need a pointer because the absence of the value (`nil`) means the same thing as the default value (`false`) for `bool` in go in this case.

But have you considered making this `EnableShootNetworkingFilter` instead? It will be also disabled by default, not need a pointer, but will be a little more intuitive to use imho, you will need to pass `true` to enable something, and if you don't, it's disabled by default just like `bool` in go is false by default.

franpog859:
@wozniakjan , thanks for the comment! I kept it this way (`disable` and not `enable`) because it's the literal equivalent of the Gardener spec. I didn't want to confuse any SRE person that it's some difference there

I also kept the pointer for two reasons. The first one is similar, it's a literal equivalent of the Gardener spec. The second, more important one, is that KEB doesn't provide this value yet so it just must be optional

franpog859:
I tested manually two scenarios:
1. Provision an SKR using the old Provisioner without `shootNetworkingFilterDisabled` provided - should be `null` in `runtimeStatus` and nothing in the Shoot - update the extension manually in the Shoot, set it to `true` - `runtimeStatus` should still give `null` - update Privisoner's image, run `upgradeShoot` with `nil` provided - it should be still `true` in the Shoot but also `true` in the `runtimeStatus` this time - update Shoot manually, changing the extension to `false` - run `upgradeRuntime` - there should be `false` in both Shoot and `runtimeStatus`
1. Provision an SKR using the new Provisioner without `shootNetworkingFilterDisabled` provided - `runtimeStatus` should give `true` and the Shoot should have `true` in the extension - run the `upgradeShoot` with `false` provided - Shoot extension and the `runtimeStatus` output should change - run the `upgradeShoot` without anything provided - nothing should change - run the `upgradeRuntime` - again nothing should change

akgalwas (approved):

PR #1757, opened 2022-06-07
Feature/upgrade in future maintenance window

0xE282B0:
**Description**
Schedule upgrades in future maintenance window

- Adds `--scheduleAfter` parameter to cli
- KEB can schedule upgrades to maintenance windows later than the next one

**Related issue(s)**

[GHE/backlog#2480](https://github.tools.sap/kyma/backlog/issues/2480)

### DIFF
--- a/components/kyma-environment-broker/common/orchestration/dto.go
+++ b/components/kyma-environment-broker/common/orchestration/dto.go
@@ -103,9 +103,10 @@ type ParallelStrategySpec struct {
 
 // StrategySpec is the strategy part common for all orchestration trigger/status API
 type StrategySpec struct {
-	Type     StrategyType         `json:"type"`
-	Schedule ScheduleType         `json:"schedule,omitempty"`
-	Parallel ParallelStrategySpec `json:"parallel,omitempty"`
+	Type          StrategyType         `json:"type"`
+	Schedule      ScheduleType         `json:"schedule,omitempty"`
+	ScheduleAfter time.Time            `json:"scheduleAfter,omitempty"`
+	Parallel      ParallelStrategySpec `json:"parallel,omitempty"`
 }
 
 // TargetSpec is the targets part common for all orchestration trigger/status API
--- a/components/kyma-environment-broker/common/orchestration/dto.go
+++ b/components/kyma-environment-broker/common/orchestration/dto.go
@@ -93,6 +93,7 @@ type ScheduleType string
 
 const (
 	Immediate         ScheduleType = "immediate"
+	Now               ScheduleType = "now"
 	MaintenanceWindow ScheduleType = "maintenanceWindow"
 )
 
@@ -103,10 +104,11 @@ type ParallelStrategySpec struct {
 
 // StrategySpec is the strategy part common for all orchestration trigger/status API
 type StrategySpec struct {
-	Type          StrategyType         `json:"type"`
-	Schedule      ScheduleType         `json:"schedule,omitempty"`
-	ScheduleAfter time.Time            `json:"scheduleAfter,omitempty"`
-	Parallel      ParallelStrategySpec `json:"parallel,omitempty"`
+	Type              StrategyType `json:"type"`
+	Schedule          string       `json:"schedule,omitempty"`
+	ScheduleTime      time.Time
+	MaintenanceWindow bool                 `json:"maintenanceWindow,omitempty"`
+	Parallel          ParallelStrategySpec `json:"parallel,omitempty"`
 }
 
 // TargetSpec is the targets part common for all orchestration trigger/status API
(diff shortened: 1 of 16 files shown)


components/kyma-environment-broker/common/orchestration/dto.go
```diff
@@ -103,9 +103,10 @@ type ParallelStrategySpec struct {
 
 // StrategySpec is the strategy part common for all orchestration trigger/status API
 type StrategySpec struct {
-	Type     StrategyType         `json:"type"`
-	Schedule ScheduleType         `json:"schedule,omitempty"`
-	Parallel ParallelStrategySpec `json:"parallel,omitempty"`
+	Type          StrategyType         `json:"type"`
+	Schedule      ScheduleType         `json:"schedule,omitempty"`
```
ebensom, line 107: I'm thinking of the following (partially backwards incompatible) API change: 

- `MaintenanceWindow` should be a bool parameter, it can be omitted (default false).
- `Schedule` can be of value `immediate` (to keep backwards compatibility), `now` , or a specific parsable timestamp format.
- When neither the above parameters are specified, or only schedule=immediate/now is specified, the current default behavior is executed, i.e. all operations are executed immediately by the parallel strategy
- When schedule=maintenanceWindow is specified (using older version of KCP CLI), the request is rejected with 400 Bad request
- When MaintenanceWindow is set to true, the initial maintenance windows are resolved for each operation. If `Schedule` is set to a valid time, operations are scheduled after that timestamp. The reschedule logic in parallel strategy for maintenance window operations will also need to be readjusted. 
- When MaintenanceWindow is set to false and `Schedule` is set to a valid time, the maintenance window begin/end are not resolved, but the parallel strategy will enqueue all operations with .`AddAfter(Schedule)` 

The KCP CLI should be adjusted to accommodate these param changes: the `--schedule` should accept `immediate`, `now` or a valid timestamp parable into `time.Time`.

components/kyma-environment-broker/common/orchestration/strategies/parallel.go
```diff
@@ -221,8 +220,9 @@ func (p *ParallelOrchestrationStrategy) updateMaintenanceWindow(execID string, o
 		}
 
 		duration = time.Until(op.MaintenanceWindowBegin)
-
-	case orchestration.Immediate:
+	} else {
+		p.executor.Reschedule(id, strategy.ScheduleTime, strategy.ScheduleTime)
+		duration = time.Until(strategy.ScheduleTime)
```
ebensom, line 225: It does work well when `strategy.ScheduleTime.IsZero()`, right? So the duration is <= 0 in that case (could be checked on https://go.dev/play/)

tools/cli/pkg/command/upgrade.go
```diff
@@ -57,11 +62,24 @@ func (cmd *UpgradeCommand) ValidateTransformUpgradeOpts() error {
...
+		}
+		cmd.orchestrationParams.Strategy.Schedule = cmd.schedule
+	} else if _, err := time.Parse(time.RFC3339, cmd.schedule); err == nil {
+		cmd.orchestrationParams.Strategy.Schedule = cmd.schedule
 	} else {
 		return fmt.Errorf("invalid value for schedule: %s. Check kcp upgrade --help for more information", cmd.schedule)
 	}
 
+	// Validate schedule
+	if scheduleAfterParam, err := time.Parse("2006-12-31", string(cmd.scheduleAfter)); err != nil {
```
ebensom, line 77: I think we won't need to to attributes for parsing a schedule time. 
How about just the `orchestrationParams.Strategy.Schedule` is kept, and the `orchestrationParams.Strategy.ScheduleAfter` is dropped? And then for `orchestrationParams.Strategy.Schedule` we would support the following valid input:

- "now"
- "immediate"
- time.Parse("2006-12-31", ...
- time.Parse(time.RFC3339, ...

Also, in handler code and orchestration manager, we would just deal with this one attribute.

components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -262,7 +262,7 @@ func (m *orchestrationManager) waitForCompletion(o *internal.Orchestration, stra
 			// handle the retrying ops during in progress orchestration
 			// use the existing resolved policy in op
 			numberOfNotFinished += numberOfRetrying
-			ops, err := m.factory.RetryOperations(o.OrchestrationID, o.Parameters.Strategy.Schedule, orchestration.MaintenancePolicy{}, false)
+			ops, err := m.factory.RetryOperations(o.OrchestrationID, o.Parameters.Strategy.ScheduleTime, o.Parameters.Strategy.MaintenanceWindow, orchestration.MaintenancePolicy{}, false)
```
ebensom, line 265: I would change the `factory.RetryOperations` interface to get the whole `o.Parameters` or `o.Parameters.Strategy` as one input instead of adding a new parameter

PR #1771, opened 2022-06-10
Ignored Error

ralikio:
**Description**

- Correction for ignored error.

**Related issue(s)**


tools/cli/pkg/ers/fetcher/initial.go
```diff
@@ -34,6 +34,9 @@ func (e InitialFetcher) GetAllInstances() ([]ers.Instance, error) {
 
 		output = append(output, instances...)
 		instances, err = e.client.GetPaged(page, e.pageSize)
+		if err != nil {
+			break
```
wozniakjan, line 38: maybe log the error at least? or aggregate and return all at the end?

https://github.com/kubernetes/apimachinery/blob/d407afb45eb78bce52710653e5f6c99e0d3493a9/pkg/util/errors/errors.go#L46

PR #1815, opened 2022-06-24
provisioner: add support for Azure multi AZ, having dedicated subnet per zone

ebensom:
**Description**

- `.spec.networking.nodes` is configured provider specific way, as per suggestion from Gardener documentation. For Azure, it's the same as the VNET CIDR. For AWS, it's the VPC CIDR. For GCP, it's the Worker CIDR from GardenerConfig, which is used to configure the regional VPC subnet.
- `GardenerConfigInput` is extended to support specifying multiple Azure zones, using the [dedicated subnet per zone](https://github.com/gardener/gardener-extension-provider-azure/blob/master/docs/usage-as-end-user.md#infrastructureconfig-with-dedicated-subnets-per-zone) config which also enables zone redundant NAT Gateway. API change is backwards compatible with the current model, so both the old config input (single AZ zoned cluster) and the new config input (multi-AZ zoned cluster) is supported.
- Gardener config model code was adjusted to support both current and new (multi-AZ with subnets per zone) Azure configuration scheme

**Related issue(s)**

### DIFF
--- a/components/provisioner/internal/model/gardener_config.go
+++ b/components/provisioner/internal/model/gardener_config.go
@@ -175,8 +175,8 @@ func (c GardenerConfig) ToShootTemplate(namespace string, accountId string, subA
 				},
 			},
 			Networking: gardener_types.Networking{
-				Type:  "calico",                        // Default value - we may consider adding it to API (if Hydroform will support it)
-				Nodes: util.StringPtr("10.250.0.0/19"), // TODO: it is required - provide configuration in API (when Hydroform will support it)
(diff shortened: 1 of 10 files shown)


koala7659:
When testing the PR I have noticed that cidr string for each azure zone can be edited in the provisioner database during Shoot Upgrade operation. In the same time Gardener rejects this change and reports this field as immutable. 

I think we should not allow for that this - just prevent this immutable field in the database form update.

koala7659 (approved):
LGTM

PR #1819, opened 2022-06-27
Added Missing Operation Status Mapping

ralikio:
**Description**

- Because of missing mapping we were publishing default result value, which was float's `0`. Added missing operation mappings not to send default 0 as failure when err is nil.

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/metrics/operation_result.go
+++ b/components/kyma-environment-broker/internal/metrics/operation_result.go
@@ -100,23 +100,7 @@ func (c *OperationResultCollector) OnUpgradeKymaStepProcessed(ctx context.Contex
 		return fmt.Errorf("expected UpgradeKymaStepProcessed but got %+v", ev)
 	}
 
-	var resultValue float64
-	switch stepProcessed.Operation.State {
-	case domain.InProgress:
-		resultValue = resultInProgress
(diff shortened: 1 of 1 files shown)


components/kyma-environment-broker/internal/metrics/operation_result.go
```diff
@@ -180,11 +180,13 @@ func (c *OperationResultCollector) OnProvisioningStepProcessed(ctx context.Conte
 
 	var resultValue float64
 	switch stepProcessed.Operation.State {
-	case domain.InProgress:
+	case domain.InProgress, Pending, Retrying:
```
wozniakjan, line 183: I would dare to say that `InProgress` != `Pending`, consider expanding the switch statement to use the full range
https://github.com/kyma-project/control-plane/blob/8a652e3dadfc0b41c96e0ffa2470f8681a4d468c/components/kyma-environment-broker/internal/metrics/operation_result.go#L16-L23
like it's done in 
https://github.com/kyma-project/control-plane/blob/8a652e3dadfc0b41c96e0ffa2470f8681a4d468c/components/kyma-environment-broker/internal/metrics/operation_result.go#L136-L150

piotrmiskiewicz (approved):

PR #1827, opened 2022-07-01
Allow authenticated access to KEB /runtimes to specified principals

ebensom:
**Description**

Allow authenticated access to KEB /runtimes to specified principals. 
KMC and mop-agent will have to be allowed to use the API.

**Related issue(s)**


szwedm (approved):
/lgtm

PR #1837, opened 2022-07-06
provisioner: fix azureZones input to be optional

ebensom:
**Description**

The new `azureZones` field in `AzureProviderConfigInput` input parameter was meant to be optional, to keep backwards compatibility with the current API (providing `zones` list only). 

**Related issue(s)**


koala7659 (approved):
LGTM

PR #1840, opened 2022-07-06
Created AWS cleanup script

Jani19966:
**Description**

- There was no AWS cleanup script. If the customer creates their own storageclasses with retain policy, the volume won't be deleted even after deprovisioning the cluster.
- The AWS script gets AWS account credentials and deletes (detaches if needed) all volumes

**Related issue(s)**


components/kyma-environment-broker/cmd/subscriptioncleanup/cloudprovider/aws.go
```diff
@@ -1,18 +1,126 @@
...
+		// Iterate through volumes
+		volumes, err := ac.ec2Client.DescribeVolumes(context.TODO(), &ec2.DescribeVolumesInput{})
+		if err != nil {
+			return err
+		}
+
+		log.Println("Listing Volumes:")
+		for _, volume := range volumes.Volumes {
+			log.Printf("Volume ID: %v", *volume.VolumeId)
+		}
```
ebensom, line 61: if you print out all volume IDs to be deleted inside deleteVolumes(), the call to `DescribeVolumes()` and the loop to print the ID seems redundant

components/kyma-environment-broker/cmd/subscriptioncleanup/cloudprovider/aws.go
```diff
@@ -1,18 +1,126 @@
...
+	}
+
+	for _, volume := range volumes.Volumes {
+		log.Printf("Deleting volume with id: %v", *volume.VolumeId)
+		if volume.State == types.VolumeStateInUse {
+			log.Printf("Volume is in-use, detaching it first: ")
+			ac.ec2Client.DetachVolume(context.TODO(), &ec2.DetachVolumeInput{
+				VolumeId: volume.VolumeId,
+			})
+		}
```
ebensom, line 82: I am thinking if we should error out in this case. A HAP account marked as dirty, which has volumes attached to VM instance? To be on the safe side I would not wipe in this case. 
Erroring out will allow us to notice as alert will fire due to K8S job being failed.

components/kyma-environment-broker/cmd/subscriptioncleanup/cloudprovider/aws.go
```diff
@@ -1,18 +1,126 @@
...
 func NewAwsResourcesCleaner(secretData map[string][]byte) (ResourceCleaner, error) {
-	return awsResourceCleaner{}, nil
+	awsResourceCleaner := awsResourceCleaner{}
+	awsConfig, err := awsResourceCleaner.toAwsConfig(secretData)
+	if err != nil {
+		return nil, err
+	}
+
+	awsResourceCleaner.credentials = awsConfig
+	awsEC2Client, err := awsResourceCleaner.newAwsEC2Client(awsConfig, defaultRegion)
```
ebensom, line 34: I think since the client is per-region, the cleaner struct shouldn't have a (global) attribute for it, and it's nto necessary to initialize a client with a "default" region, which is never being used during code execution.

I would just-in-time initialize the client when looping through the regions below, and pass the client as a parameter of  `deleteVolumes(client)`

components/kyma-environment-broker/cmd/subscriptioncleanup/cloudprovider/aws_test.go
```diff
@@ -0,0 +1,26 @@
+package cloudprovider_test
+
+import (
+	"testing"
+
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/cmd/subscriptioncleanup/cloudprovider"
+)
+
+func TestGetEC2Instances(t *testing.T) {
+	// Before running the test, please search for a development aws account credentials
```
ebensom, line 10: Unit tests stored together with source code should never target real systems. In this case the unit test could work with a mock client, which responds back with some fake generated data, useful for testing out the cleaner code logic.

- I suggest you to create an interface for the  EC2 Client functions you use in your code, and a Factory interaface which can produce instance of an EC2 Client interface implementation for a given region. 
- Your cleaner struct will be then initialized with an instance of the Factory interface. 
- In the real implementation, you implement the Factory interface like it can be constructed with AWS credentials, and then the `NewClient(region string)` method will initialize and yield a real *ec2.Client, which is compatible with your Client interface.
- In unit test code, you implement a mock factory, whose `NewClient(region string)` yields a mock client implementation, so just generating responses useful for to unit test your code.

On example (from AWS :D) about a simplified version of this technique (without factory, but you will need because you have client per region):
https://aws.amazon.com/blogs/developer/mocking-out-then-aws-sdk-for-go-for-unit-testing/

components/kyma-environment-broker/go.mod
```diff
@@ -60,6 +69,8 @@ require (
 	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
 	github.com/agnivade/levenshtein v1.1.1 // indirect
 	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d // indirect
+	github.com/aws/aws-sdk-go-v2/credentials v1.12.8
```
ebensom, line 72: These should be moved up in the top require section, among the direct imports.

README.md
```diff
@@ -1,5 +1,4 @@
 # Kyma Control Plane
-
```
ebensom, line 2: I suggest adding back this line to avoid approval need from TWs

ebensom (approved):

ebensom (commented):
Please don't forget to bump the subscription cleanup job image:
https://github.com/kyma-project/control-plane/blob/main/resources/kcp/values.yaml#L17

ebensom (approved):

PR #1854, opened 2022-07-11
Elimination of Bash from Schema Migrator

ralikio:
**Description**

- Refactoring from bash script in schema migrator to more optimised go program.

**Related issue(s)**


components/schema-migrator/main.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	dbName := os.Getenv("DB_NAME")
+
+	_, present := os.LookupEnv("DB_SSL")
+	if present {
+		dbName = fmt.Sprintf("%s?sslmode=%s", dbName, os.Getenv("DB_SSL"))
+	}
+
+	connectionString := fmt.Sprintf(
```
tillknuesting, line 44:
```suggestion
        // IPv6 not supported in construction of network address.
	connectionString := fmt.Sprintf(
```

components/schema-migrator/main.go
```diff
@@ -0,0 +1,106 @@
...
+		_, present := os.LookupEnv(env)
+		if !present {
+			fmt.Printf("ERROR: %s is not set\n", env)
+			os.Exit(1)
+		}
+	}
+
+	direction := os.Getenv("DIRECTION")
+
+	if direction == "up" {
```
tillknuesting, line 28: I would prefer to use a switch here
```
switch direction {
	case "up":
		fmt.Println("Migration UP")
	case "down":
		fmt.Println("Migration DOWN")
	default:
		fmt.Println("ERROR: DIRECTION variable accepts only two values: up or down")
		os.Exit(1)
	}
```

components/schema-migrator/Dockerfile
```diff
@@ -1,28 +1,23 @@
...
 
-WORKDIR /migrate
+COPY go.mod /program/go.mod
+COPY main.go /program/main.go
 
-RUN apk --no-cache add --update git make
-RUN git clone https://github.com/golang-migrate/migrate.git .
+WORKDIR /program
+RUN go mod tidy && go mod vendor && \
+    CGO_ENABLED=0 go build -o /bin/program /program/main.go
```
tillknuesting, line 8: Stripping would it make it even smaller :-) 
```suggestion
    CGO_ENABLED=0 go build -ldflags="-s -w" -o /bin/program /program/main.go
```

components/schema-migrator/main.go
```diff
@@ -0,0 +1,106 @@
...
+		"postgres", driver)
+	if err != nil {
+		fmt.Printf("Error during migration initialization, %s\n", err)
+	}
+
+	defer m.Close()
+	m.Log = &Logger{}
+
+	if direction == "up" {
+		m.Up()
```
tillknuesting, line 89: I would log a here a possible error value returned by Up()

components/schema-migrator/Dockerfile
wozniakjan: But then you lose convenience of continuous profiling tools based on ebpf and similar technologies..

tillknuesting (dismissed):

tillknuesting (approved):

PR #1865, opened 2022-07-14
Generic Read Operation

ralikio:
**Description**

- POC for generic read operation.

**Related issue(s)**


components/kyma-environment-broker/internal/storage/postsql/generic_read.go
```diff
@@ -0,0 +1,33 @@
...
+
+type GenericRead[T any] struct {
+	Session   *dbr.Session
+	IdName    string
+	TableName string
+
+	NewItem func() T
+}
+
+func (r GenericRead[T]) GetInstanceByID(instanceID string) (T, dberr.Error) {
```
piotrmiskiewicz, line 16: the name of that method is not generic, it should be something like GetByID

components/kyma-environment-broker/internal/storage/postsql/generic_read_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+		session := connection.NewSession(nil)
+		defer session.Close()
+
+		tx, err := session.Begin()
+		assert.NoError(t, err)
+
+		testServiceName := "Test Service Name"
+
+		_, err = session.InsertInto(postsql.InstancesTableName).
```
piotrmiskiewicz, line 52: in my opinion, the PR should be about Read and Write operation so in the unit test we can test "write" and "read" operation then compare objects if are equals

PR #1875, opened 2022-07-18
Staged manager uses plain Operations instead of ProvisioningOperations

jaroslaw-pieszka:
**Description**

Changes proposed in this pull request:

- added package with staged manager operating just on Operations - not on ProvisioningOperations
- necessary events added 
- tests adjusted

**Related issue(s)**

Issue #1824, opened 2022-06-29
KEB: Support additional component per version and plan


PK85:
**Description**

Let's provide concept and implementation that we can easily change additional components list per Kyma version. Then going further lets allow to provide that list even for version + plan.

**Reasons**

Right now we can set additional components globally. That means for example if new Kyma version will not want to have some additional component then we cannot achieve that.

**Acceptance Criteria**
1) Prepare examples that will show we can achieve requirements in easy and understandable way (leave them in the comment under this ticket)
1.1) prepare example configuration files where all Kyma versions and plans has the same list of AddCom (include change in kcp.yaml too)
1.2) prepare example configuration files where since 2.4 Kyma version we have new AddCom list
1.3) prepare example configuration files where since 2.4 Kyma version we have new AddCom list and azure plan has different new AddCom list
1.4) prepare example configuration files where since 2.4 only azure plan has new AddCom list
1.5) prepare example configuration files where since 2.4 only azure and gcp plans have new AddCom list, and azure_lite new AddCom2
1.6) prepare example configuration files where since 2.4 trial plan has no AddCom list
2) Implement, test and document given feature

szwedm:
Providing configuration examples as stated in:
> 1. Prepare examples that will show we can achieve requirements in easy and understandable way (leave them in the comment under this ticket)

Examples are yaml files which can be created in KCP env.
>    1.1) prepare example configuration files where all Kyma versions and plans has the same list of AddCom (include change in kcp.yaml too)
```yaml
# kcp.yaml
kyma-environment-broker.defaultAdditionalRuntimeComponents: |-
  - name: "additional-component1"
    namespace: "kyma-system"
  - name: "additional-component2"
    namespace: "test-system"
  - name: "additional-component3"
    namespace: "kyma-system"
    source:
      url: "https://test.local/artifacts/additional-component3-0.0.1.tgz"
```

>    1.2) prepare example configuration files where since 2.4 Kyma version we have new AddCom list
```yaml
---
# runtimes/2.4/base/kustomization.yaml // only DELTA
commonLabels:
  runtime-version-2.4.0: "true"
  # OR
  runtime-version: "2.4.0"
resources:
  - additional-components.yaml
---
# additional-components-2.4.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: additional-components
  namespace: kcp-system
  labels:
    # one option for clarity
    runtime-version-2.4.0: "true"
    additional-components-plan-default: "true"
data:
[...]

PR #1881, opened 2022-07-20
Incorporate ConfigProvider in the queue

szwedm:
**Description**

- incorporate ConfigProvider in KEB's workflow,
- get RuntimeVersionData where necessary (ConfigProvider requirement)
- adjust tests - add ConfigProvider to the tested flows.

This is part 2 of #1875 

**Related issue(s)**
Issue: #1824


components/kyma-environment-broker/cmd/broker/broker_suite_test.go
```diff
@@ -122,6 +123,11 @@ func NewBrokerSuiteTest(t *testing.T, version ...string) *BrokerSuiteTest {
 	componentsYAML := kebRuntime.ReadYAMLFromFile(t, "kyma-components.yaml")
 	fakeHTTPClient := kebRuntime.NewTestClient(t, installerYAML, componentsYAML, http.StatusOK)
 
+	configProvider := kebConfig.NewConfigProvider(
```
piotrmiskiewicz, line 126: I'd like to see one constructor which initializes inside the reader and validator, so on the high level, you will have kebConfig.NewConfigProvider(ctx, cli, logrus.New())

components/kyma-environment-broker/internal/process/update/initialisation_step.go
```diff
@@ -28,19 +29,29 @@ const (
...
 type InitialisationStep struct {
-	operationManager *process.UpdateOperationManager
-	operationStorage storage.Operations
-	instanceStorage  storage.Instances
-	inputBuilder     input.CreatorForPlan
+	operationManager       *process.UpdateOperationManager
+	operationStorage       storage.Operations
+	instanceStorage        storage.Instances
+	runtimeStatesStorage   storage.RuntimeStates
+	runtimeVerConfigurator RuntimeVersionConfiguratorForUpdating
```
piotrmiskiewicz, line 42: runtimeStateStorage and runtimeVerConfigurator - where is it used?

components/kyma-environment-broker/internal/process/update/initialisation_step.go
```diff
@@ -100,9 +127,18 @@ func (s *InitialisationStep) Run(operation internal.UpdatingOperation, log logru
 	return s.initializeUpgradeShootRequest(operation, log)
 }
 
+func (s *InitialisationStep) getRuntimeIdFromProvisioningOp(operation *internal.UpdatingOperation, log logrus.FieldLogger) error {
```
piotrmiskiewicz, line 130: log not used

components/kyma-environment-broker/cmd/broker/broker_suite_test.go
szwedm: The idea was to explicitly show where the configuration comes from and how it's provided. On first sight we can see that ConfigProvider provides the configuration from a ConfigMap. I left some space for future implementations, for example, configuration provided from a file or some other endpoint. I think we can hide the components initialised in NewConfigProvider() if ConfigMap stays as a default source of the configuration.

components/kyma-environment-broker/internal/process/update/initialisation_step.go
szwedm: nowhere, this is a leftover from the one of options I was trying to implement for fetching last runtimeID. Thanks for spotting it, will be removed

components/kyma-environment-broker/internal/process/update/initialisation_step.go
szwedm: oversight from extracting method using IDE, thanks for spotting it, will be removed

components/kyma-environment-broker/cmd/broker/broker_suite_test.go
szwedm: We discussed the matter and decided to keep the proposed solution

piotrmiskiewicz (approved):

PR #1922, opened 2022-08-05
Consume expired

piotrmiskiewicz:
**Description**

- KEB consumes "expired" param

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/cmd/broker/update_test.go
+++ b/components/kyma-environment-broker/cmd/broker/update_test.go
@@ -236,7 +236,7 @@ func TestExpirationOfNonTrial(t *testing.T) {
 			"expired": true
 		}
    }`)
-	assert.Equal(t, http.StatusAccepted, resp.StatusCode)
+	assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
 	instance := suite.GetInstance(iid)
 	assert.False(suite.t, instance.IsExpired())
 }
@@ -1858,6 +1858,7 @@ func TestUpdateSCMigrationSuccess(t *testing.T) {
 	i, err = suite.db.Instances().GetByID(id)
 	assert.NoError(t, err, "getting instance after update")
 	assert.True(t, i.InstanceDetails.SCMigrationTriggered, "instance SCMigrationTriggered after update")
+	time.Sleep(5 * time.Millisecond)
 	rsu2, err := suite.db.RuntimeStates().GetLatestWithReconcilerInputByRuntimeID(i.RuntimeID)
 	assert.NoError(t, err, "getting runtime after update")
 	assert.NotEqual(t, rsu1.ID, rsu2.ID, "runtime_state ID from first call should differ runtime_state ID from second call")
--- a/components/kyma-environment-broker/cmd/broker/update_test.go
+++ b/components/kyma-environment-broker/cmd/broker/update_test.go
@@ -138,6 +138,109 @@ func TestUpdateFailedInstance(t *testing.T) {
 	assert.Equal(t, "Unable to process an update of a failed instance", errResponse.Description)
 }
 
+func TestExpiration(t *testing.T) {
+	// given
+	suite := NewBrokerSuiteTest(t)
+	defer suite.TearDown()
+	iid := uuid.New().String()
+
+	resp := suite.CallAPI("PUT", fmt.Sprintf("oauth/cf-eu10/v2/service_instances/%s?accepts_incomplete=true&plan_id=7d55d31d-35ae-4438-bf13-6ffdfa107d9f&service_id=47c9dcbf-ff30-448e-ab36-d3bad66ba281", iid),
+		`{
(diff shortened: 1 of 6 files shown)


components/kyma-environment-broker/internal/broker/instance_update.go
```diff
@@ -348,3 +358,31 @@ func (b *UpdateEndpoint) isKyma2(instance *internal.Instance) (bool, string, err
 	kv := s.GetKymaVersion()
 	return internal.DetermineMajorVersion(kv) == 2, kv, nil
 }
+
+// processExpirationParam returns true, if expired
+func (b *UpdateEndpoint) processExpirationParam(instance *internal.Instance, details domain.UpdateDetails) (*internal.Instance, error) {
+	// if the instance was expired before (in the past), then no need to do any work
+	if instance.IsExpired() {
```
wozniakjan, line 365: it's worth keeping this to be on the safe side but I believe this is not reachable because KEB update endpoint rejects any update requests earlier in the call stack with 422
https://github.com/kyma-project/control-plane/blob/445ba9036db31730f1073ae1f6524e606b8c0f1d/components/kyma-environment-broker/internal/broker/instance_update.go#L93-L96

components/kyma-environment-broker/internal/broker/instance_update.go
```diff
@@ -348,3 +358,31 @@ func (b *UpdateEndpoint) isKyma2(instance *internal.Instance) (bool, string, err
...
+	if len(details.RawParameters) != 0 {
+		var params internal.UpdatingParametersDTO
+		err := json.Unmarshal(details.RawParameters, &params)
+		if err != nil {
+			return instance, err
+		}
+		if params.Expired {
+			if !IsTrialPlan(instance.ServicePlanID) {
+				b.log.Warn("Expiration for non-trial instance is not supported")
+				return instance, nil
```
wozniakjan, line 377: should KEB return an error and information to the caller as well so the consumer realizes expiring non-trial is not supported?

components/kyma-environment-broker/internal/broker/instance_update_test.go
```diff
@@ -85,6 +88,58 @@ func TestUpdateEndpoint_UpdateSuspension(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// then
+
+	assert.Equal(t, internal.ERSContext{
+		Active: ptr.Bool(false),
+	}, handler.ersContext)
+
+	require.NotNil(t, handler.Instance.Parameters.ErsContext.Active)
+	assert.True(t, *handler.Instance.Parameters.ErsContext.Active)
```
wozniakjan, line 136: nit: perhaps I'm reading this wrong but the update request contains 
https://github.com/kyma-project/control-plane/blob/445ba9036db31730f1073ae1f6524e606b8c0f1d/components/kyma-environment-broker/internal/broker/instance_update_test.go#L124

I would expect this to be `assert.False(..)`, no?

resources/kcp/charts/kyma-environment-broker/values.yaml
```diff
@@ -1,8 +1,8 @@
 global:
   images:
     kyma_environment_broker:
-      dir:
-      version: "PR-1930"
+      dir:
```
wozniakjan, line 4: nit: extra whitespace

components/kyma-environment-broker/internal/broker/instance_update.go
piotrmiskiewicz: yes, you are right

components/kyma-environment-broker/internal/broker/instance_update_test.go
piotrmiskiewicz: the context change handler expects 2 arguments: an instance (before the change) and the new context so the handler can see how it was changed

components/kyma-environment-broker/internal/broker/instance_update.go
piotrmiskiewicz: fixed

components/kyma-environment-broker/cmd/broker/update_test.go
```diff
@@ -1755,6 +1858,7 @@ func TestUpdateSCMigrationSuccess(t *testing.T) {
 	i, err = suite.db.Instances().GetByID(id)
 	assert.NoError(t, err, "getting instance after update")
 	assert.True(t, i.InstanceDetails.SCMigrationTriggered, "instance SCMigrationTriggered after update")
+	time.Sleep(5 * time.Millisecond)
```
piotrmiskiewicz, line 1861: Sometimes the test fails, I'm waiting for removing the migration stuff so the test will be also removed

components/kyma-environment-broker/internal/broker/instance_update_test.go
piotrmiskiewicz: I know it is not clean...

wozniakjan (dismissed):
/lgtm
/approve

PR #1930, opened 2022-08-08
Fix getting last kyma version

piotrmiskiewicz:
**Description**

- fix getting last kyma version

The problem is that the code is looking for a runtime state with kyma version, but getting a version from such record uses different places than the filtering "where" statement.

The in memory implementation differs than SQL so unit tests were working fine

**Related issue(s)**


components/kyma-environment-broker/cmd/broker/upgrade_cluster_test.go
```diff
@@ -109,4 +109,43 @@ func TestClusterUpgrade_UpgradeAfterUpdateWithNetworkPolicy(t *testing.T) {
 	disabled := true
 	suite.AssertDisabledNetworkFilterRuntimeState(i.RuntimeID, upgradeClusterOperationID, &disabled)
 	assert.Equal(suite.t, "CUSTOMER", *instance2.Parameters.ErsContext.LicenseType)
+
```
piotrmiskiewicz, line 112: added few updates to check if it works fine

szwedm (approved):
/lgtm

PR #1942, opened 2022-08-10
set EnableNatGateway boolean to true for Azure, AzureLite and AzureHA for upgrades

tillknuesting:
As part of enabling NAT Gateway for Azure clusters, this PR sets the EnableNatGateway boolean to true for Azure, AzureLite, and AzureHA plans.

See for more information: https://github.com/kyma-project/kyma/issues/13553

When a partial payload is provided, Provisioner gets the missing information from the database with previously used data, and therefore only values that are different are changed and only providing EnableNatGateway should be sufficient.

### DIFF
--- a/components/kyma-environment-broker/internal/process/input/builder.go
+++ b/components/kyma-environment-broker/internal/process/input/builder.go
@@ -2,6 +2,7 @@ package input
 
 import (
 	"fmt"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/ptr"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


ebensom (commented):
We MUST NOT reconfigure existing cluster to use NAT gateway as it would involve change of the outbound IPs used for the exiting runtimes. Who requested this change?

tillknuesting:
> We MUST NOT reconfigure existing cluster to use NAT gateway as it would involve change of the outbound IPs used for the exiting runtimes. Who requested this change?

https://github.com/kyma-project/kyma/issues/13553#issuecomment-1167053459

PR #1945, opened 2022-08-11
Allow to expire failed instsances

piotrmiskiewicz:
**Description**

- allow to expire failed instances
- block expiration request if the active=false is not set
- use proper logger

**Related issue(s)**


wozniakjan (approved):
/lgtm
/approve

PR #1951, opened 2022-08-11
Setting expiration period in hours

jaroslaw-pieszka:
**Description**

Invalid expiration period specification caused parsing error

- Setting expiration period using hours as unit instead of days

**Related issue(s)**


szwedm (approved):
/lgtm

PR #1961, opened 2022-08-16
Send patch to broker for to be expired instances

jaroslaw-pieszka:
**Description**

Sending regular OSB API Update request from trial cleaner job with context.active = false and expired = false. This update request is supposed to trigger deletion of the trial.

- New method added to broker client - method sending at the moment PATCH request without active and expired fields set - just to check communication between trial cleaner job and KEB

**Related issue(s)**

https://github.tools.sap/kyma/backlog/issues/2730 - Task 9


piotrmiskiewicz (approved):
lgtm

PR #1972, opened 2022-08-18
KEB Trial expiration labels

szwedm:
**Description**

- add labels related to a trial expiration to the response of GetInstance endpoint

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/broker/instance_get.go
+++ b/components/kyma-environment-broker/internal/broker/instance_get.go
@@ -13,6 +13,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	trialGAForTests = "e449f875-b5b2-4485-b7c0-98725c0571bf"
+	trialSAForTests = "a45be5d8-eddc-4001-91cf-48cc644d571f"
+)
+
 type GetInstanceEndpoint struct {
 	config            Config
 	instancesStorage  storage.Instances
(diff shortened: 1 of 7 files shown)


components/kyma-environment-broker/internal/broker/instance_get.go
```diff
@@ -68,5 +73,16 @@ func (b *GetInstanceEndpoint) GetInstance(_ context.Context, instanceID string,
 			Labels: ResponseLabels(*op, *instance, b.config.URL, b.config.EnableKubeconfigURLLabel),
 		},
 	}
+
+	if b.config.ShowTrialExpireInfo {
+		// for test purposes on DEV env
+		if instance.GlobalAccountID == trialGAForTests && instance.SubAccountID == trialSAForTests {
```
wozniakjan, line 79: nit: the plan is to later remove this filter for a specific GA and SA? given it's behind a feature flag, I'd vote for showing when the flag is on (that way we can control whether it is on or off per env)

Alternatively, there could be one more flag of string type taking comma separated list of `GA:SA` for fine-grained control over the roll out?

wozniakjan (approved):
/lgtm
/approve
/hold

components/kyma-environment-broker/internal/broker/instance_get.go
szwedm: GA and SA are only temporarily placed in the code, I needed to specify my own runtime to limit the change to only one instance on DEV.

wozniakjan (approved):
/lgtm
/approve

PR #1956, opened 2022-08-12
Remove UAA cleanup from deprovisioning queue

wozniakjan:
This was a forgotten residue regarding the service catalog. Catalog is no longer part of kyma and these deprovisioning steps were noop because no UAA issuer `ServiceInstance` is part of any kyma instance.

No other step in deprovisioning uses kube clients to access the instance.

PR #1974, opened 2022-08-18
BTP-Operator cleanup for suspensions

wozniakjan:
**Description**

After removing https://github.com/kyma-project/control-plane/pull/1956, it was identified that there is currently a need for SKR cleanup of BTP Operator resources by KEB for suspensions.

KEB doesn't currently have any built-in mechanism to mark the operation as failed while still going through all the steps of deprovisioning. As a result, if the suspension operation fails to cleanup BTP-Operator resources, the errors appear in logs only and after the 30 min timeout, the operation processing continues forward with next steps as if nothing was wrong. This may result in external resource leakage but the team consensus is that this is acceptable.

### DIFF
--- a/components/kyma-environment-broker/cmd/broker/broker_suite_test.go
+++ b/components/kyma-environment-broker/cmd/broker/broker_suite_test.go
@@ -52,13 +52,12 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	coreV1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
@@ -104,8 +103,8 @@ func (s *BrokerSuiteTest) TearDown() {
 
 func NewBrokerSuiteTest(t *testing.T, version ...string) *BrokerSuiteTest {
 	ctx := context.Background()
-	sch := runtime.NewScheme()
-	require.NoError(t, coreV1.AddToScheme(sch))
+	sch := scheme.Scheme
+	apiextensionsv1.AddToScheme(sch)
 	additionalKymaVersions := []string{"1.19", "1.20", "main", "2.0"}
 	additionalKymaVersions = append(additionalKymaVersions, version...)
 	cli := fake.NewClientBuilder().WithScheme(sch).WithRuntimeObjects(fixK8sResources(defaultKymaVer, additionalKymaVersions)...).Build()
@@ -182,9 +181,7 @@ func NewBrokerSuiteTest(t *testing.T, version ...string) *BrokerSuiteTest {
 	provisioningQueue.SpeedUp(10000)
 	provisionManager.SpeedUp(10000)
 
-	scheme := runtime.NewScheme()
-	apiextensionsv1.AddToScheme(scheme)
-	fakeK8sSKRClient := fake.NewClientBuilder().WithScheme(scheme).Build()
+	fakeK8sSKRClient := fake.NewClientBuilder().WithScheme(sch).Build()
 
 	updateManager := update.NewManager(db.Operations(), eventBroker, time.Hour, logs)
 	rvc := runtimeversion.NewRuntimeVersionConfigurator(cfg.KymaVersion, nil, db.RuntimeStates())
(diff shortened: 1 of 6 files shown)


components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -432,7 +432,7 @@ func k8sClientProvider(kcfg string) (client.Client, error) {
 		return nil, err
 	}
 
-	sch := runtime2.NewScheme()
+	sch := scheme.Scheme
```
wozniakjan, line 435: the BTP-Operator cleanup needs to list `Namespaces` because `ServiceInstances` and `ServiceBindings` are namespaced resources and we need to clean them up in each namespace. As a result, I think it's best to use the default scheme from client-go rather than form a scheme from scratch and register core resources one by one when needed.

components/kyma-environment-broker/internal/process/deprovisioning/btp_operator_cleanup.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (s *BTPOperatorCleanupStep) deleteResource(k8sClient client.Client, namespaces corev1.NamespaceList, gvk schema.GroupVersionKind, log logrus.FieldLogger) (requeue bool) {
+	listGvk := gvk
+	listGvk.Kind = gvk.Kind + "List"
+	stillExistingCount := 0
+	for _, ns := range namespaces.Items {
+		list := &unstructured.UnstructuredList{}
+		list.SetGroupVersionKind(listGvk)
+		if err := k8sClient.List(context.Background(), list, client.InNamespace(ns.Name)); err != nil {
+			log.Errorf("failed listing resource %v in namespace %v", gvk, ns.Name)
```
wozniakjan, line 110: the deletion follows listing because it's desirable to allow BTP-Operator to properly delete these objects and perform finalization logic. The cleanup logic follows these steps:

1) list `ServiceInstances` and `ServiceBindings` in all namespaces
2) if none exist, consider cleanup as finished
3) if some do exist, call `DeleteAll` and requeue after 10s (limit up to 30 minutes)
4) the operation may be requeued a few times but eventually terminate as BTP-Operator removes all relevant resources

components/kyma-environment-broker/cmd/broker/broker_suite_test.go
```diff
@@ -182,9 +179,7 @@ func NewBrokerSuiteTest(t *testing.T, version ...string) *BrokerSuiteTest {
 	provisioningQueue.SpeedUp(10000)
 	provisionManager.SpeedUp(10000)
 
-	scheme := runtime.NewScheme()
-	apiextensionsv1.AddToScheme(scheme)
-	fakeK8sSKRClient := fake.NewClientBuilder().WithScheme(scheme).Build()
+	fakeK8sSKRClient := fake.NewClientBuilder().WithScheme(sch).Build()
```
wozniakjan, line 182: there is no need for multiple `scheme` in tests, the broker also uses the same `scheme` for all operation queues

components/kyma-environment-broker/internal/fixture/fixture.go
```diff
@@ -17,6 +17,7 @@ const (
 	ServiceId                   = "47c9dcbf-ff30-448e-ab36-d3bad66ba281"
 	ServiceName                 = "kymaruntime"
 	PlanId                      = "4deee563-e5ec-4731-b9b1-53b42d855f0c"
+	TrialPlan                   = "7d55d31d-35ae-4438-bf13-6ffdfa107d9f"
```
szwedm, line 20: nit: you can use plan IDs consts from broker package:
https://github.com/kyma-project/control-plane/blob/main/components/kyma-environment-broker/internal/broker/plans.go#L13

wozniakjan: I actually tried that before but it creates an import cycle, I think that is the reason why `PlanID` is also re-defined here one line higher

components/kyma-environment-broker/internal/process/deprovisioning/btp_operator_cleanup.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+	}
+	return
+}
+
+func (s *BTPOperatorCleanupStep) isNotFoundErr(err error) bool {
+	return strings.Contains(err.Error(), "not found")
+}
+
+func (s *BTPOperatorCleanupStep) retryErrors(op internal.DeprovisioningOperation, err error, log logrus.FieldLogger, msg string) (internal.DeprovisioningOperation, time.Duration, error) {
```
szwedm, line 141: I don't see any calls of this method, seems overlooked.

wozniakjan: Thanks, you are correct. I indeed forgot to replace `handleError` calls with `retryErrors`. Addressed now and also renamed to `retryOnError` as I think it better describes the function.

components/kyma-environment-broker/internal/process/deprovisioning/btp_operator_cleanup.go
```diff
@@ -0,0 +1,195 @@
...
+		return operation, 0, nil
+	}
+	if operation.RuntimeID == "" {
+		log.Info("instance has been deprovisioned already")
+		return operation, 0, nil
+	}
+
+	kclient, err := s.getKubeClient(operation, log)
+	if err != nil {
+		return s.retryOnError(operation, err, log, "failed to get kube client")
```
wozniakjan, line 66: may be worth to run only `handleError` because cleaning up finalizers likely doesn't matter?

szwedm: after discussing the matter, we decided to keep retryOnError

szwedm (approved):
/lgtm

PR #1979, opened 2022-08-19
Expiration labels fix

szwedm:
**Description**

- changes to the code introduced in #1972 with accordance to the established labels

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/broker/broker.go
+++ b/components/kyma-environment-broker/internal/broker/broker.go
@@ -33,12 +33,13 @@ type KymaEnvironmentBroker struct {
 
 // Config represents configuration for broker
 type Config struct {
-	EnablePlans                     EnablePlans `envconfig:"default=azure"`
-	OnlySingleTrialPerGA            bool        `envconfig:"default=true"`
-	URL                             string
-	EnableKubeconfigURLLabel        bool `envconfig:"default=false"`
-	IncludeAdditionalParamsInSchema bool `envconfig:"default=false"`
-	ShowTrialExpireInfo             bool `envconfig:"default=false"`
+	EnablePlans                       EnablePlans `envconfig:"default=azure"`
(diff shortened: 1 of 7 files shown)


components/kyma-environment-broker/internal/broker/broker.go
```diff
@@ -33,12 +33,13 @@ type KymaEnvironmentBroker struct {
...
-	EnableKubeconfigURLLabel        bool `envconfig:"default=false"`
-	IncludeAdditionalParamsInSchema bool `envconfig:"default=false"`
-	ShowTrialExpireInfo             bool `envconfig:"default=false"`
+	EnablePlans                       EnablePlans `envconfig:"default=azure"`
+	OnlySingleTrialPerGA              bool        `envconfig:"default=true"`
+	URL                               string
+	EnableKubeconfigURLLabel          bool   `envconfig:"default=false"`
+	IncludeAdditionalParamsInSchema   bool   `envconfig:"default=false"`
+	ShowTrialExpireInfo               bool   `envconfig:"default=false"`
+	SubaccountIDToShowTrialExpireInfo string `envconfig:"default=a45be5d8-eddc-4001-91cf-48cc644d571f"`
```
wozniakjan, line 42: nit: I think this is environment specific, I believe this defaulting belongs to kustomize overlays in management-plane config. I would vote for leaving default blank here and for tests set some dummy value, but I defer to @PK85 and @piotrmiskiewicz for their opinion.

wozniakjan (approved):
/lgtm
/approve
/hold

components/kyma-environment-broker/internal/broker/broker.go
szwedm: @wozniakjan thanks for remarks, I fully agree with you. I would left it blank if it stays in the config for a longer period of time, but it's only a temporary solution, probably will be gone in the next week.

PR #1984, opened 2022-08-22
Add pagesize of 20 to list operations command

berkanercan:
**Description**

- list operations return a huge list of operations, which is impractical. We would like to limit the default number to 20.

**Related issue(s)**

### DIFF
--- a/tools/cli/pkg/command/orchestration.go
+++ b/tools/cli/pkg/command/orchestration.go
@@ -306,11 +306,12 @@ func (cmd *OrchestrationCommand) Validate(args []string) error {
 }
 
 func (cmd *OrchestrationCommand) showOrchestrations() error {
+	cmd.listParams.PageSize = 20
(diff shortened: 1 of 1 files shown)


tools/cli/pkg/command/orchestration.go
```diff
@@ -306,11 +306,12 @@ func (cmd *OrchestrationCommand) Validate(args []string) error {
 }
 
 func (cmd *OrchestrationCommand) showOrchestrations() error {
+	cmd.listParams.PageSize = 20
 	srl, err := cmd.client.ListOrchestrations(cmd.listParams)
 	if err != nil {
 		return errors.Wrap(err, "while listing orchestrations")
 	}
-
+	cmd.listParams.PageSize = 0
```
piotrmiskiewicz, line 314: why do you set it to zero?

PR #1992, opened 2022-08-25
keb: introduce new operations metrics compass_keb_operation_result

wozniakjan:
**Description**
The `compass_keb_{operation_type}_result` metrics have architectural challenges that make them difficult to consume.

1) they are difficult to aggregate - they are gauges with values having [nontrivial meaning](https://github.com/kyma-project/control-plane/blob/8c7a89d3660b9a56f75c3e2007c5b9063ff9b921/components/kyma-environment-broker/internal/metrics/operation_result.go#L16-L23) where 0 means failed, 1 means succeeded, 2 means in process ... This makes aggregation of the data difficult. If user queries `compass_keb_provisioning_result` and gets 7, they don't know easily if that means 7 succeeded SKRs or 3 succeeded and 2 processing. Also given failed means 0, there can be an arbitrary number of failed SKRs for any aggregation query.

2) they are used for cross-correlation with `compass_keb_operations_{operation_type}_failed_total`, but `totals` are based on data from database (persisted) and `result` are based on data stored in memory (not persisted through restarts). This means the metrics are mismatching and unnecessary alerts get triggered.

This PR proposes a new metric:
```
compass_keb_operation_result{global_account_id="[ga_id]",instance_id="[i_id]",operation_id="[o_id]",plan_id="[p_id]",state="[state]",type="[o_type]"} [val]
```
The metric is based on data persisted in the database and the state is represented as a label to allow for metric aggregation and convenient filtering.

**Related issue(s)**


components/kyma-environment-broker/internal/metrics/operations_db_collector.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (s *opsMetricService) updateOperation(op internal.Operation) {
+	oldOp, found := s.cache[op.ID]
+	if found {
+		s.setOperation(oldOp, 0)
+	}
+	s.setOperation(op, 1)
+	if op.State == domain.Failed || op.State == domain.Succeeded {
+		delete(s.cache, op.ID)
```
wozniakjan, line 78: operations in one of the terminal states can be safely removed from the cache reducing the memory footprint.

components/kyma-environment-broker/internal/metrics/operations_db_collector.go
```diff
@@ -0,0 +1,119 @@
...
+			// it's not desirable to panic metrics goroutine, instead it should return and log the error
+			err = fmt.Errorf("panic recovered: %v", r)
+		}
+	}()
+	now := time.Now()
+	operations, err := s.db.ListOperationsInTimeRange(s.lastUpdate, now)
+	if err != nil {
+		return fmt.Errorf("failed to list operations: %v", err)
+	}
+	s.logger.Infof("updating operations metrics for: %v operations", len(operations))
```
wozniakjan, line 96: this is logged every 30 seconds, I think it has value checking it's operational but can be easily persuaded to remove it as too noisy

wozniakjan:
ptal @ebensom, do you think these metrics would have value? I personally dislike using `compass_keb_{operation_type}_result` because of the two mentioned points and I think this may be a better replacement.

ralikio (dismissed):

ralikio (approved):

PR #1995, opened 2022-08-26
Fix missing quote around .Values.gardener.kubernetesVersion

ebensom:
**Description**

Partial kubernetes versions (e.g. major.minor: 1.23) can be configured in Gardener, however KEB chart doesn't support it as helm will convert the input value from string to a float number automatically. 

**Related issue(s)**


wozniakjan (approved):
/lgtm
/approve

PR #2001, opened 2022-08-29
KEB returns http 200 for an update of expired instances

piotrmiskiewicz:
**Description**

- KEB returns http 200 on update endpoint for expired instance

This PR makes KEB to return 200 for second expiration request or suspension of an expired instance.

**Related issue(s)**


szwedm (approved):
/lgtm

PR #2015, opened 2022-09-02
Improve trial expiry labels

szwedm:
**Description**

- change messages in the trial expiry labels to be more specific.

**Related issue(s)**


components/kyma-environment-broker/internal/broker/instance_get.go
```diff
@@ -69,8 +69,8 @@ func (b *GetInstanceEndpoint) GetInstance(_ context.Context, instanceID string,
 		},
 	}
 
-	if b.config.ShowTrialExpireInfo && instance.ServicePlanID == TrialPlanID {
-		spec.Metadata.Labels = ResponseLabelsWithExpireInfo(*op, *instance, b.config.URL, b.config.EnableKubeconfigURLLabel)
+	if b.config.ShowTrialExpireInfo && instance.ServicePlanID == TrialPlanID && instance.SubAccountID == b.config.SubaccountIDToShowTrialExpireInfo {
```
wozniakjan, line 72: nit: personally, I would make this configuration option more flexible allowing any subset of subaccounts per environment. I see a benefit in allowing all in dev but restricting to one in prod.

wozniakjan (dismissed):
/lgtm
/approve
/hold

feel free to unhold if extensive configuratbility is not in scope

components/kyma-environment-broker/internal/broker/instance_get.go
szwedm: I liked the idea of more fexibility in providing more subaccount IDs. However, this requires more changes in the code, which would be removed not too long after it has been merged. If we keep PR as it is now, the subaccount id is read from a environment variable. Since it's impossible to store an array in a environment variable, I would need to provide a serialization/deserialization mechanism for reading subaccount IDs. And we still have to consider limitations of a ConfigMap.

piotrmiskiewicz (dismissed):

components/kyma-environment-broker/internal/broker/instance_get.go
szwedm: I changed the condition to support more subaccount IDs.

jaroslaw-pieszka (approved):

PR #2035, opened 2022-09-12
Configure btp-operator priorityClass

wozniakjan:
**Description**

BTP-Operator will allow configuring `priorityClass` in https://github.com/kyma-incubator/sap-btp-service-operator/pull/23, KEB should set it properly to `kyma-system` for the deployment.

**Related issue(s)**


szwedm (approved):
/lgtm

PR #2065, opened 2022-09-21
Non-repetitive definition of build targets for CLI tools

wozniakjan:
Followup to https://github.com/kyma-project/control-plane/pull/2053, this reduces repetitiveness of build target definitions.

The change is rather cosmetic, main goal is actually to trigger automated builds from https://github.com/kyma-project/test-infra/pull/6031 to check if they work ok.


wozniakjan:
looks like I need to brush up on bash fundamentals, double quotes don't prevent variable expansion but they do prevent globbing - https://github.com/kyma-project/test-infra/pull/6048

PR #2066, opened 2022-09-22
fix the failed data occasionally not list in `kcp o id operaitons -s failed` 

lauraanddolay:
Fix the scenario of execution  `kcp o **** operations -s failed` (see [issue2886](https://github.tools.sap/kyma/backlog/issues/2886)  ) when the `failed data` not existed in the  db `orchestration` `page:1 pagesize:100` which is cause by that we changed the way of decision `failed` operations in the [task 2681](https://github.tools.sap/kyma/backlog/issues/2681)(see below)
```
In manager code, when determining final state of orchestration (when there are no operations in pending / in progress / retrying), for each failed operation it should be checked that another operation for the same instance ID exists within the orchestration with succeeded state.
```
And currently logic is :
1. in 1st loop:
 a. if `-s filter` includes the `failed`, it will search the whole page&pages'  `failed` data with `filter = dbmodel.OperationFilter{States: []string{"failed"}}` which is default identical to  `page:0 pagesize:0` and append to the 
orl.Data.
```
#operation.go
	//only for "failed" states
	if filterFailedFound {
	-	filter.States = []string{Failed}
	+	filter = dbmodel.OperationFilter{States: []string{"failed"}}
		failedOperations, failedCount, failedtotalCount, err := s.showUpgradeKymaOperationDTOByOrchestrationID(orchestrationID, filter)
```
 b. if `filter` not includes the `failed`, go base code.

2. in 2nd and latter loops, if `-s filter` included the `failed`, removed it because 1st loop search all of `failed` datas in orchestration db, and continue with base code.
```
#client.go
for !fetchedAll {
+		if params.Page > 1 {
+			failedFound, failedIndex := c.searchFilter(params.States, "failed")
+			if failedFound {
+				params.States = c.removeIndex(params.States, failedIndex)
+			}
+		}

		req, err := http.NewRequest("GET", url, nil)
		if err != nil {
			return operations, errors.Wrap(err, "while creating request")
@@ -199,6 +206,7 @@ func (c client) ListOperations(orchestrationID string, params ListParameters) (O

		operations.TotalCount = orl.TotalCount
		operations.Count += orl.Count

		operations.Data = append(operations.Data, orl.Data...)
		if getAll {
			params.Page++
@@ -211,6 +219,31 @@ func (c client) ListOperations(orchestrationID string, params ListParameters) (O
	return operations, nil
}

```

### DIFF
--- a/components/kyma-environment-broker/internal/orchestration/handlers/orchestration_handler.go
+++ b/components/kyma-environment-broker/internal/orchestration/handlers/orchestration_handler.go
@@ -236,6 +236,8 @@ func (h *orchestrationHandler) listOperations(w http.ResponseWriter, r *http.Req
 	var response commonOrchestration.OperationResponseList
 	switch o.Type {
 	case commonOrchestration.UpgradeKymaOrchestration:
+		fmt.Println("orchestrationID is:", orchestrationID)
+		fmt.Println("filter is:", filter)
 		operations, count, totalCount, err := h.operations.ListUpgradeKymaOperationsByOrchestrationID(orchestrationID, filter)
 		if err != nil {
 			h.log.Errorf("while getting operations: %v", err)
--- a/components/kyma-environment-broker/internal/orchestration/handlers/orchestration_handler.go
+++ b/components/kyma-environment-broker/internal/orchestration/handlers/orchestration_handler.go
@@ -236,8 +236,6 @@ func (h *orchestrationHandler) listOperations(w http.ResponseWriter, r *http.Req
 	var response commonOrchestration.OperationResponseList
 	switch o.Type {
 	case commonOrchestration.UpgradeKymaOrchestration:
-		fmt.Println("orchestrationID is:", orchestrationID)
-		fmt.Println("filter is:", filter)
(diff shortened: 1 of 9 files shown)


components/kyma-environment-broker/internal/orchestration/handlers/orchestration_handler.go
```diff
@@ -236,6 +236,8 @@ func (h *orchestrationHandler) listOperations(w http.ResponseWriter, r *http.Req
 	var response commonOrchestration.OperationResponseList
 	switch o.Type {
 	case commonOrchestration.UpgradeKymaOrchestration:
+		fmt.Println("orchestrationID is:", orchestrationID)
+		fmt.Println("filter is:", filter)
```
wozniakjan, line 240: I don't think these `Println` statements are a good idea, I would recommend using logger.

lauraanddolay: Hi @wozniakjan I will remove in next version and i would like to make a build image to test resetting totalCount.

lauraanddolay: hi @wozniakjan i fix the println issue and could you help to take a review? thanks.

ebensom (approved):

PR #2069, opened 2022-09-22
Aggregate error from operation manager when marking operation as failed

wozniakjan:
**Description**
The `OperationManager` was swallowing errors from steps executions and only returning errors when storage update itself failed which resulted in metrics for failed operations no longer contained `LastError` categorization. This bugifx should address the problem.

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/process/operation_manager.go
+++ b/components/kyma-environment-broker/internal/process/operation_manager.go
@@ -4,11 +4,10 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/storage/dberr"
-
 	"github.com/kyma-project/control-plane/components/kyma-environment-broker/common/orchestration"
 	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal"
 	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/storage"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/storage/dberr"
 
 	"github.com/pivotal-cf/brokerapi/v8/domain"
 	"github.com/sirupsen/logrus"
@@ -29,7 +28,15 @@ func (om *OperationManager) OperationSucceeded(operation internal.Operation, des
 
 // OperationFailed marks the operation as failed and returns status of the operation's update
 func (om *OperationManager) OperationFailed(operation internal.Operation, description string, err error, log logrus.FieldLogger) (internal.Operation, time.Duration, error) {
-	return om.update(operation, domain.Failed, description, log)
+	op, t, updateErr := om.update(operation, domain.Failed, description, log)
+	retErr := fmt.Errorf("operation failed")
+	if err != nil {
+		retErr = fmt.Errorf("%v: %w", retErr, err)
+	}
+	if updateErr != nil {
+		retErr = fmt.Errorf("%v: %w", retErr, updateErr)
+	}
+	return op, t, retErr
 }
 
 // OperationCanceled marks the operation as canceled and returns status of the operation's update
--- a/components/kyma-environment-broker/internal/process/deprovisioning/check_runtime_removal_step_test.go
+++ b/components/kyma-environment-broker/internal/process/deprovisioning/check_runtime_removal_step_test.go
@@ -58,6 +58,6 @@ func TestCheckRuntimeRemovalStep_ProvisionerFailed(t *testing.T) {
 	op, _, err := svc.Run(dOp, log)
 
 	// then
-	require.NoError(t, err)
+	require.Error(t, err)
 	assert.Equal(t, domain.Failed, op.State)
 }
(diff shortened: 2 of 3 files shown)


tillknuesting (approved):

components/kyma-environment-broker/internal/process/operation_manager.go
```diff
@@ -29,7 +28,15 @@ func (om *OperationManager) OperationSucceeded(operation internal.Operation, des
 
 // OperationFailed marks the operation as failed and returns status of the operation's update
 func (om *OperationManager) OperationFailed(operation internal.Operation, description string, err error, log logrus.FieldLogger) (internal.Operation, time.Duration, error) {
-	return om.update(operation, domain.Failed, description, log)
+	op, t, updateErr := om.update(operation, domain.Failed, description, log)
```
vvxxvvxx, line 31: if t !=0, it should directly return `return op, t, nil`, right?

components/kyma-environment-broker/internal/process/operation_manager.go
```diff
@@ -29,7 +28,15 @@ func (om *OperationManager) OperationSucceeded(operation internal.Operation, des
...
+	retErr := fmt.Errorf("operation failed")
```
vvxxvvxx, line 32: i think below part could use the original logic. when `updateErr != nil (t!==0)`, it's returned `return op, t, nil`. 
so here is to handle the returned error combining `descritpion` and real `err`.
```
	var retErr error
	if err == nil {
		// no exact err passed in
		retErr = errors.New(description)
	} else {
		// keep the original err object for error categorizer
		retErr = errors.Wrap(err, description)
	}
```

wozniakjan: the error aggregation is done here too, just without using deprecated and archived https://github.com/pkg/errors and instead using error wrapping that is part of go core
https://github.com/wozniakjan/control-plane/blob/b10c665c95baa1ab2d181747d65a4c50f2239c30/components/kyma-environment-broker/internal/process/operation_manager.go#L32-L38

components/kyma-environment-broker/internal/process/operation_manager.go
wozniakjan: I'm not sure it makes any difference. It is actually written twice, once by the step and it looks like on error, the new staged manager tries to write it to the db as well

https://github.com/kyma-project/control-plane/blob/b10c665c95baa1ab2d181747d65a4c50f2239c30/components/kyma-environment-broker/internal/process/staged_manager.go#L194-L205

but most importantly, when operation is marked as failed, the staged manager ignores the requeue period
https://github.com/kyma-project/control-plane/blob/b10c665c95baa1ab2d181747d65a4c50f2239c30/components/kyma-environment-broker/internal/process/staged_manager.go#L140-L148

components/kyma-environment-broker/internal/process/operation_manager.go
vvxxvvxx: Hi Jan, in the current staged manager logic [(https://github.com/kyma-project/control-plane/blob/0e8ca0d46165094a29087d4765d2477d8357b966/components/kyma-environment-broker/internal/process/staged_manager.go#L221)](https://github.com/kyma-project/control-plane/blob/0e8ca0d46165094a29087d4765d2477d8357b966/components/kyma-environment-broker/internal/process/staged_manager.go#L221), if the duration returned from update() which is t here !=0, the operation will be retried. if you return an err `retErr = fmt.Errorf("%v: %w", retErr, updateErr)`, then the loop will break and stop retrying. Then the duration t will be no use. 

the err from update() will not be considered as the source of err to return, as if update() returns err!=0, t will be !=0, it will be retried.

in your codes, `retErr = fmt.Errorf("%v: %w", retErr, err)` will overwrite the source of err. because LastError relies on the wrapped err, and in https://github.com/kyma-project/control-plane/blob/0e8ca0d46165094a29087d4765d2477d8357b966/components/kyma-environment-broker/internal/error/errors.go#L90, it will unwrap the source of err to get its error category and err reason.

components/kyma-environment-broker/internal/process/operation_manager.go
vvxxvvxx: This is outside of m.runStep() already. inside m.runStep(), if less than 10*time.Minute, it will be retried.
https://github.com/kyma-project/control-plane/blob/0e8ca0d46165094a29087d4765d2477d8357b966/components/kyma-environment-broker/internal/process/staged_manager.go#L221

wozniakjan: ah, you are right!

components/kyma-environment-broker/internal/process/operation_manager.go
wozniakjan: ah, I see the unwrap logic is probably not compatible with post go 1.13 error wrapping and only works with the deprecated wrapping from https://github.com/pkg/errors. I will put the old logic back and make a note to replace https://github.com/pkg/errors all at once

PR #2092, opened 2022-09-29
Update gardener-creds-secret.yaml

i332322:
gardener-sa credential rotation is handled by SRE. there's already a cronJob implemented to do the rotation.
this part of provisioner can override the secret when helm chart is applied, hence the changes.


resources/kcp/charts/provisioner/templates/gardener-creds-secret.yaml
```diff
@@ -1,9 +1,11 @@
-{{- if eq .Values.gardener.manageSecrets true }}
+{{- if eq .Values.gardener.manageSecrets false }}
```
ebensom, line 1: This must not be changed, we need to provide an override for this value in config :)
```suggestion
{{- if eq .Values.gardener.manageSecrets false }}
```

ebensom (changes requested):

PR #2103, opened 2022-10-04
disabling gardener sa handling from provisioner

i332322:
Disabling gardener sa handling from provisioner component.

### DIFF
--- a/resources/kcp/charts/provisioner/values.yaml
+++ b/resources/kcp/charts/provisioner/values.yaml
@@ -14,6 +14,7 @@ security:
   skipTLSCertificateVeryfication: false
 
 gardener:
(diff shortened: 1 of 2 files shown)


resources/kcp/charts/provisioner/values.yaml
```diff
@@ -23,7 +24,7 @@ gardener:
   maintenanceWindowConfigMapName: ""
   secretName: "gardener-credentials"
   auditLogsPolicyConfigMap: ""
-  manageSecrets: true
+  manageSecrets: false
```
ebensom, line 27: Please don't modify either the chart template, or the chart default values. We should provide an override in our config repo.

PR #2124, opened 2022-10-07
Allow to retrigger suspension of expired instance

piotrmiskiewicz:
**Description**

- an expired instance could be "resuspended" if the last suspension has failed.
- the scenario is covered by the test: https://github.com/kyma-project/control-plane/blob/f00e19395114a801d65947d6aa9721e4d72cf69e/components/kyma-environment-broker/cmd/broker/update_test.go#L144

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/suspension/handler.go
+++ b/components/kyma-environment-broker/internal/suspension/handler.go
@@ -73,7 +73,6 @@ func (h *ContextUpdateHandler) handleContextChange(newCtx internal.ERSContext, i
 			return false, nil
 		}
 		if !isActivated {
-			l.Debugf("not activated, %s", lastDeprovisioning.State)
 			// instance is inactive and incoming context update is suspension - verify if KEB should retrigger the operation
 			if lastDeprovisioning.Temporary && (lastDeprovisioning.State == domain.Failed) {
 				l.Infof("Retriggering suspension for instance id %s", instance.InstanceID)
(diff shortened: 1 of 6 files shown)


jaroslaw-pieszka (dismissed):

piotrmiskiewicz:
fixes: https://github.tools.sap/kyma/backlog/issues/3038

components/kyma-environment-broker/internal/suspension/handler.go
```diff
@@ -125,6 +125,10 @@ func (h *ContextUpdateHandler) suspend(instance *internal.Instance, log logrus.F
 }
 
 func (h *ContextUpdateHandler) unsuspend(instance *internal.Instance, log logrus.FieldLogger) error {
+	if instance.IsExpired() {
+		log.Info("Expired instance cannot be unsuspended")
```
wozniakjan, line 129: nit: maybe return this as an error so the caller knows not to keep trying? I can imagine being confused why my `PATCH` with `active: true` is returning OK 200 but nothing is happening and my instance is still suspended

piotrmiskiewicz: it is by design. Expiration is not triggered by BTP (user). From BTP point of view it would be a bug - we do not accept unsuspension.

wozniakjan (approved):
/lgtm
/approve

PR #2132, opened 2022-10-11
Fix false nil errors for KEB events

wozniakjan:
**Description**

Error wrapping check for nil errors

**Related issue(s)**


szwedm (dismissed):
/lgtm

Issue #1912, opened 2022-08-04
Change the administrator binding label to something more appropriate


janmedrek:
PR for review https://github.com/kyma-project/control-plane/pull/2866

Follow Up Issue: https://github.com/kyma-project/control-plane/issues/2858 

**Description**

We would need to use something better suited than the `type=admin` label on the administrator Role Binding that we use right now.

We could potentially use:
- `reconciler.kyma-project.io/managed-by=reconciler` that is implemented right now and blocks any editing in Busola
- `provisioner.kyma-project.io/managed-by=provisioner` to mimic the approach we have in reconciler

With the Cluster Operator in mind, the first option makes sense and it could be considered - at some point Provisioner will be migrated to the reconciler scheme.

Also, we could potentially think about `reconciler.kyma-project.io/managed-by=provisioner` if Busola is somehow configured to only block resources based on the label key.

**Reasons**

We want to have clear labels that indicate which resources are managed by the Control Plane.

**Acceptance Criteria**

- Label changed to something that indicates the Binding is managed by the Provisioner.
- Provisioner should remove bindings with both `type=admin` and the new `reconciler.kyma-project.io/managed-by=reconciler` labels during the upgrade (and then apply the desired ones), this is to avoid name conflicts with existing bindings that use the old label. Later on, the Cluster Operator should use only the new label.

janmedrek:
Deletion of labelled bindings: https://github.com/kyma-project/control-plane/blob/fc5da85e9872f49a5e73910733503ee549112029/components/provisioner/internal/operations/stages/provisioning/create_operators_bindings.go#L139

and the creation: https://github.com/kyma-project/control-plane/blob/fc5da85e9872f49a5e73910733503ee549112029/components/provisioner/internal/operations/stages/provisioning/create_operators_bindings.go#L131

koala7659:
More details for developers: 

During Kyma cluster creation process provisioner creates some Kubernetes `clusterroles` and `clusterrolebindings` on newly provisioner Gardener cluster (gardener shoot). Those objects are required later on for correct setup of the cluster. As described in the original issue description labeling of one of the bindings should be changed. 

**More source code context:** 

The step of creating those  `clusterroles` and `clusterrolebindings` is part of the provisioning queue executed immediately after fresh Kubernetes cluster becomes active as the third operation (should be read from the bottom).

The provisioning queues are built in the file  

`components/provisioner/internal/operations/queue/queues_constructor.go`
For Kyma 1.x

```
[...]

PR #2133, opened 2022-10-11
Change the administrator binding label to something more appropriate

VOID404:
**Description**

I used `reconciler.kyma-project.io/managed-by=reconciler` as suggested in the original Issue (#1912).
I kept the `type=admin`, so we don't remove some other reconciler managed resources.
For now all Cluster Role Bindings with `type=admin` will still be removed, this has to change after clusters are upgraded.

- Add `reconciler.kyma-project.io/managed-by=reconciler` to Cluster Role Bindings for admins
- Add filter to remove Cluster Role Bindings with new label

**Related issue(s)**

#1912


components/provisioner/internal/operations/stages/provisioning/create_operators_bindings.go
```diff
@@ -128,17 +129,25 @@ func (s *CreateBindingsForOperatorsStep) Run(cluster model.Cluster, _ model.Oper
...
+						"type":                                  "admin",
+					}))
 		}
 	}
 
 	if err := createClusterRoles(k8sClient.RbacV1().ClusterRoles(), clusterRoles); err != nil {
 		return operations.StageResult{}, err
 	}
 
+	//TODO: remove me after all clusters migrated
```
franpog859, line 144: I'd not add a TODO in the code :d It ends up living there for years. It's better to make a follow-up issue

franpog859 (commented):
I think I'd add these labels to the L2 and L3 ClusterRoleBindings too

components/provisioner/internal/operations/stages/provisioning/create_operators_bindings.go
VOID404: Fixed

VOID404:
Moved to #2209

PR #2166, opened 2022-10-19
Enhance upgrade command to promt the user to continue

berkanercan:
**Description**

- Added two templates for previewing kyma and k8s upgrades
- Added the logic to promt the user before continuing the upgrade. This allows the user to check the orchestration parameters before starting the execution of the upgrade.

**Related issue(s)**


tools/cli/pkg/command/orchestration.go
```diff
@@ -134,6 +134,40 @@ Operations:
 {{- end }}
 `
 
+var kymaUpgradePreviewTpl = `Kyma Upgrade Preview
+Strategy:         {{.Parameters.Strategy.Type}}
```
ebensom, line 138: All the Strategy, Schedule, Workers and Kyma Version are optional parameters. Currently they don'T have meaningful default values, and server applies default values for them.
See https://github.com/berkanercan/control-plane/blob/main/tools/cli/pkg/command/upgrade.go#L45-L47

I suggest to introduce the following explicit default values in KCP CLI:
- strategy: parallel
- schedule: immediate
- Workers: 1

tools/cli/pkg/command/orchestration.go
```diff
@@ -134,6 +134,40 @@ Operations:
...
+Schedule:         {{.Parameters.Strategy.Schedule}}
+Workers:          {{.Parameters.Strategy.Parallel.Workers}}
+Kyma Version:     {{with .Parameters.Kyma}}{{.Version}}{{end}}
```
ebensom, line 141: If .Parameters.Kyma.Version is not empty, it should be displayed. Otherwise this should be displayed:

`Kyma Version: <determined after start>`

tools/cli/pkg/command/orchestration.go
```diff
@@ -134,6 +134,40 @@ Operations:
...
+  {{ orchestrationTarget $t }}
+{{- end -}}
+{{- end }}
+`
+
+var kubernetesUpgradePreviewTpl = `Kubernetes Upgrade Preview
+Strategy:         {{.Parameters.Strategy.Type}}
+Schedule:         {{.Parameters.Strategy.Schedule}}
+Workers:          {{.Parameters.Strategy.Parallel.Workers}}
+K8s Version:      {{with .Parameters.Kubernetes}}{{.KubernetesVersion}}{{end}}
```
ebensom, line 158: Currently only dynamic KEB determination is supported.

```suggestion
K8s Version:      <determined after start>
```

ebensom (approved):

ebensom (approved):

PR #2179, opened 2022-10-21
Enable Preview plan in KEB

piotrmiskiewicz:
**Description**

- create Preview plan
- does not bump the image, this PR is needed for next Lifecycle Manager integration tasks

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/cmd/broker/main.go
+++ b/components/kyma-environment-broker/cmd/broker/main.go
@@ -666,6 +666,8 @@ func NewProvisioningProcessingQueue(ctx context.Context, provisionManager *proce
 		{
 			stage: createRuntimeStageName,
 			step:  provisioning.NewOverridesFromSecretsAndConfigStep(db.Operations(), runtimeOverrides, runtimeVerConfigurator),
+			// Preview plan does not calls Reconciler so it does not need overrides
+			condition: skipForPreviewPlan,
 		},
 		{
 			condition: provisioning.WhenBTPOperatorCredentialsProvided,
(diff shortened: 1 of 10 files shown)


components/kyma-environment-broker/cmd/broker/provisioning_test.go
```diff
@@ -46,6 +46,34 @@ func TestProvisioning_HappyPath(t *testing.T) {
...
+					"parameters": {
+						"name": "testing-cluster"
+					}
+		}`)
+	opID := suite.DecodeOperationID(resp)
+
+	suite.processProvisioningByOperationID(opID)
+
+	suite.WaitForOperationState(opID, domain.Succeeded)
+	// todo: assert Kyma resource is created
```
wozniakjan, line 74: nit: is it planned as part of this PR or a later follow-up?

components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -666,6 +666,8 @@ func NewProvisioningProcessingQueue(ctx context.Context, provisionManager *proce
 		{
 			stage: createRuntimeStageName,
 			step:  provisioning.NewOverridesFromSecretsAndConfigStep(db.Operations(), runtimeOverrides, runtimeVerConfigurator),
+			// Preview plan does not calls Reconciler so it does not need overrides
```
wozniakjan, line 669: nit: in that case, I think we can skip BTP Operator related steps as well here and in the `update` queue.

wozniakjan (commented):
if the `update` operations are allowed for preview plans, we may want to skip reconciler interactions there too

components/kyma-environment-broker/cmd/broker/provisioning_test.go
piotrmiskiewicz: it is for later PR when create kyma steop is done

components/kyma-environment-broker/cmd/broker/main.go
piotrmiskiewicz: BTP operator steps must be run - we will provide thhe secret directly to SKR

wozniakjan: and just to make sure, is it ok to run these reconciler interactions in the update queue? or do we want to also skip them?
https://github.com/kyma-project/control-plane/blob/e3b5f1eab946516db36765a7bf8a56c81430f0e5/components/kyma-environment-broker/cmd/broker/main.go#L767-L774

wozniakjan: for the record, @piotrmiskiewicz just explained to me it's actually skipped for preview plan, just [here](https://github.com/kyma-project/control-plane/pull/2179/files#diff-84e967d32a4a3ed65022482cc8724cfd81b0b8e902c2a835775a40e009e04b49R10), I completely missed that

wozniakjan (approved):

PR #2209, opened 2022-10-26
Change the administrator binding label to something more appropriate

VOID404:
**Description**

I used `reconciler.kyma-project.io/managed-by=reconciler` as suggested in the original Issue (#1912).
I kept the `type=admin`, so we don't remove some other reconciler managed resources.
For now all Cluster Role Bindings with `type=admin` will still be removed, this has to change after clusters are upgraded.

- Add `reconciler.kyma-project.io/managed-by=reconciler` to Cluster Role Bindings for admins
- Add filter to remove Cluster Role Bindings with new label

**Related issue(s)**

#1912


franpog859 (approved):
The Provisioner correctly adds the `reconciler.kyma-project.io/managed-by: reconciler` label to the administrators' bindings. It does not add it to the L2 and L3 operators' bindings while upgrading the shoot though (as they are not recreated due to the lack of the `type: admin` label). As it wasn't the acceptance criteria in this issue it's fine for me. The labels will be added to the newly created runtimes. 

Please, remember to create a follow-up issue to change the label selection in the deletion process after the next shoot upgrade!

PR #2257, opened 2022-11-10
Update Dockerfile for kubeconfig-service to use scratch image

hamdymoussa:
Update Dockerfile for kubeconfig-service to use scratch image.

### DIFF
--- a/components/kubeconfig-service/Dockerfile
+++ b/components/kubeconfig-service/Dockerfile
@@ -17,8 +17,8 @@ COPY . .
 RUN go build -v -o main ./cmd/generator/main.go
 RUN mkdir /app && mv ./main /app/main
 
-FROM eu.gcr.io/kyma-project/external/alpine:3.16.2
-LABEL source = git@github.com:kyma-project/control-plane.git
+
+FROM eu.gcr.io/kyma-project/external/alpine:3.16.2 as alpine
 
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


components/kubeconfig-service/Dockerfile
```diff
@@ -17,26 +17,25 @@ COPY . .
...
 # Copy binary
-#
-
 COPY --from=builder /app /app
-
-#
+COPY --from=alpine /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=alpine /usr/bin/openssl /usr/bin/
+COPY --from=alpine /bin/busybox /bin/
+COPY --from=alpine /usr/lib/* /usr/lib/
```
ebensom, line 39:
```suggestion
COPY --from=alpine /usr/bin/openssl /usr/bin/
COPY --from=alpine /bin/busybox /bin/
COPY --from=alpine /usr/lib/* /usr/lib/
```

I believe we won't need these.

ebensom (commented):
Also please don't forget to bump the kubeconfig-service image here:
https://github.com/kyma-project/control-plane/blob/main/resources/oidc-kubeconfig-service/values.yaml#L26

components/kubeconfig-service/Dockerfile
hamdymoussa: Ok i'll remove them, i just kept them as the original image.

PR #2264, opened 2022-11-14
KEB ClusterRole to manage Secrets in all namespaces for Lifecycle Manager

wozniakjan:
**Description**

https://github.com/kyma-project/control-plane/pull/2261 allowed KEB to manage `Secrets` in `kcp-system` only, but they primarily reside in `kyma-system` at the moment with a plan to potentially put them in any other namespace.

**Related issue(s)**


wozniakjan:
no longer necessary, `Secrets` will reside in `kcp-system` instead so KEB won't need access to `kyma-system`

PR #2265, opened 2022-11-14
Fix Apply kyma step - more logging

piotrmiskiewicz:
**Description**

- more logs in apply kyma steps

**Related issue(s)**


resources/kcp/charts/kyma-environment-broker/templates/rbac.yaml
```diff
@@ -19,6 +19,9 @@ rules:
   - apiGroups: ["core.gardener.cloud"]
     resources: ["secretbindings"]
     verbs: ["list", "get", "update"]
+  - apiGroups: [ "operator.kyma-project.io" ]
+    resources: [ "kymas" ]
+    verbs: [ "list", "get", "update", "create" ]
```
wozniakjan, line 24: I think we will need 'delete' too at which point it's easier to give all as '*'

PR #2277, opened 2022-11-17
Assert the removal of service binding and instances during trial deprovisioning

jaroslaw-pieszka:
**Description**

Logic to cover the case when during `trial` deprovisioning (suspension) we are not able to remove service bindings and service instances. During BTPOperator_Cleanup step we should remove all service bindings and all service instances.

- While testing trial plan after SKR creation we add one service instance and one service binding
- After deprovisioning we check if the resources no longer exist

**Related issue(s)**


wozniakjan (approved):
/approve
/lgtm

Issue #2256, opened 2022-11-10
Allow KCP operators to get events from deleted service instances


PK85:
**Description**

There are cases where we want easily check in the KEB component what happend to the deleted service instance in the past.

**Acceptance Criteria**

1. - [ ] KEB APIs return events for deleted service instances. The response includes information that the service instance doesn't exist in KEB anymore (was deleted)
2. - [ ] KCP CLI adjusted + help documentation updated
3. - [ ] extends retention policy for keeping events for two weeks (right now it's 3 days)

wozniakjan:
findings:
* rename the flag to `--only-deleted`
* restrict the API output to display only deleted
* keep instance ID for KCP as mandatory
* restrict events to up to 100 instances

PR #2288, opened 2022-11-22
KCP skip fetching events for too many instances

wozniakjan:
**Description**

During the demo of tracing events for deleted instances, it was recommended to limit on KCP side fetching events for more than 100 instances - https://github.com/kyma-project/control-plane/issues/2256#issuecomment-1321910703

### DIFF
--- a/tools/cli/pkg/command/runtime.go
+++ b/tools/cli/pkg/command/runtime.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"golang.org/x/oauth2"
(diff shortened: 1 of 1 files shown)


tools/cli/pkg/command/runtime.go
```diff
@@ -116,21 +117,29 @@ func (cmd *RuntimeCommand) Run() error {
...
+				return errors.Wrap(err, "while listing events")
+			}
 		}
 	}
 	err = cmd.printRuntimes(rp, eventList)
 	if err != nil {
 		return errors.Wrap(err, "while printing runtimes")
 	}
+	if eventsSkipped {
+		fmt.Fprintln(os.Stderr, "\nPlease narrow down the instance list by additional filters. fetching events limitted to 100 instances, received", rp.Count)
```
ralikio, line 141: What do you think about printing this info message where `eventsSkipped` is set to true?

ralikio (approved):

tools/cli/pkg/command/runtime.go
wozniakjan: I had that initially but my problem was it could get buried all the way on the top by thousands of lines of output
```
$ kcp rt --events
Please narrow down the instance list by additional filters. fetching events limitted to 100 instances, received 1895
.....
.....
```
while now it shows on the bottom and is not that easy to miss the information
```
$ kcp rt --events
.....
.....

Please narrow down the instance list by additional filters. fetching events limitted to 100 instances, received 1895
```

PR #2341, opened 2022-12-15
Do not Log kubeconfig hash

jaroslaw-pieszka:
**Description**

Currently hash is logged as binary converted to string.

- `kubeconfig` hash is not logged any more

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/process/deprovisioning/btp_operator_cleanup.go
+++ b/components/kyma-environment-broker/internal/process/deprovisioning/btp_operator_cleanup.go
@@ -2,8 +2,6 @@ package deprovisioning
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/base64"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


wozniakjan (dismissed):
/lgtm
/approve

wozniakjan (dismissed):

wozniakjan (dismissed):

ralikio (changes requested):
Do we need to log that hash of Kubeconfig? It seems to be kind of a secret for me. Can we remove that line completely?

jaroslaw-pieszka:
> Do we need to log that hash of Kubeconfig? It seems to be kind of a secret for me. Can we remove that line completely?

It is secret hence we log hash of it. To detect differences but to conceal the secret itself.

jaroslaw-pieszka:
@TorstenD-SAP suggested not to log `kubeconfig` hash since `kubeconfig` may contain static tokens and logging hash make brute force attack possible .

tillknuesting (approved):

PR #2380, opened 2023-01-04
Encrypt data in Provisioners database

franpog859:
**Description**

- Encrypt kubeconfigs and administrators' user IDs in the Provisioner's database. While provisioning, encrypt them right away. While upgrading the shoot, check if they are encrypted, if not, encrypt them
- Keep in mind that while reverting the schema changes, the encrypted values won't be decrypted. I see a potential scenario where it could lead to some faults. Let's have the encryption logic already merged, provision a runtime, revert the schema changes, apply the schema changes again, and upgrade the runtime. In such a case the kubeconfig is double encrypted so the runtime status query will provide the encrypted value

When every runtime is upgraded and encrypted, remove the upgrade and encryption checking logic so there is less clutter

**Related issue(s)**

See `https://someinternalgithub/kyma/backlog/issues/3133`


mvshao (dismissed):
LGTM

PR #2391, opened 2023-01-12
kubeconfig-service: use SA TokenRequest API instead of secret bound token

ebensom:
**Description**

In K8S 1.24, the default behavior of generating tokens for service accounts has changed:

> The LegacyServiceAccountTokenNoAutoGeneration feature gate is beta, and enabled by default. When enabled, Secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount. Use the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API to acquire service account tokens, or if a non-expiring token is required, create a Secret API object for the token controller to populate with a service account token by following this [guide](https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets). (https://github.com/kubernetes/kubernetes/pull/108309, [@zshihang](https://github.com/zshihang))

To accommodate the breaking change in default behavior, kubeconfig-service shall now use TokenRequest API to request short-living token.

### DIFF
--- a/components/kubeconfig-service/pkg/runtime/runtime.go
+++ b/components/kubeconfig-service/pkg/runtime/runtime.go
@@ -11,6 +11,7 @@ import (
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
+	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


i332322 (approved):
very elegant

PR #2402, opened 2023-01-16
For testing purposes always fail instance creation (EU Access story)

jaroslaw-pieszka:
**Description**

We need to test how BTP presents the result of instance creation failure.

- Every instance creation request with predefined subaccont ID fails with a distinctive message.

**Related issue(s)**


components/kyma-environment-broker/internal/broker/instance_create.go
```diff
@@ -142,6 +142,11 @@ func (b *ProvisionEndpoint) Provision(ctx context.Context, instanceID string, de
 		return b.handleExistingOperation(existingOperation, provisioningParameters)
 	}
 
+	// temporarily reject all request to test one scenario in EU Access story
```
PK85, line 145: we should fail only for one Subacccount ID. I will show you how to create such SA in the BTP cockpit and you can then use it.

PR #2412, opened 2023-01-19
Store failed steps during deprovisioning

MarekMichali:
**Description**

If a step cannot finish his job and goes to the next one, the step needs to be marked. The idea is to store those steps

- Store the step that was retried via `RetryOperationWithoutFail` without success in the DB

**Related issue(s)**


components/kyma-environment-broker/internal/model.go
```diff
@@ -216,6 +216,7 @@ type Operation struct {
 	ClusterConfigurationDeleted bool      `json:"clusterConfigurationDeleted"`
 	Retries                     int       `json:"-"`
 	ReconcilerDeregistrationAt  time.Time `json:"reconcilerDeregistrationAt"`
+	FailedToFinishSteps         []string  `json:"failedToFinishSteps"`
```
piotrmiskiewicz, line 219: ExcutedButNotCompleted

piotrmiskiewicz (approved):

Issue #2406, opened 2023-01-18
KEB-core: Implement core logic in the KEB to be able to retry failed steps for deleted service instances


PK85:
**Description**

This is part of [Identify failed steps in KEB deprovisioning](https://github.com/kyma-project/control-plane/issues/2392)
This ticket is about the main implementation in the KEB component.

**AC**

- Once the step cannot do the job and goes to the next one, marks the operation that such step needs a retry. All deprovisioning steps must be reviewed (13 steps). Consider some steps could fail deprovisioning instead of marking "needs to retry". Check if the page is still up to date: https://github.com/kyma-project/control-plane/blob/main/docs/kyma-environment-broker/03-03-runtime-operations.md
- at the end, the instance is not permanently removed, but marked as deleted (deleted_at)
- the get_instance endpoint has a check, if deleted_at is set. If yes, returns 404 (or 410). Check if any instance contains this field in the DB.

MarekMichali:
Deprovisioning steps mentioned here https://github.com/kyma-project/control-plane/blob/main/docs/kyma-environment-broker/03-03-runtime-operations.md are up to date.

MarekMichali:
First PR: #2412

wozniakjan:
> at the end, the instance is not permanently removed, but marked as deleted (deleted_at)

I think it might be possible to keep the KEB behavior as is at the moment and it could be **\*not necessary\*** to keep the instance in the `Instances` table - https://github.com/kyma-project/control-plane/issues/2407#issuecomment-1398347464. KCP has `--only-deleted`, the failed steps are stored on the operation in `Operations` table and the GC job can find it there, but not sure how easy would it be to re-run operations if the entry from `Instances` table is gone.

MarekMichali:
Second PR for marking operations that require retry: #2418
Covered steps:
- BTPOperator_Cleanup

MarekMichali:
Third PR for marking operations that require retry: #2422 
Covered steps:
- Check_Cluster_Deregistration
- Deregister_Cluster
- EDP_Deregistration
- Release_Subscription

MarekMichali:
PR for marking the instalce as deleted using deleted_at: #2429

MarekMichali:
Think about metrics

jaroslaw-pieszka:
Job retriggering deprovision - https://github.com/kyma-project/control-plane/pull/2449

jaroslaw-pieszka:
[...]

PR #2429, opened 2023-01-24
Mark instance as deleted_at without removing it

MarekMichali:
**Description**

Once the step cannot do the job and goes to the next one, marks the operation that such step needs a retry. at the end, the instance is not permanently removed, but marked as deleted (deleted_at). See #2406 

- mark the instance using deleted_at
- adjust tests
- adjust SQL queries

**Related issue(s)**


piotrmiskiewicz:
Please think about the following AC:
 - an instance with deleted_at is not available in get_instance endpoint
 - when KEB gets trial provisioning request, KEB checks if the instance exists (must take deleted_at field into account)

components/kyma-environment-broker/internal/broker/instance_get.go
```diff
@@ -51,19 +51,24 @@ func (b *GetInstanceEndpoint) GetInstance(_ context.Context, instanceID string,
 		statusCode := http.StatusNotFound
 		if !dberr.IsNotFound(err) {
 			statusCode = http.StatusInternalServerError
+			return domain.GetInstanceDetailsSpec{}, apiresponses.NewFailureResponse(fmt.Errorf("failed to get instanceID %s", instanceID), statusCode, fmt.Sprintf("failed to get instanceID %s", instanceID))
 		}
-		return domain.GetInstanceDetailsSpec{}, apiresponses.NewFailureResponse(err, statusCode, fmt.Sprintf("failed to get instanceID %s", instanceID))
+		return domain.GetInstanceDetailsSpec{}, apiresponses.NewFailureResponse(fmt.Errorf("instance with instanceID %s does not exist", instanceID), statusCode, fmt.Sprintf("instance with instanceID %s does not exist", instanceID))
```
jaroslaw-pieszka, line 56: code duplication

components/kyma-environment-broker/internal/storage/postsql/read.go
```diff
@@ -605,7 +605,7 @@ func (r readSession) GetOperationStatsForOrchestration(orchestrationID string) (
 
 func (r readSession) GetInstanceStats() ([]dbmodel.InstanceByGlobalAccountIDStatEntry, error) {
 	var rows []dbmodel.InstanceByGlobalAccountIDStatEntry
-	_, err := r.session.SelectBySql(fmt.Sprintf("select global_account_id, count(*) as total from %s group by global_account_id",
+	_, err := r.session.SelectBySql(fmt.Sprintf("select global_account_id, count(*) as total from %s where deleted_at = '0001-01-01T00:00:00.000Z' group by global_account_id",
```
jaroslaw-pieszka, line 608: This expression is format dependable. This is risky. Is there no other way to query?

components/kyma-environment-broker/internal/broker/instance_get_test.go
```diff
@@ -90,6 +90,42 @@ func TestGetEndpoint_GetProvisioningInstance(t *testing.T) {
...
+	err = st.Instances().Insert(instance)
+	require.NoError(t, err)
+
+	svc := broker.NewGetInstance(cfg, st.Instances(), st.Operations(), logrus.New())
+
+	// when
+	_, err = svc.GetInstance(context.Background(), instanceID, domain.FetchInstanceDetails{})
+
+	// then
+	assert.IsType(t, err, &apiresponses.FailureResponse{}, "Get returned error of unexpected type")
```
jaroslaw-pieszka, line 124: Kind of misleading message. It is not so obvious that `Get` means GET or Get request.

jaroslaw-pieszka: And why we got here `unexpected type`?

components/kyma-environment-broker/internal/storage/postsql/read.go
MarekMichali: This is the default value for deleted_at in the database. I couldn't think of a better solution, maybe you have an idea?

components/kyma-environment-broker/internal/broker/instance_get_test.go
```diff
@@ -90,6 +90,42 @@ func TestGetEndpoint_GetProvisioningInstance(t *testing.T) {
 	assert.Len(t, response.Metadata.Labels, 2)
 }
 
+func TestGetEndpoint_GetInstanceWhereDeletedAtIsNotZero(t *testing.T) {
```
jaroslaw-pieszka, line 93: In this test we test NOT GETTING the instance.

MarekMichali: My bad. I will rename it.

components/kyma-environment-broker/internal/broker/instance_get_test.go
MarekMichali: The messages come from `TestGetEndpoint_GetNonExistingInstance` since we expect similar result in both tests that the instance doesn't exist.

components/kyma-environment-broker/internal/storage/postsql/read.go
jaroslaw-pieszka: Is there an index containing deleted_at?

MarekMichali: I don't see any declarations in the schema.

jaroslaw-pieszka: It seems index is not necessary.

components/kyma-environment-broker/internal/broker/instance_get_test.go
jaroslaw-pieszka: Got it.

components/kyma-environment-broker/internal/process/deprovisioning/remove_instance_step.go
```diff
@@ -103,3 +109,22 @@ func (s RemoveInstanceStep) removeInstancePermanently(instanceID string, log log
 
 	return 0
 }
+
+func (s RemoveInstanceStep) markInstanceNeedsRetrySomeSteps(instanceID string, log logrus.FieldLogger) time.Duration {
+	backoff := time.Second
+
+	instance, err := s.instanceStorage.GetByID(instanceID)
```
piotrmiskiewicz, line 116: if the instance is not existing (an edge case) - let's log a warning and leave the step without backoff

piotrmiskiewicz (approved):

PR #2459, opened 2023-02-03
Continuation of Cleanup Despite Errors & Adding Removal of Unused Gardener Shoots

ralikio:
**Description**

The following PR fixes 2 bugs of Environments Cleanup job. When the got executed every time instance is not found in KEB's database then https://github.com/kyma-project/control-plane/blob/2b6ba638e622e09363bf996b7c294a91bc8de60d/components/kyma-environment-broker/internal/environmentscleanup/service.go#L139 stops the further cleanup. This causes stack of uncleaned cluster that are not used anymore. The second issue is that the job is not touching clusters created on Gardener by hand. The PR fixes that by deleting such shoots explicitly.

- Not stopping job execution when KEB instance is not found in its database.
- Removing shoots from Gardener when when no account annotations found.

**Related issue(s)**


components/kyma-environment-broker/internal/environmentscleanup/service.go
```diff
@@ -69,78 +71,112 @@ func (s *Service) Run() error {
 }
 
 func (s *Service) PerformCleanup() error {
-
-	staleShoots, err := s.getStaleShoots(s.LabelSelector)
+	runtimesToDelete, shootsToDelete, err := s.getStaleRuntimesByShoots(s.LabelSelector)
+	// runtimesToDelete, _, err := s.getStaleRuntimesByShoots(s.LabelSelector)
```
jaroslaw-pieszka, line 75: Why do we keep commented out code?

components/kyma-environment-broker/internal/environmentscleanup/service.go
```diff
@@ -69,78 +71,112 @@ func (s *Service) Run() error {
...
+		s.logger.Error(fmt.Errorf("while cleaning runtimes: %w", err))
+		return err
+	}
 
-	s.logger.Infof("Runtimes to process: %+v\n", runtimesToDelete)
+	return s.cleanupShoots(shootsToDelete)
+}
 
-	if len(runtimesToDelete) == 0 {
+func (s *Service) cleanupRuntimes(runtimes []runtime) error {
```
jaroslaw-pieszka, line 89: So only reason to extract cleanupRuntime is to check if len > 0?

components/kyma-environment-broker/internal/environmentscleanup/service.go
```diff
@@ -69,78 +71,112 @@ func (s *Service) Run() error {
...
-		if shootAge.Hours() >= s.MaxShootAge.Hours() {
-			log.Infof("Shoot %q is older than %f hours with age: %f hours", shoot.GetName(), s.MaxShootAge.Hours(), shootAge.Hours())
+		if shootAge.Hours() < s.MaxShootAge.Hours() {
+			continue
+		}
+
+		log.Infof("Shoot %q is older than %f hours with age: %f hours", shoot.GetName(), s.MaxShootAge.Hours(), shootAge.Hours())
+		staleRuntime, err := s.shootToRuntime(shoot)
+		if err != nil {
+			s.logger.Info("found a shoot without accounts references")
```
jaroslaw-pieszka, line 139: or without runtime-id annotation...

components/kyma-environment-broker/internal/environmentscleanup/service.go
```diff
@@ -69,78 +71,112 @@ func (s *Service) Run() error {
...
 func (s *Service) cleanUp(runtimesToDelete []runtime) error {
-	kebInstancesToDelete, err := s.getInstancesForRuntimes(runtimesToDelete)
-	if err != nil {
-		err = fmt.Errorf("while getting instance IDs for Runtimes: %w", err)
-		s.logger.Error(err)
-		return err
+	kebInstancesToDelete := make([]internal.Instance, 0)
+
+	for _, rt := range runtimesToDelete {
+		instance, err := s.getInstancesForRuntimes([]runtime{rt})
```
jaroslaw-pieszka, line 172: Why we use this function in so limited way selecting instances one by one instead of selection the set.

jaroslaw-pieszka: Implementation get array of runtime ids and selects as follows:
```
		Select("*").
		From(InstancesTableName).
		Where("runtime_id IN ?", runtimeIdList).
		LoadOne(&instances)
```
so if not all runtime ids are found we should not encounter error but get shorter list.

ralikio: This is the main reason for the change. If only one runtime does not have an instance then an error is returned and the execution fail.

components/kyma-environment-broker/internal/environmentscleanup/service.go
ralikio: I changed it to "... without kcp references"

components/kyma-environment-broker/internal/environmentscleanup/service.go
ralikio: Corrected to check the NOT FOUND error explicitly.

jaroslaw-pieszka (approved):

ebensom (approved):

PR #2468, opened 2023-02-09
App Builder

ralikio:
**Description**

The way in which project initialises go app makes it hard to reuse that code. For example, if there is a need to create a new job with the same configuration it is not possible to be done without extensive copying. The following PR introduces a app_builder that shares common configuration of Environments Cleanup Job.

### DIFF
--- a/components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
+++ b/components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/dlmiddlecote/sqlstats"
+	"github.com/gocraft/dbr"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/common/gardener"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/common/runtime"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/broker"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/environmentscleanup"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/events"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/provisioner"
+	"github.com/kyma-project/control-plane/components/kyma-environment-broker/internal/storage"
+	"github.com/kyma-project/control-plane/components/schema-migrator/cleaner"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
+	"github.com/vrischmann/envconfig"
+	"golang.org/x/oauth2/clientcredentials"
+	"k8s.io/client-go/dynamic"
+)
+
+type config struct {
+	MaxAgeHours   time.Duration `envconfig:"default=24h"`
+	LabelSelector string        `envconfig:"default=owner.do-not-delete"`
+	Gardener      gardener.Config
+	Database      storage.Config
+	Broker        broker.ClientConfig
+	Provisioner   provisionerConfig
+}
+
+type AppBuilder struct {
+	cfg                 config
+	shootClient         dynamic.ResourceInterface
+	brokerRuntimeClient environmentscleanup.BrokerRuntimesClient
+	db                  storage.BrokerStorage
+	conn                *dbr.Connection
+	logger              *logrus.Logger
+
+	brokerClient      *broker.Client
+	provisionerClient provisioner.Client
+}
+
+type Job interface {
+	Run() error
+}
+
+func NewAppBuilder() AppBuilder {
+	cfg := config{}
+	err := envconfig.InitWithPrefix(&cfg, "APP")
+	if err != nil {
+		FatalOnError(fmt.Errorf("while loading cleanup config: %w", err))
+	}
+
+	return AppBuilder{}
+}
+
+func (b *AppBuilder) withGardenerClient() {
+	clusterCfg, err := gardener.NewGardenerClusterConfig(b.cfg.Gardener.KubeconfigPath)
+	if err != nil {
+		FatalOnError(fmt.Errorf("while creating Gardener cluster config: %w", err))
+	}
+	cli, err := dynamic.NewForConfig(clusterCfg)
+	if err != nil {
+		FatalOnError(fmt.Errorf("while creating Gardener client: %w", err))
+	}
+	gardenerNamespace := fmt.Sprintf("garden-%s", b.cfg.Gardener.Project)
+	b.shootClient = cli.Resource(gardener.ShootResource).Namespace(gardenerNamespace)
+}
+
+func (b *AppBuilder) withBrokerClient() {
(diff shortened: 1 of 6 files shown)


components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
```diff
@@ -0,0 +1,124 @@
+package main
```
wozniakjan, line 1: How do you feel about moving this to `components/kyma-environment-broker/common` or `components/kyma-environment-broker/internal`? The `cmd` dir is usually reserved for executables and this feels more like a library package.
https://github.com/golang-standards/project-layout/tree/509e1be/cmd

wozniakjan (commented):
some refactoring is probably incomplete, fails to build with:
```
cmd/environmentscleanup/main.go:9:13: undefined: NewAppBuilder
cmd/environmentscleanup/main.go:20:3: undefined: FatalOnError
error building image: error building stage: failed to execute command: waiting for process to exit: exit status 1
build encountered error: exit status 1
```

components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		FatalOnError(err)
+	}
+	dbStatsCollector := sqlstats.NewStatsCollector("broker", b.conn)
+	prometheus.MustRegister(dbStatsCollector)
+
+	b.logger = log.New()
+}
+
+func (b *AppBuilder) Cleanup() {
```
jaroslaw-pieszka, line 101: Is this used anywhere? I could not find the invocation.
And why this is a part of builder? Do we have possibility to create app with cleanup (more like withCleanupAfterJobIsRun) or this is mandatory and must be called after each job is executed.

components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
```diff
@@ -0,0 +1,124 @@
...
+		FatalOnError(fmt.Errorf("while creating Gardener client: %w", err))
+	}
+	gardenerNamespace := fmt.Sprintf("garden-%s", b.cfg.Gardener.Project)
+	b.shootClient = cli.Resource(gardener.ShootResource).Namespace(gardenerNamespace)
+}
+
+func (b *AppBuilder) withBrokerClient() {
+	ctx := context.Background()
+	b.brokerClient = broker.NewClient(ctx, b.cfg.Broker)
+	b.provisionerClient = provisioner.NewProvisionerClient(b.cfg.Provisioner.URL, b.cfg.Provisioner.QueryDumping)
```
jaroslaw-pieszka, line 75: If we separate clients in builder calls why not `withProvisionerClient`. Or single method `withBrokerAndProvisionerClients`.

components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
```diff
@@ -0,0 +1,124 @@
...
+	cipher := storage.NewEncrypter(b.cfg.Database.SecretKey)
+	var err error
+	b.db, b.conn, err = storage.NewFromConfig(b.cfg.Database, events.Config{}, cipher, log.WithField("service", "storage"))
+	if err != nil {
+		FatalOnError(err)
+	}
+	dbStatsCollector := sqlstats.NewStatsCollector("broker", b.conn)
+	prometheus.MustRegister(dbStatsCollector)
+
+	b.logger = log.New()
```
jaroslaw-pieszka, line 98: Shouldn't we separate logger setup? Setting logger as a part of adding storage seems odd.

components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
```diff
@@ -0,0 +1,124 @@
...
+	clusterCfg, err := gardener.NewGardenerClusterConfig(b.cfg.Gardener.KubeconfigPath)
+	if err != nil {
+		FatalOnError(fmt.Errorf("while creating Gardener cluster config: %w", err))
+	}
+	cli, err := dynamic.NewForConfig(clusterCfg)
+	if err != nil {
+		FatalOnError(fmt.Errorf("while creating Gardener client: %w", err))
+	}
+	gardenerNamespace := fmt.Sprintf("garden-%s", b.cfg.Gardener.Project)
+	b.shootClient = cli.Resource(gardener.ShootResource).Namespace(gardenerNamespace)
```
jaroslaw-pieszka, line 69: `shootClient` vs `GardenerClient`. It is inconsistent.

components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
```diff
@@ -0,0 +1,124 @@
...
+func (b *AppBuilder) Create() Job {
+	return environmentscleanup.NewService(
+		b.shootClient,
+		b.brokerClient,
+		b.provisionerClient,
+		b.db.Instances(),
+		b.logger,
+		b.cfg.MaxAgeHours,
+		b.cfg.LabelSelector,
+	)
```
jaroslaw-pieszka, line 123: Why not `withGardenerClient`, `withBrokerClient`, `withProvisionerClient`, `withStorage`, 'withLogger(logger)`, 'withConfiguration(config)` - more or less.

components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// do not use defer, close must be done before halting
+	err = cleaner.Halt()
+	if err != nil {
+		FatalOnError(err)
+	}
+}
+
+func (b *AppBuilder) Create() Job {
```
jaroslaw-pieszka, line 114: `AppBuilder.Create()` creates `Job`. Wouldn't it be better to use `JobBuilder`? Or Rename `Job` to `App`.

components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
ralikio: The method closes db connection and stops every sidecar containers so that the Job has a chance to succeed. I have added missing invocation. Thank you!

components/kyma-environment-broker/cmd/environmentscleanup/app_builder.go
ralikio: I have renamed it to shorter version - Bulider. What do you think about it?

jaroslaw-pieszka: It is always better to have consistent naming, this is obvious, even if the code would be maintained by chatGPT in the near future. So you have file `app_builder` with function `NewAppBuilder` that returns `Builder`. And builder builds `Job`.

ralikio: In that case I would leave it as `app_builder` since it is a job only when used in the K8S context. Other than that it is just a program that runs cleanup logic.

jaroslaw-pieszka: So you want to stick with the interface named `Job`?

jaroslaw-pieszka (approved):

Issue #2512, opened 2023-02-23
KEB: soft delete SKR Service Instances and Service Bindings when subaccount deletion event sent to us


PK85:
**Description**
When subaccount event arrives from BTP, KEB starts Kyma deprovisioning and  btp-operator fails to be uninstalled cause instances and bindings have finalizers. In such a scenario we need to soft delete instances and bindings.

**AC**
- [x] 1) Provide something like```kymaContext``` root element in the payload when KEB is called and put information that ```subaccountDeletion=true```. Then when it is present, do a soft delete. 
- [x] 2) Adjust subaccount deletion job to send such a kymaContext to KEB
- [x] 3) Describe in the KEB documentation this adjustment.

wozniakjan:
> 1) Provide something likekymaContext root element in the payload when KEB is called and put information that subaccountDeletion=true. Then when it is present, do a soft delete.￼
> 2) Adjust subaccount deletion job to send such a kymaContext to KEB

the instance `DELETE` endpoint call is defined by OSB API. So it has fixed parameters and no payload body. But there appears to be `force` parameter defined by OSB API which is currently unused by KEB at all, so I would like to propose https://github.com/kyma-project/control-plane/pull/2514 using `DELETE` on `/instance/${instance_id}?force=true` for this case.

PK85:
@wozniakjan lets discuss that solution tomorrow, maybe including maper as well

wozniakjan:
@PK85 here is an alternative proposal using HTTP header to identify the caller: https://github.com/kyma-project/control-plane/pull/2520. I decided to go for `User-Agent` but am happy to change this to KEB special-purpose header if you think that would work better.

PR #2514, opened 2023-02-23
KEB cleanup btp-operator resources on instance force delete

wozniakjan:
**Description**

Instances and bindings from btp-operator can be problematic for subaccount cleanup causing extended delays and many false errors during the deprovisioning. This PR introduces usage of OSB API `force` parameter for instance `DELETE` endpoint call which is propagated to the deprovisioning operation and picked up by BTP Operator cleanup step ensuring resources are deleted and finalizers from them removed. The external resources are going to be removed by other SAP service providers because this is triggered only when the whole subaccount is triggered for deletion.

**Related issue(s)**

see: https://github.com/kyma-project/control-plane/issues/2512

### DIFF
--- a/components/kyma-environment-broker/internal/cis/account_cleanup.go
+++ b/components/kyma-environment-broker/internal/cis/account_cleanup.go
@@ -16,7 +16,7 @@ type CisClient interface {
 
 //go:generate mockery --name=BrokerClient --output=automock
 type BrokerClient interface {
-	Deprovision(instance internal.Instance) (string, error)
+	Deprovision(instance internal.Instance, force bool) (string, error)
 }
 
 type SubAccountCleanupService struct {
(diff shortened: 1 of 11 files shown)


components/kyma-environment-broker/internal/cis/account_cleanup.go
```diff
@@ -77,7 +77,7 @@ func (ac *SubAccountCleanupService) executeDeprovisioning(subaccounts []string,
 	}
 
 	for _, instance := range instances {
-		operation, err := ac.brokerClient.Deprovision(instance)
+		operation, err := ac.brokerClient.Deprovision(instance, true)
```
wozniakjan, line 80: > 2) Adjust subaccount deletion job to send such a kymaContext to KEB

@PK85 is this the subaccount cleanup you mentioned in https://github.com/kyma-project/control-plane/issues/2512#issue-1596917634 ?

wozniakjan:
closing in favor of https://github.com/kyma-project/control-plane/pull/2520

PR #2516, opened 2023-02-24
Additional test case for zombie instances

jaroslaw-pieszka:
**Description**
`Zombie` means instance deleted logically but not physically, under the hood having `deletedAt` > 0.
Such instances are invisible unless are visible in some contexts/calls.

- Test case checking if GetInstance **do not** return `zombie` instance

**Related issue(s)**


MarekMichali:
Isn't it the same case as `TestGetEndpoint_DoNotReturnInstanceWhereDeletedAtIsNotZero`?

MarekMichali:
Closed as a duplicate.

Issue #2518, opened 2023-02-24
Invalid Cals to Non-idempotent AVS


ralikio:
**Description**

Steps in KEB are designed to run multiple times. The step https://github.com/kyma-project/control-plane/blob/59dbec7826e8d48fcd948b6d31dd7cb7c2b7df38/components/kyma-environment-broker/internal/process/deprovisioning/avs_evaluations.go#L40 does not follow this conventions. Because of the compound "if" condition if any of AVSExternalEvaluationDeleted or AVSInternalEvaluationDeleted is not set because of additional logic or failure, then the whole process repeats calling non-idempotent AVS again.

MarekMichali:
See #2519

Other scenarios, such as a lack of connection to the DB, should be checked to see if they cause the step to be false repeated.

MarekMichali:
Upon further investigation, the original logic is correct, but requires some refactoring to clean up redundant code that is confusing. A separate ticket will be opened.

PR #2519, opened 2023-02-24
Fix deprovision of avs evaluations for trial/freemium plan

MarekMichali:
**Description**

If the `De-provision_AVS_Evaluations` step was run again for the trial/freemium plan, even if the internal evaluation was removed successfully in the previous run, the step was not skipped.

- skip `De-provision_AVS_Evaluations` if internal evaluation have been deleted for trial/freemium plan

**Related issue(s)**

See also #2518

### DIFF
--- a/components/kyma-environment-broker/internal/process/deprovisioning/avs_evaluations.go
+++ b/components/kyma-environment-broker/internal/process/deprovisioning/avs_evaluations.go
@@ -37,9 +37,15 @@ func (ars *AvsEvaluationRemovalStep) Name() string {
 
 func (ars *AvsEvaluationRemovalStep) Run(operation internal.Operation, logger logrus.FieldLogger) (internal.Operation, time.Duration, error) {
 	logger.Infof("Avs lifecycle %+v", operation.Avs)
-	if operation.Avs.AVSExternalEvaluationDeleted && operation.Avs.AVSInternalEvaluationDeleted {
-		logger.Infof("Both internal and external evaluations have been deleted")
-		return operation, 0, nil
+	if operation.Avs.AVSInternalEvaluationDeleted {
+		if operation.Avs.AVSExternalEvaluationDeleted {
+			logger.Infof("Both internal and external evaluations have been deleted")
+			return operation, 0, nil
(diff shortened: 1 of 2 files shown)


components/kyma-environment-broker/internal/process/deprovisioning/avs_evaluations.go
```diff
@@ -37,9 +37,15 @@ func (ars *AvsEvaluationRemovalStep) Name() string {
 
 func (ars *AvsEvaluationRemovalStep) Run(operation internal.Operation, logger logrus.FieldLogger) (internal.Operation, time.Duration, error) {
 	logger.Infof("Avs lifecycle %+v", operation.Avs)
-	if operation.Avs.AVSExternalEvaluationDeleted && operation.Avs.AVSInternalEvaluationDeleted {
-		logger.Infof("Both internal and external evaluations have been deleted")
-		return operation, 0, nil
+	if operation.Avs.AVSInternalEvaluationDeleted {
```
jaroslaw-pieszka, line 40: How about two functions:

1. logic for trial or freemium
2. logic for remaining plans

 Wouldn't it be more readable and bring us closer to extraction on step level (possibly in the future).

jaroslaw-pieszka: Or much simpler.
Each call of DeleteAVSEvaluation is guarded with flag checking.
We wouldn't increase the complexity, nesting level and we would avoid code duplication which is present in your solution.

resources/kcp/charts/kyma-environment-broker/values.yaml
```diff
@@ -2,7 +2,7 @@ global:
   images:
     kyma_environment_broker:
       dir:
-      version: "PR-2513"
+      version: "PR-2519"
```
jaroslaw-pieszka, line 5: No test cases added.

MarekMichali:
Upon further investigation, the original logic is correct, but requires some refactoring to clean up redundant code that is confusing. A separate ticket will be opened.

PR #2555, opened 2023-03-07
Fix Kyma deprovisioning test

MarekMichali:
**Description**

`TestKymaDeprovision` logged an error in the `EDP_Deregistration` step during deprovisioning, and it shouldn't have. This was due to different SubAccountID in the instance and the provisioning operation. In addition, the mocked EDPClient was also missing the `MaasConsumerServicePlan`.

- Fix the SubAccountID mismatch
- Add `MaasConsumerServicePlan` to the mocked EDPClient

**Related issue(s)**

### DIFF
--- a/resources/kcp/charts/kyma-environment-broker/values.yaml
+++ b/resources/kcp/charts/kyma-environment-broker/values.yaml
@@ -2,7 +2,7 @@ global:
   images:
     kyma_environment_broker:
       dir:
-      version: "PR-2546"
+      version: "PR-2555"
     kyma_environments_cleanup_job:
       dir:
       version: "PR-2540"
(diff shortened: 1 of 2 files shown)


resources/kcp/charts/kyma-environment-broker/values.yaml
```diff
@@ -2,7 +2,7 @@ global:
   images:
     kyma_environment_broker:
       dir:
-      version: "PR-2546"
+      version: "PR-2555"
```
wozniakjan, line 5: nit: given it's change exclusively in the tests, there is no need to bump the image (but it doesn't hurt either, `pull-keb-build` pipeline produced the image)

wozniakjan (approved):

PR #2570, opened 2023-03-10
Add unit tests for redeprovisioning

MarekMichali:
**Description**

- Add unit test for successfully redeprovision
- Add unit test for failed redeprovision

**Related issue(s)**

See also #2406

### DIFF
--- a/components/kyma-environment-broker/internal/provisioner/fake_client.go
+++ b/components/kyma-environment-broker/internal/provisioner/fake_client.go
@@ -82,6 +82,19 @@ func (c *FakeClient) FindOperationByRuntimeIDAndType(runtimeID string, operation
 	return schema.OperationStatus{}
 }
 
+func (c *FakeClient) FindLastOperationByRuntimeIDAndType(runtimeID string, operationType schema.OperationType) schema.OperationStatus {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	var opStatus schema.OperationStatus
+	for _, status := range c.operations {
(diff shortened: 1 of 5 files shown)


components/kyma-environment-broker/internal/provisioner/fake_client.go
```diff
@@ -82,6 +82,19 @@ func (c *FakeClient) FindOperationByRuntimeIDAndType(runtimeID string, operation
...
 }
 
+func (c *FakeClient) FindLastOperationByRuntimeIDAndType(runtimeID string, operationType schema.OperationType) schema.OperationStatus {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	var opStatus schema.OperationStatus
+	for _, status := range c.operations {
+		if *status.RuntimeID == runtimeID && status.Operation == operationType {
+			opStatus = status
```
wozniakjan, line 92: nit: if iterated from the end, you could break the cycle here and save a few CPU instructions :)

resources/kcp/charts/kyma-environment-broker/values.yaml
```diff
@@ -2,7 +2,7 @@ global:
   images:
     kyma_environment_broker:
       dir:
-      version: "PR-2555"
+      version: "PR-2570"
```
wozniakjan, line 5: nit: technically no need to bump for tests only

wozniakjan (approved):
/approve
/lgtm

MarekMichali:
We can't assume that the last operation in the map will be in fact the most recent. We need to make sure in another way.

PR #2578, opened 2023-03-13
feature: periodic resolving orchestration runtime

AdrianPei:
**Description**

-  refactored for periodic resolution of runtime targets into runtime operations
- .fix some mismatching data structure between KEB and CNS

### DIFF
--- a/components/kyma-environment-broker/internal/orchestration/manager/manager.go
+++ b/components/kyma-environment-broker/internal/orchestration/manager/manager.go
@@ -26,8 +26,12 @@ import (
 type OperationFactory interface {
 	NewOperation(o internal.Orchestration, r orchestration.Runtime, i internal.Instance, state domain.LastOperationState) (orchestration.RuntimeOperation, error)
 	ResumeOperations(orchestrationID string) ([]orchestration.RuntimeOperation, error)
+	CancelOperation(orchestrationID string, runtimeID string) error
 	CancelOperations(orchestrationID string) error
 	RetryOperations(operationIDs []string) ([]orchestration.RuntimeOperation, error)
+	QueryOperation(orchestrationID string, r orchestration.Runtime) (bool, orchestration.RuntimeOperation, error)
+	QueryOperations(orchestrationID string) ([]orchestration.RuntimeOperation, error)
+	NotifyOperation(orchestrationID string, runtimeID string, oState string, notifyState orchestration.NotificationStateType) error
 }
 
 type orchestrationManager struct {
@@ -67,14 +71,24 @@ func (m *orchestrationManager) Execute(orchestrationID string) (time.Duration, e
 		return m.failOrchestration(o, fmt.Errorf("failed to get orchestration: %w", err))
 	}
 
-	maintenancePolicy, err := m.getMaintenancePolicy()
+	operations, runtimeNums, err := m.waitForStart(o)
 	if err != nil {
-		m.log.Warnf("while getting maintenance policy: %s", err)
+		m.failOrchestration(o, fmt.Errorf("failed while waiting start for operations: %w", err))
 	}
 
-	operations, err := m.resolveOperations(o, maintenancePolicy)
-	if err != nil {
-		return m.failOrchestration(o, fmt.Errorf("failed to resolve operations: %w", err))
+	if o.Parameters.Kyma == nil || o.Parameters.Kyma.Version == "" {
+		o.Parameters.Kyma = &orchestration.KymaParameters{Version: m.kymaVersion}
+	}
+	if o.Parameters.Kubernetes == nil || o.Parameters.Kubernetes.KubernetesVersion == "" {
+		o.Parameters.Kubernetes = &orchestration.KubernetesParameters{KubernetesVersion: m.kubernetesVersion}
+	}
+
+	if o.State == orchestration.Pending || o.State == orchestration.Retrying {
+		if runtimeNums != 0 {
+			o.State = orchestration.InProgress
+		} else {
+			o.State = orchestration.Succeeded
+		}
 	}
 
 	err = m.orchestrationStorage.Update(*o)
@@ -89,15 +103,6 @@ func (m *orchestrationManager) Execute(orchestrationID string) (time.Duration, e
 	}
 
 	strategy := m.resolveStrategy(o.Parameters.Strategy.Type, m.executor, logger)
-
-	// ctreate notification after orchestration resolved
-	if !m.bundleBuilder.DisabledCheck() {
-		err := m.sendNotificationCreate(o, operations)
-		//currently notification error can only be temporary error
-		if err != nil && kebError.IsTemporaryError(err) {
-			return 5 * time.Second, nil
-		}
-	}
 
 	execID, err := strategy.Execute(operations, o.Parameters.Strategy)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -199,9 +188,16 @@ func (m *orchestrationManager) NewOperationForPendingRetrying(o *internal.Orches
...
 		if err != nil {
-			return nil, o, len(runtimes), fmt.Errorf("while creating new operation for runtime id %q: %w", r.RuntimeID, err)
+			return nil, o, len(runtimes), fmt.Errorf("while quering operation for runtime id %q: %w", r.RuntimeID, err)
+		}
+
+		if !exist {
+			op, err = m.factory.NewOperation(*o, r, *inst, orchestration.Pending)
+			if err != nil {
+				return nil, o, len(runtimes), fmt.Errorf("while creating new operation for runtime id %q: %w", r.RuntimeID, err)
+			}
```
ebensom, line 200: This querying and checking for existence should be done when orchestration is in `Pending` state (i.e. it's not `Retrying`).
Also, it should be the first thing when looping through the runtime operations, and the above if/else code should also not be executed (skipped when exists, no need to resolve maintenance window time and get instance).

components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -593,3 +644,45 @@ func updateRetryingDescription(desc string, newDesc string) string {
...
+		if err != nil {
+			_, err = m.failOrchestration(o, fmt.Errorf("failed to resolve operations: %w", err))
+			return true, err
+		}
+
+		//send notification for each operation which doesn't have one
+		if !m.bundleBuilder.DisabledCheck() && o.State == orchestration.Pending {
+			for _, operation := range operations {
+				if operation.NotificationState == "" {
+					err = m.sendNotificationCreate(o, operation)
```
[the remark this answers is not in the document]
AdrianPei: refactor send/cancel notifications to use bundle

components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -593,3 +644,45 @@ func updateRetryingDescription(desc string, newDesc string) string {
...
+	if err != nil {
+		m.log.Warnf("while getting maintenance policy: %s", err)
+	}
+
+	//polling every 5 min until ochestration start
+	pollingInterval := 5 * time.Minute
+	var operations []orchestration.RuntimeOperation
+	err = wait.PollImmediateInfinite(pollingInterval, func() (bool, error) {
+		//resolve operations, cancel non existent ones
+		operations, err := m.resolveOperations(o, maintenancePolicy)
```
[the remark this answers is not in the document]
AdrianPei: fixed

components/kyma-environment-broker/internal/orchestration/manager/manager.go
AdrianPei: fixed as suggested

components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -67,14 +71,9 @@ func (m *orchestrationManager) Execute(orchestrationID string) (time.Duration, e
...
-	maintenancePolicy, err := m.getMaintenancePolicy()
-	if err != nil {
-		m.log.Warnf("while getting maintenance policy: %s", err)
-	}
-
-	operations, err := m.resolveOperations(o, maintenancePolicy)
+	operations, err := m.waitForStart(o)
 	if err != nil {
-		return m.failOrchestration(o, fmt.Errorf("failed to resolve operations: %w", err))
+		return 0, err
```
ebensom, line 76: Shouldn't we just log the error, and  allow retrying the `waitForStart` in temporary error?

```suggestion
		return m.pollingInterval, nil
```

components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -593,3 +657,55 @@ func updateRetryingDescription(desc string, newDesc string) string {
...
+	}
+
+	//polling every 5 min until ochestration start
+	pollingInterval := 5 * time.Minute
+	var operations, unnotified_operations []orchestration.RuntimeOperation
+	err = wait.PollImmediateInfinite(pollingInterval, func() (bool, error) {
+		//resolve operations, cancel non existent ones
+		operations, err = m.resolveOperations(o, maintenancePolicy)
+		if err != nil {
+			_, err = m.failOrchestration(o, fmt.Errorf("failed to resolve operations: %w", err))
```
ebensom, line 674: I think we shouldn't fail the orchestration if there is a (temporary) resolution failure

```suggestion
```

components/kyma-environment-broker/internal/orchestration/manager/manager.go
```diff
@@ -223,6 +224,43 @@ func (m *orchestrationManager) NewOperationForPendingRetrying(o *internal.Orches
 	return result, o, len(fileterRuntimes), nil
```
ebensom, line 224: These should be moved out from `NewOperationForPendingRetrying`, so that they do n't get executed when in `waitForStart`. They should be included in manager.Execute(), after `waitForStart`.


```
if o.Parameters.Kyma == nil || o.Parameters.Kyma.Version == "" {
		o.Parameters.Kyma = &orchestration.KymaParameters{Version: m.kymaVersion}
	}
	if o.Parameters.Kubernetes == nil || o.Parameters.Kubernetes.KubernetesVersion == "" {
		o.Parameters.Kubernetes = &orchestration.KubernetesParameters{KubernetesVersion: m.kubernetesVersion}
	}
	if len(fileterRuntimes) != 0 {
		o.State = orchestration.InProgress
	} else {
		o.State = orchestration.Succeeded
	}
```

AdrianPei: This part of codes is also used in waitForCompletion() polling loop, judging if there is still runtime left in each loop by len(fileterRuntimes). Do you think we should refactor waitForCompletion() as well?

AdrianPei: This part of codes is also used in waitForCompletion() polling loop, judging if there is still runtime left in each loop by len(fileterRuntimes). Do you think we should refactor waitForCompletion() as well?

ebensom: This part is not relevant in waitforCompletion(), feel free to extract it and move it to `manager.Execute()`, after `waitForStart`

components/kyma-environment-broker/internal/orchestration/manager/manager.go
AdrianPei: I kept `return m.failOrchestration(o, fmt.Errorf("failed to resolve operations: %w", err))` instead. A failOrchestration() needs be to triggered here to retry failed orchestration. Also when orchestration is empty, failOrchestration() is needed so that o.IsFinished() can quit orchestration.

components/kyma-environment-broker/internal/orchestration/manager/manager.go
AdrianPei: moved first part of them to `manager.Execute()` and kept

```
if len(fileterRuntimes) != 0 {
	o.State = orchestration.InProgress
} else {
	o.State = orchestration.Succeeded
}
```
did many tests and confirmed that when retrying failed orchestration, orchestration state can't be switched correctly without this part.

ebensom: But if we assign `o.State` something different than `Pending` here, then subsequent calls in waitForStart -> resolveOperations loop would not hit the correct branch for Pending:

`if o.State == orchestration.Pending {`

Why failed orchestration retrying didn't work if you moved this part right after the others in `Execute()`, like this:

```
	if o.Parameters.Kyma == nil || o.Parameters.Kyma.Version == "" {
		o.Parameters.Kyma = &orchestration.KymaParameters{Version: m.kymaVersion}
	}
	if o.Parameters.Kubernetes == nil || o.Parameters.Kubernetes.KubernetesVersion == "" {
		o.Parameters.Kubernetes = &orchestration.KubernetesParameters{KubernetesVersion: m.kubernetesVersion}
	}
	if len(operations) != 0 {
		o.State = orchestration.InProgress
	} else {
		o.State = orchestration.Succeeded
	}
```

ebensom (approved):

PR #2598, opened 2023-03-16
Correct runtime state when op state is pending

vvxxvvxx:
**Description**
When operation state is `orchestration.Pending`, in `adjustRuntimeState` for runtimeDTO, the state will be converted to `StateSucceeded` which is wrong. this can happened to [Deprovisioning](https://github.com/kyma-project/control-plane/blob/a5574a6d5e158d298b18ea98a497c6ee38cd9074/components/kyma-environment-broker/internal/model.go#L512), [Suspension](https://github.com/kyma-project/control-plane/blob/a5574a6d5e158d298b18ea98a497c6ee38cd9074/components/kyma-environment-broker/internal/model.go#L564) and [Update](https://github.com/kyma-project/control-plane/blob/a5574a6d5e158d298b18ea98a497c6ee38cd9074/components/kyma-environment-broker/internal/model.go#L530) operations as they have Pending state when their new operations are created in the beginning.
https://github.com/kyma-project/control-plane/blob/a5574a6d5e158d298b18ea98a497c6ee38cd9074/components/kyma-environment-broker/internal/runtime/converter.go#L213

We have [this issue](https://github.tools.sap/kyma/backlog/issues/3723) occurred because of this bug causing mop-agent got the wrong `StateSucceeded` state when the runtime is in progress of provisioning but deprovisioning is on hold (state is pending for deprovisioning as the runtime is still in progress of provisioning)

and if runtime state is orchestration.Pending, then the converter.go [adjustRuntimeState](https://github.com/ebensom/control-plane/blob/99e642152f8f2e3f93e984a5e7e6d7b9c8e9b2d1/components/kyma-environment-broker/internal/runtime/converter.go#L188), the dto.Status.State returned will be wrongly set to state `StateSucceeded`.

- No convert to the deprovision operation with `orchestration.Pending` state and suspension operation with `orchestration.Pending` state

**Related issue(s)**


ebensom (commented):
Hi @vvxxvvxx Pending operations should not pop up as a last operation either by dto.LastOperation or any DB query. 
I think there is a regression then in underlying DB lookup code of dto.LastOperation which needs to be looked into.

vvxxvvxx:
> Hi @vvxxvvxx Pending operations should not pop up as a last operation either by dto.LastOperation or any DB query. I think there is a regression then in underlying DB lookup code of dto.LastOperation which needs to be looked into.

Hi Benjamin, seems this runtimeDTO will be reflected in the runtime data (output `kcp rt -c *** -ojson), if we exclude the pending state, then for example, the pending deprovisioning operation will not be shown in the `kcp rt -c *** -ojson` ouput..

what about we add one state `pkg.StatePending` in https://github.com/kyma-project/control-plane/blob/c18b4343c2f0b0a365b34d8b6815bb4808d9cb31/components/kyma-environment-broker/internal/runtime/converter.go#L212 to indicate correctly the corresponding op type is in `"state": "pending"`?

vvxxvvxx:
> Hi @vvxxvvxx Pending operations should not pop up as a last operation either by dto.LastOperation or any DB query. I think there is a regression then in underlying DB lookup code of dto.LastOperation which needs to be looked into.

excluded the pending operation in LastOperation()..

components/kyma-environment-broker/internal/storage/postsql/read.go
```diff
@@ -282,13 +282,15 @@ func (r readSession) GetOperationByTypeAndInstanceID(inID string, opType interna
...
 	typeCondition := dbr.Eq("type", string(opType))
+	stateCondition := dbr.Neq("state", orchestration.Pending)
 	var operations []dbmodel.OperationDTO
 
 	_, err := r.session.
 		Select("*").
 		From(OperationTableName).
 		Where(idCondition).
 		Where(typeCondition).
+		Where(stateCondition).
```
ebensom, line 293: Are you sure it's safe in all calls to not return Pending operations?
It might be OK for the runtimes endpoint case, but introduce big issue in others.

vvxxvvxx: there will be affect to other calls. it's better to not change the underlying db function..
I changed the logic again to only modify the `ApplySuspensionOperations` and `ApplyUnsuspensionOperations` to not convert pending operation to runtime dto. then mop-agent will not get the wrong state for the below scenario..

> when the runtime is in progress of provisioning but deprovisioning is on hold (state is pending for deprovisioning as the runtime is still in progress of provisioning)

jaroslaw-pieszka (approved):

PR #2616, opened 2023-03-22
Introduced Bool Type for Incorrect DB Values

ralikio:
**Description**

If orchestration field `immediate` is persisted as a string value `"true"` in database the request send to "orchestration" endpoint fails on unmarshaling this value to bool of https://github.dev/kyma-project/control-plane/blob/c825f45f79d9d4b1b1fe9f5f9ac54c779e494a1a/components/kyma-environment-broker/internal/storage/dbmodel/orchestration.go#L50. Error example: 
```
while getting orchestrations: json: cannot unmarshal string into Go struct field RetryOperationParameters.retryoperation.immediate of type bool"}
``` 

- Tests that mitigate the orginal problem (see case with `"immediate": "true"`)
- New `stringBoolean` with custom unmarshal that handles bool represented as strings.

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/common/orchestration/dto.go
+++ b/components/kyma-environment-broker/common/orchestration/dto.go
@@ -1,6 +1,8 @@
 package orchestration
 
 import (
+	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/kyma-project/control-plane/components/provisioner/pkg/gqlschema"
@@ -18,9 +20,10 @@ type Parameters struct {
 	NotificationState notificationStateType    `json:"notificationstate,omitempty"`
 	RetryOperation    RetryOperationParameters `json:"retryoperation,omitempty"`
 }
+
 type RetryOperationParameters struct {
-	RetryOperations []string `json:"retryoperations,omitempty"`
-	Immediate       bool     `json:"immediate,omitempty"`
+	RetryOperations []string      `json:"retryoperations,omitempty"`
+	Immediate       stringBoolean `json:"immediate,omitempty"`
 }
 
 type KubernetesParameters struct {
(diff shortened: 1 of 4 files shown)


components/kyma-environment-broker/common/orchestration/dto.go
```diff
@@ -209,3 +212,21 @@ const (
 	NotificationCreated   notificationStateType = "created"
 	NotificationCancelled notificationStateType = "cancelled"
 )
+
+type stringBoolean bool
+
+func (sb *stringBoolean) UnmarshalJSON(data []byte) error {
+	str := string(data)
```
jaroslaw-pieszka, line 219: Why not `string(bytes.Trim(str, `"`))` and fallback to parseBoolean?

components/kyma-environment-broker/common/orchestration/dto.go
```diff
@@ -209,3 +212,21 @@ const (
...
+	if str == `"true"` {
+		*sb = stringBoolean(true)
+	} else if str == `"false"` {
+		*sb = stringBoolean(false)
+	} else {
+		v, err := strconv.ParseBool(string(data))
```
jaroslaw-pieszka, line 225: So we accept:
- "true"
- true
- "false"
- false
Am I correct?

components/kyma-environment-broker/common/orchestration/dto.go
jaroslaw-pieszka, line 1: We fix the issue by relaxing some restrictions.
Have we found the root cause? Which process (and why) is inserting "unorthodox" bool values to json (and database)?
Should we correct the root cause as well?
Did we considered other scenario - correcting the source, then correcting the data in database?

jaroslaw-pieszka: Our current implementation marshals data correctly:
```json
{"targets":{"include":[{"runtimeID":"ebfa15f1-b56b-4f3f-8af8-285289630036"}]},"strategy":{"type":"parallel","schedule":"2023-03-22T14:43:56+01:00","ScheduleTime":"0001-01-01T00:00:00Z","parallel":{"workers":1}},"kubernetes":{"kubernetesVersion":"","machineImage":"","machineImageVersion":""},"kyma":{},"retryoperation":{"immediate":true}}
```

jaroslaw-pieszka (approved):

PR #2678, opened 2023-04-17
Dashbaord Redirect Assertion - Status Correction

ralikio:
**Description**

- Change http status from redirection (302) to successful response (200) since there is no more redirection happening. 
- Go Mod migration.

**Related issue(s)**


tests/e2e/provisioning/README.md
```diff
@@ -91,3 +91,4 @@ You can configure the test execution by using the following environment variable
 | **APP_CLEANUP_PHASE** | Specifies if the test executes the cleanup phase. | `false` |
 | **APP_CONFIG_NAME** | Specifies the name of the ConfigMap and Secret created in the test. | `e2e-runtime-config` |
 | **APP_DEPLOY_NAMESPACE** | Specifies the Namespace of the ConfigMap and Secret created in the test. | `kcp-system` |
+| **APP_BUSOLA_URL** | URL to the expected Kyma Dashboard used when asserting redirection to the UI Console.  | `kcp-system` |
```
IwonaLanger, line 94: In all the other descriptions it is said what a variable does so I'd opt for a more consistent description here. How about something like: `Specifies the URL to the expected Kyma Dashboard`. And if you want to be more specific, you could say: `Specifies the URL to the expected Kyma Dashboard when redirecting to the UI Console` WDYT?

ralikio: Good point @IwonaLanger. I will follow the "Specifies..." convention. However, I would like to leave "used when asserting" as it is. The value is not really used in redirecting part of code but only in the assertion.

IwonaLanger (dismissed):

piotrmiskiewicz (approved):

IwonaLanger (approved):

PR #2692, opened 2023-04-19
Wrong kyma version on retry upgrade command 

berkanercan:
**Description**
This PR introduces the changes required to fix the bug that the kyma version is wrongly calculated when we use kcp cli's retry command. 

Note: Once these changes are merged, another PR is going to be introduced for kcp cli to use these changes. 

- These changes introduce a new property in `orchestrations.KymaParameters` struct called `displayVersion`. 

**Related issue(s)**


components/kyma-environment-broker/internal/orchestration/manager/manager.go
AdrianPei, line 1: we are now populating kyma version at 
https://github.com/kyma-project/control-plane/blob/f3f684c29c704d285f75ee1f5b4e275199915b3b/components/kyma-environment-broker/internal/orchestration/manager/manager.go#L79

berkanercan: Thanks

PR #2712, opened 2023-04-27
Migrate KEB, provisioner and KMC images to artifact registry

MarekMichali:
**Description**

Since we are moving to the artifact registry, we need to switch from using PR images to images built after merging to the main branch. Since we don't yet have a solution to automate image bumping, we have to create a second PR from now on after merging a PR with changes to bump the image version to the image built after merging the previous PR. To summarise the workflow will look like this:
- A PR with some changes is created and merged to main
- The image is built using post build jobs
- A second PR with bumps is created using the image tag from post build jobs

- Move KEB images to artifact registry
- Move Provisioner images to artifact registry
- Move KMC images to artifact registry
- Move test images to artifact registry

**Related issue(s)**


szwedm (approved):
/lgtm

mvshao (approved):
LGTM

PR #2738, opened 2023-05-09
Stoping using of deprecated field EnableBasicAuthentication from shoot

koala7659:
Change requested by Gardener team:

Updating all related code in open-source Kyma to not use Shoot with basic auth any more, as it's a deprecated field which is getting removed very very soon.


koala7659:
This PR requires manual verification before merge

Disper (approved):
LGTM, verified on Dev

PR #2887, opened 2023-06-28
Remove hydra

piotrmiskiewicz:
**Description**

- remove hydra resources from control plane

**Related issue(s)**


PK85 (approved):

triffer:
Removing the template will cause the removal of the resource. This means that we should not merge the PR before 03.07., as this is the day we want to delete the Hydra client from KCP Dev.

PR #2965, opened 2023-08-08
adds a check for timeout tool for provisioner e2e test

Disper:
**Description**

- checks if `timeout` is installed and gives a MacOs hint if not
- ...
- ...

**Related issue(s)**

### DIFF
--- a/components/provisioner/e2e_test/test.sh
+++ b/components/provisioner/e2e_test/test.sh
@@ -60,7 +60,6 @@ fi
 
 printf '\n########## SETTING UP PROVISIONER ##########\n\n'
 
-
 sleep 20
 
 docker network create $POSTGRES_NETWORK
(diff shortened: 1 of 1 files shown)


components/provisioner/e2e_test/test.sh
```diff
@@ -33,6 +33,12 @@ function has_to_succeed {
 function wait_for {
     echo "Waiting $1 to launch on $2"
 
+
+    if ! which timeout ; then
+        printf '\n Please install timeout; e.g. in MacOs you can use `brew install coreutils` \n\n'
+        exit 2
```
VOID404, line 39: All the other `exit`s use code `1`, why use `2` for this one?
Either each should be different or all should be the same imo

VOID404 (approved):

components/provisioner/e2e_test/test.sh
VOID404: /hold
so you can read the comment before merging

Disper: I've changed that exist code to `1`

VOID404 (approved):

Issue #2946, opened 2023-07-28
Provisioner - Extend Provisioning API to support IP ranges for pods and services networks


koala7659:
**Description:**

There is a customer request to extend `ProvisionRuntime` API call with possibility to configure network plugin for created shoot cluster with IP address ranges for pods and services networks.

At the moment Provisioner configures the network plugin only with node network IP address range.  

Example networking configuration in Gardener shoot yaml: 

```yaml
spec: 
  ....
  networking:
    type: calico
    ....
    pods: 100.64.0.0/12
    nodes: 10.250.0.0/16
    services: 100.104.0.0/13
    ....
```

**Suggested implementation:**

New Provisioning parameters will be added to `ProvisionRuntime` mutation. They will be optional and will be passed by the KEB when user provides them in Kyma Runtime creation dialog. 
The GraphQL input contains `GardenerConfigInput` struct (part of `ClusterConfigInput` which is a part of `ProvisionRuntimeInput`) where input shoot parameters are stored .  

In GraphQL schema we will extend `GardenerConfigInput` [struct](https://github.com/kyma-project/control-plane/blob/6eab33fcaf32434530efc06ed267670b7458d549/components/provisioner/pkg/gqlschema/schema.graphql#L211) with two new optional fields: **podsNetworkCidr** and **srvNetworkCidr** :

```graphql
input GardenerConfigInput {
....
  podsNetworkCidr: String
  podsNetworkCidr: String 
  workerCidr: String!
...
}
```
  
If any of new parameters is specified - it will be applied to created shoot as part of its Networking configuration:

``` go
			Networking: &gardener_types.Networking{
				Type:  &networkingType, // Default value - we may consider adding it to API (if Hydroform will support it)
				Nodes: util.StringPtr(c.GardenerProviderConfig.NodeCIDR(c)),
				Services: util.StringPtrOrNil(c.GardenerProviderConfig.ServicesCIDR(c)),
				Pods: util.StringPtrOrNil(c.GardenerProviderConfig.PodsCIDR(c)),
			},
```

Database schema will have to be extended accordingly with new flelds.

We need also to extend `GardenerConfig` [struct](https://github.com/kyma-project/control-plane/blob/6eab33fcaf32434530efc06ed267670b7458d549/components/provisioner/pkg/gqlschema/schema.graphql#L9C11-L9C11) and fill it correctly in `RuntimeStatus` mutation. 

```graphql
input GardenerConfig {
....
  podsNetworkCidr: String
[...]

PR #2968, opened 2023-08-08
Adds ip ranges fields to provisioner

Disper:
**Description**

- adds ip ranges fields to provisioner
- ...
- ...

**Related issue(s)**

#2946


m00g3n (approved):
LGTM

Disper:
I've changed the title to include `do-not-merge/work-in-progress` label as during team meeting @koala7659 raised a concern regarding database migration.

koala7659 (changes requested):
The implemented functionality does not seem to work. Gardener is not able to create a cluster with provided custom configuration for CNI networking plugin. This requires further investigation. Maybe such functionality is not supported at all by the Gardener, or maybe used values should be different? 

image

I can also noticed that values from ProvisionRuntime mutation were passed into shootTemplate as slightly modified. Could they be were changed by Gardener during runtime creation? We need to investigate.

Mutation config

```
                workerCidr: "10.250.0.0/19",
                podsCidr: "10.64.0.0/11",
                servicesCidr: "10.243.0.0/16"  
``` 

Created shoot template: 

```
  networking:
    type: calico
    providerConfig:
      overlay:
        enabled: false
    pods: 10.64.0.0/11
    nodes: 10.250.0.0/16
    services: 10.243.0.0/16
```

I think also we should return values for new fields when they are not provided by the user and are  assigned by the Gardener as we do for seed. Here is an example of current result of the query in case where values are not provided

```
                     "seed": "az-ha-eu1",
                    "workerCidr": "10.250.0.1/19",
                    "podsCidr": null,
                    "servicesCidr": null,
```
All we need to do is to store them in the database just after the moment when the cluster was successfully created. 
You can find the relevant example for seed in the [proceedToInstallation function](https://github.com/kyma-project/control-plane/blob/main/components/provisioner/internal/operations/stages/provisioning/wait_for_cluster_creation.go#L91C1-L114C9) in the Provisioning process.

Disper:
Sharing some experiments that aim to narrow down the issue 
image

Disper:
@koala7659 
> I can also noticed that values from ProvisionRuntime mutation were passed into shootTemplate as slightly modified. Could they be were changed by Gardener during runtime creation? We need to investigate.

I hope I've understood you correctly. Pods/nodes/services are how the gardener call those fields. For some unknown reason, in the past, we've named what represent`nodes` field on the provisioned side as `workedCidr`. Gardener is not changing those field, but rather we're assigning the values to [them here](https://github.com/kyma-project/control-plane/pull/2968/files#diff-503774e2175c9715d5639e33b7b8cd1a66097f93490cd49b6ed4d4ea9e5c1775R195-R198).

> https://github.com/kyma-project/control-plane/pull/2968/files#diff-503774e2175c9715d5639e33b7b8cd1a66097f93490cd49b6ed4d4ea9e5c1775R195-R198
Thanks! I'll look into that.

Disper:
What's left as of now:
- [ ] Fix returning values of new field when they are not provided by the user
- [ ] extend components/provisioner/internal/api/resolver_integration_test.go

koala7659 (approved):
LGTM

grego952 (approved):

PR #3020, opened 2023-08-25
Networking params

piotrmiskiewicz:
**Description**

- allow to specify Network IP ranges: worker CIDR (nodes CIDR), Pods CIDR, services CIDR
- default nodes CIDR: 10.250.0.0/22
- ...

**Related issue(s)**

### DIFF
--- a/components/kyma-environment-broker/internal/broker/instance_create.go
+++ b/components/kyma-environment-broker/internal/broker/instance_create.go
@@ -470,7 +470,7 @@ func (b *ProvisionEndpoint) validateNetworking(parameters internal.ProvisioningP
 	}
 	// error is handled before, in the validate CIDR
 	cidr, _ := netip.ParsePrefix(parameters.Networking.NodesCidr)
-	if cidr.Bits() > 26 {
+	if cidr.Bits() > 23 {
 		err = multierror.Append(err, fmt.Errorf("the suffix of the node CIDR must not be greater than 26"))
 	}
 
--- a/components/kyma-environment-broker/internal/broker/instance_create.go
+++ b/components/kyma-environment-broker/internal/broker/instance_create.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/netip"
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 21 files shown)


piotrmiskiewicz:
can be merged when https://github.com/kyma-project/control-plane/pull/2968 is merges

components/kyma-environment-broker/internal/provider/aws_provider.go
```diff
@@ -121,54 +124,105 @@ func MultipleZonesForAWSRegion(region string, zonesCount int) []string {
...
+	//    PublicCidr:   "10.250.32.0/20",
+	//    InternalCidr: "10.250.48.0/20",
+	// 4 * delta  - difference between two worker (zone) CIDRs
+	delta := big.NewInt(1)
+	delta.Lsh(delta, uint(31-workerPrefixLength))
+
+	// base - it is an integer, which is based on IP bytes
+	base := new(big.Int).SetBytes(workerPrefix.Addr().AsSlice())
+
+	for _, name := range zoneNames {
```
jaroslaw-pieszka, line 163: Is number of zones limited somehow?

components/kyma-environment-broker/internal/provider/azure_provider.go
```diff
@@ -249,15 +276,29 @@ func generateRandomAzureZones(zonesCount int) []int {
 	return zones[:zonesCount]
 }
 
-func generateMultipleAzureZones(zoneNames []int) []*gqlschema.AzureZoneInput {
-	subnetFmt := "10.250.%d.0/19"
-	zones := []*gqlschema.AzureZoneInput{}
-	for i, zone := range zoneNames {
+func generateAzureZones(workerCidr string, zoneNames []int) []*gqlschema.AzureZoneInput {
```
jaroslaw-pieszka, line 279: Are we sure that there won't be more zones than 2?

components/kyma-environment-broker/internal/provider/aws_provider.go
jaroslaw-pieszka: For 6 zones (6 iterations) we will create the zone outside the CIDR from parameter `workerCidr`

components/kyma-environment-broker/internal/broker/instance_create.go
```diff
@@ -427,3 +440,68 @@ func (b *ProvisionEndpoint) createDashboardURL(planID, instanceID string) string
...
+	if parameters.Networking == nil {
+		return nil
+	}
+	var nodes, services, pods *net.IPNet
+	if nodes, e = validateCidr(parameters.Networking.NodesCidr); e != nil {
+		err = multierror.Append(err, fmt.Errorf("while parsing nodes CIDR: %w", e))
+	}
+	// error is handled before, in the validate CIDR
+	cidr, _ := netip.ParsePrefix(parameters.Networking.NodesCidr)
+	if cidr.Bits() > 26 {
```
ebensom, line 473: The maximum Bits we should support is /24, because when going with /24 nodes allows assigning 3x /26 subnet for the 3 worker subnets, while there are still 8x /28 subnets, 6x of which would be used to assign to public and internal ELB subnets.

However to support this scenario, you would also need to adjust the hard-coded `workerPrefixLength := cidr.Bits() + 3` logic, as when /24 was input, you would need to go with `workerPrefixLength := cidr.Bits() + 2`, and `workerPrefixLength + 2` for `public` and `internal`

TL;DR: in general, the smallest CIDR for any workers subnet per AZ should be /26 , as it allows 62 nodes. /27 would be too low already. For Azure, we could even split /24 easily into 4x /26, but for AWS, it's trickier than the current implementation in aws_provider.go

ebensom (changes requested):

components/kyma-environment-broker/internal/broker/instance_create.go
ebensom: Or do you think if we explicitly document to the user than when providing the smallest possible /26 nodes CIDR, the runtime will only be able to scale up to maximum 6 nodes per AZ (18 nodes total) using /29 CIDR?, And furthermore on AWS they will only be able to create 2 public LoadBalancer Services (because of the /30 public CIDR per AZ)?

I think this is a big limitation though

piotrmiskiewicz: I'll add a validation - max value in the suffix is 23

jaroslaw-pieszka (dismissed):

components/kyma-environment-broker/internal/broker/instance_create.go
piotrmiskiewicz: let's do it +2 and suffix 24

jaroslaw-pieszka (approved):

ebensom (approved):

PR #3028, opened 2023-08-29
KEB AvS Maintenance Mode config

szwedm:
**Description**

- ...
- ...
- ...

**Related issue(s)**


components/kyma-environment-broker/cmd/broker/main.go
```diff
@@ -118,17 +118,18 @@ type Config struct {
...
+	ManagedRuntimeComponentsYAMLFilePath                                string
+	NewAdditionalRuntimeComponentsYAMLFilePath                          string
+	SkrOidcDefaultValuesYAMLFilePath                                    string
+	SkrDnsProvidersValuesYAMLFilePath                                   string
+	DefaultRequestRegion                                                string `envconfig:"default=cf-eu10"`
+	UpdateProcessingEnabled                                             bool   `envconfig:"default=false"`
+	UpdateSubAccountMovementEnabled                                     bool   `envconfig:"default=false"`
+	LifecycleManagerIntegrationDisabled                                 bool   `envconfig:"default=true"`
+	ReconcilerIntegrationDisabled                                       bool   `envconfig:"default=false"`
+	AvsMaintenanceModeDuringUpgradeAlwaysDisabledGlobalAccountsFilePath string
```
piotrmiskiewicz, line 132: "long property name" competition won!

components/kyma-environment-broker/internal/process/upgrade_cluster/initialisation.go
```diff
@@ -155,7 +155,12 @@ func (s *InitialisationStep) performRuntimeTasks(step int, operation internal.Up
 
 	switch step {
 	case UpgradeInitSteps:
-		if hasMonitors && !inMaintenance {
+		if s.evaluationManager.IsMaintenanceModeDisabled() {
```
piotrmiskiewicz, line 158: I thnik the performRuntimeTask plays only with AvS and whole code of that method can be skipped when maintanance mode is disabled. Let's discuss

components/kyma-environment-broker/cmd/broker/main.go
szwedm: I knew you'll like it :) The name reflects env var name associated with deployment which is connected to the value in chart's values.yaml. It's just a temporary variable, we will remove it when zero downtime maintenance is default.

components/kyma-environment-broker/internal/process/upgrade_cluster/initialisation.go
szwedm: We discussed the matter and decided to leave the change as it is.

piotrmiskiewicz (approved):

PR #2956, opened 2023-08-02
Workload Identity

lumi017:
**Description**

Changes proposed in this pull request:
Workload Identity toggle

**Related issue(s)**

PR #3069, opened 2023-09-08
Enable workload identity for jobs

lumi017:
**Description**

- Enable workload identity for jobs

**Related issue(s)**

### DIFF
--- a/resources/kcp/charts/kyma-environment-broker/templates/deprovision-retrigger-job.yaml
+++ b/resources/kcp/charts/kyma-environment-broker/templates/deprovision-retrigger-job.yaml
@@ -9,6 +9,7 @@ spec:
     spec:
       template:
         spec:
+          serviceAccountName: {{ .Values.global.kyma_environment_broker.serviceAccountName }}
           shareProcessNamespace: true
           {{- with .Values.deployment.securityContext }}
           securityContext:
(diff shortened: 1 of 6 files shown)


resources/kcp/charts/kyma-environment-broker/templates/deprovision-retrigger-job.yaml
```diff
@@ -78,19 +79,20 @@ spec:
                   readOnly: true
               {{- end}}
             {{- if and (eq .Values.global.database.embedded.enabled false) (eq .Values.global.database.cloudsqlproxy.enabled true)}}
-                - name: cloudsql-instance-credentials
```
ebensom, line 81: If workloadidentidy is disabled we won't mount the cloudsql-instance-credentials, this would still be needed if .Values.global.database.cloudsqlproxy.enabled and not .Values.global.database.cloudsqlproxy.workloadIdentity.enabled

Or I'm missing something? (applicable for all jobs as the pattern is the same).

lumi017: I strongly believe it's useless to mount the credentials there because the connection is established using cloudsql-proxy sidecar and credentials should only be required by it. I have performed a [similar cleanup for KEB](https://github.com/kyma-project/control-plane/pull/2956/files), which works fine after it, and think the jobs in this PR were reusing the pattern from KEB.

lumi017: They are mounted a bit lower, the mount I removed is not for cloudsql-proxy container. I strongly believe it's useless to mount the credentials there because the connection is established using cloudsql-proxy sidecar and credentials should only be required by it. I have performed a [similar cleanup for KEB](https://github.com/kyma-project/control-plane/pull/2956/files), which works fine after it, and think the jobs in this PR were reusing the pattern from KEB.

ebensom (approved):
If you would like, you can also make the cloudsql-instance-credentials volume conditionally rendered only if cloudsqlproxy.enabled and cloudsqlproxy.workloadIdentity.enabled, but of course having the volume and not mounting it in any container won't hurt.

akgalwas (approved):

PR #3112, opened 2023-10-11
extend AvS integration with KEB

i332322:
related: [Kyma runtime AvS compatibility with CE ANS subscription (Outage Mapper)](https://github.tools.sap/kyma/backlog/issues/4251)


i332322:
i've bump the tag version as requested.

ebensom (approved):

PR #3147, opened 2023-10-30
feature: kubeconfig-service secret management change

AdrianPei:
**Description**

- kubeconfig-service can obtain cluster admin kubeconfigs from infrastructure-manager Secrets.
- legacy codes are deleted.
- optimize initialization logic for the service. 

**Related issue(s)**

https://github.tools.sap/kyma/backlog/issues/3116


components/kubeconfig-service/pkg/runtime/runtime_kcp.go
```diff
@@ -308,3 +307,27 @@ func cleanConfigMap(coreClientset kubernetes.Interface, userID string, runtimeID
...
+
+func GetRawConfig(runtimeID string) ([]byte, error) {
+	coreClientset, err := GetK8sClient()
+	if err != nil {
+		log.Errorf("Failed to get kcp k8s client.")
+		return nil, err
+	}
+	secretList, err := coreClientset.CoreV1().Secrets(KcpNamespace).List(context.Background(), metav1.ListOptions{LabelSelector: fmt.Sprintf("kyma-project.io/runtime-id=%s", runtimeID)})
+	if len((*secretList).Items) == 0 {
+		log.Infof("No secrect found for runtime %s.", runtimeID)
```
Disper, line 319: Small typo

```suggestion
		log.Infof("No secret found for runtime %s.", runtimeID)
```

components/kubeconfig-service/pkg/runtime/runtime_kcp.go
```diff
@@ -308,3 +307,27 @@ func cleanConfigMap(coreClientset kubernetes.Interface, userID string, runtimeID
...
+		return nil, k8serrors.NewNotFound(schema.GroupResource{}, "")
+	} else if err != nil {
+		log.Errorf("Failed to filter secrect for runtime %s: %s", runtimeID, err.Error())
```
Disper, line 322: small typo
```suggestion
		log.Errorf("Failed to filter secret for runtime %s: %s", runtimeID, err.Error())
```

！
```diff
@@ -0,0 +1,10 @@
+fix
```
Disper, line 1: I'm not sure if you've committed this file intentionally

components/kubeconfig-service/pkg/runtime/runtime_kcp.go
```diff
@@ -308,3 +307,27 @@ func cleanConfigMap(coreClientset kubernetes.Interface, userID string, runtimeID
...
+		return nil, err
+	}
+
+	var secret []byte
+	for _, item := range (*secretList).Items {
```
Disper, line 327: I'm not sure if there will be a situation when label selector used (`kyma-project.io/runtime-id:%s`) will return more than one secret, but I wonder if it at least should be checked and logged as an error or warn. What do you think?

Disper (commented):
Hi @AdrianPei,
I've left few comments. Have you had a chance to test that already on Dev?

components/kubeconfig-service/pkg/runtime/runtime_kcp.go
AdrianPei: fixed

components/kubeconfig-service/pkg/runtime/runtime_kcp.go
AdrianPei: fixed

！
AdrianPei: my mistake. deteled.

components/kubeconfig-service/pkg/runtime/runtime_kcp.go
AdrianPei: A reasonable suggestion, but I don't know if that will happen, since infrastructure-manager should be responsible for preventing such situation. I will log a warn instead of error here.

AdrianPei:
> Hi @AdrianPei, I've left few comments. Have you had a chance to test that already on Dev?

HI, @Disper .
Yes, I've tested it on Dev already, and it's still running. So fa so good.

components/kubeconfig-service/pkg/runtime/runtime_kcp.go
Disper: Thanks. IM will not create more than one secret, however if someone manually does that, we don't have a mechanism that will decide which secret is a correct one and currently also just logging an error in such case.

Disper (dismissed):

ebensom (commented):
lgtm!

Minor thing: Could you please run `go mod tidy` and commit & push the result as I'm pretty sure the go.mod file will change as kubeconfig-service should not depend on provisioner package any more

AdrianPei:
> lgtm!
> 
> Minor thing: Could you please run `go mod tidy` and commit & push the result as I'm pretty sure the go.mod file will change as kubeconfig-service should not depend on provisioner package any more

Hi @ebensom, go.mod and go.sum updated.

Disper:
Hi @AdrianPei,
out of curiosity, is there any reason you're force-pushing the changes and having a single commit in the PR instead of having a full history of commits?

AdrianPei:
> Hi @AdrianPei, out of curiosity, is there any reason you're force-pushing the changes and having a single commit in the PR instead of having a full history of commits?

Hi @Disper ,
I'm following this [workflow](https://github.com/kyma-project/community/blob/main/docs/contributing/03-git-workflow.md), which suggests to 'leave the top commit, and squash the remaining ones'

ebensom (approved):

PR #3268, opened 2024-01-11
Corrected Invalid Values Reference

ralikio:
**Description**

- corrected invalid reference to .dir property in subscription-cleanup-job.yaml

**Related issue(s)**

### DIFF
--- a/resources/kcp/charts/provisioner/templates/subscription-cleanup-job.yaml
+++ b/resources/kcp/charts/provisioner/templates/subscription-cleanup-job.yaml
@@ -12,7 +12,7 @@ spec:
       template:
         spec:
           containers:
-            - image: "{{ .Values.global.images.containerRegistry.path }}/{{ .Values.global.images.subscription_cleanup_job.dir }}subscription-cleanup-job:{{ .Values.global.images.subscription_cleanup_job.version }}"
+            - image: "{{ .Values.global.images.containerRegistry.path }}/{{ .Values.global.images.subscription_cleanup_job.dir | required "Please specify .global.images.subscription_cleanup_job.dir" }}subscription-cleanup-job:{{ .Values.global.images.subscription_cleanup_job.version }}"
               name: subscription-cleanup-job
               command:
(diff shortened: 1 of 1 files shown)


resources/kcp/charts/provisioner/templates/subscription-cleanup-job.yaml
```diff
@@ -12,7 +12,7 @@ spec:
       template:
         spec:
           containers:
-            - image: "{{ .Values.global.images.containerRegistry.path }}/{{ .Values.global.images.kyma_environments_subscription_cleanup_job.dir }}subscription-cleanup-job:{{ .Values.global.images.subscription_cleanup_job.version }}"
+            - image: "{{ .Values.global.images.containerRegistry.path }}/{{ .Values.global.images.subscription_cleanup_job.dir }}subscription-cleanup-job:{{ .Values.global.images.subscription_cleanup_job.version }}"
```
gczuczy, line 15: When using `.Values.global.images.subscription_cleanup_job.dir` could you please put a `required` on it?
```
{{ .Values.global.images.subscription_cleanup_job.dir | required "Please specify .global.images.subscription_cleanup_job.dir"  }}
```
This is unspecified in this subchart's `values.yaml`, and in the parent chart's `values.yaml` it is specified as an empty value, which should bring a `nil`. This way, without the value unchecked and having no valid defaults, we might be getting unwanted and suboptimal default functionality.

gczuczy (changes requested):

resources/kcp/charts/provisioner/templates/subscription-cleanup-job.yaml
ralikio: For the sake of consistency we would like to leave it as is. Frankly speaking we can get rid of this property at all. Let's do it in further PRs.

Disper (approved):
Not a chart expert myself but I see that it can help with syncing provisioner image that is currently broken.

gczuczy (approved):

PR #3269, opened 2024-01-11
Corrected Image Path for Subscription Cleanup Job

ralikio:
**Description**

Subscription Cleanup Job (SCJ) has been moved outside of `control-plane` directory. This PR updates paths to allow usage of no dir value for SCJ but leaving other places unchanged.

- set dir to `control-plane` for all components,
- removed `control-plane` from `containerRegistry` property,
- left dir empty for SCJ.

**Related issue(s)**


Disper (dismissed):
I don't fully grasp why this change has happened but it might unblock images syncing on dev

Disper (approved):

PR #3308, opened 2024-01-31
Version check for kcp cli

berkanercan:
**Description**

- We pull latest git tag and compare it with the local version of kcp cli. If there is a mismatch, it warns the user and provides link to update kcp cli

**Related issue(s)**


berkanercan:
Implemented in kcp-cli repo

PR #3323, opened 2024-02-06
The Provisioner adds OpenStack parameters

mvshao:
**Description**

From now Provisioner is responsible for adding (if KEB does not provide) these parameters during provision of the OpenStack cluster:
- `FloatingPoolName`
- `ExposureClassName`
- GraphQL `provisionRuntime` mutation now does not require `floatingPoolName` to be provided


components/provisioner/internal/api/resolver_integration_test.go
```diff
@@ -211,7 +211,7 @@ func openStackGardenerClusterConfigInput() gqlschema.ClusterConfigInput {
 			ProviderSpecificConfig: &gqlschema.ProviderSpecificInput{
 				OpenStackConfig: &gqlschema.OpenStackProviderConfigInput{
 					Zones:                []string{"eu-de-1a"},
-					FloatingPoolName:     "FloatingIP-external-cp",
+					FloatingPoolName:     util.StringPtr("FloatingIP-external-cp"),
```
VOID404, line 214: could we move `util.*Ptr` to generics?

VOID404 (approved):

Issue #3306, opened 2024-01-30
rework KMC alerting (pair)


k15r:
## Description

As KMC is used for billing users, it is essential that it exposes metrics usable to fire alerts whenever the billing would be impacted. To ensure this, the metrics exposed by KMC and the alerting rules using these new metrics need to be reworked.

> **Abbreviations**
> - KMC sends the data needed for billings users to the **Event Data Platform** (EDP).
> - **Kyma Environment Broker** [(KEB)](https://github.com/kyma-project/kyma-environment-broker/) is a component that allows provisioning Kyma as a runtime on clusters provided by third-party providers.

## Helpful Links
- Overview of [current metrics](https://github.com/kyma-project/control-plane/blob/main/components/kyma-metrics-collector/metrics.md) in KMC
- [Example metrics](https://github.com/kyma-project/control-plane/blob/main/components/kyma-metrics-collector/pkg/edp/metrics.go) in KMC
- Golang [Prometheus package](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus)
- Prometheus [Metrics Intro](https://prometheus.io/docs/concepts/metric_types/)
- Prometheus [Naming Convention](https://prometheus.io/docs/practices/naming/)
- Prometheus [Alerting Best Practices](https://prometheus.io/docs/practices/alerting/)

## Acceptance
- [ ] ensure all clusters (listed by KEB) which have to be considered active are scraped by KMC
  - in the [KEB output](https://github.com/kyma-project/kyma-environment-broker/blob/main/common/runtime/model.go): `.status.state` is set ot one of these values per SKR: `succeeded` , `error` , `upgrading` , `updating`
  - currently we rely on `.status.provisioning.state` == `succeeded`
- [ ] check with KEB team for a metric which contains all active clusters which need to be scraped (all clusters with status is also fine). If they dont have any, then expose one in KMC
- [x] Metric per scraped instance for overall success / error.
- [x] Metric per subcomponent (keb, gardener, edp) for overall success / error.
  - extend `kmc_skr_calls_total` with labels for `instanceID`, `shootID`, `subaccount`, `globalaccount`
- [x] Metric for number of consecutive resends to edp per cluster
- [x] `kmc_edp_request_total` contains information about which clusters data was sent (label `instanceID`, `shootID`, `subaccount`, `globalaccount`)
[...]

PR #3334, opened 2024-02-12
feat: scrape all active clusters

grischperl:
**Description**

- refine which clusters are trackable
- introduce new `kmc_process_fetched_clusters` metric
- add tests
- add new alerting rule to alert for:
   Rate of change of the number of clusters obtained by KMC from KEB for each poll.
   No data obtained by KMC from KEB.
- add alerting tests

**Related issue(s)**

#3306


components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,25 +353,75 @@ func isSuccess(status int) bool {
 	return false
 }
 
-func isClusterTrackable(runtime *kebruntime.RuntimeDTO) bool {
-	if runtime.Status.Provisioning != nil &&
-		runtime.Status.Provisioning.State == string(kebruntime.StateSucceeded) &&
-		runtime.Status.Deprovisioning == nil {
+func (p *Process) isClusterTrackable(runtime *kebruntime.RuntimeDTO) bool {
+	// Check if the cluster is in a trackable state
+	trackableStates := map[kebruntime.State]bool{
```
mfaizanse, line 358: (nit) How about we create a method like:
```
func isTrackableState(state string) bool {
	switch state {
	case "succeeded", "error", "upgrading", "updating":
		return true
	}
	return false
}
```

components/kyma-metrics-collector/pkg/process/metrics.go
```diff
@@ -48,8 +48,29 @@ var (
...
 		[]string{shootNameLabel, instanceIdLabel, runtimeIdLabel, subAccountLabel, globalAccountLabel},
 	)
+	kebTotalClusters = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "clusters_total",
+			Help:      "Number of all clusters got from KEB including trackable and not trackable.",
+		},
+		[]string{"trackable", shootNameLabel, instanceIdLabel, runtimeIdLabel, subAccountLabel, globalAccountLabel},
```
mfaizanse, line 58: We can also define a const for "trackable".

components/kyma-metrics-collector/pkg/process/metrics.go
```diff
@@ -48,8 +48,29 @@ var (
 		},
 		[]string{shootNameLabel, instanceIdLabel, runtimeIdLabel, subAccountLabel, globalAccountLabel},
 	)
+	kebTotalClusters = promauto.NewGaugeVec(
```
mfaizanse, line 51: Can we use counter here? Seems like we are going to reset the metric anyway.

components/kyma-metrics-collector/pkg/process/process.go
mfaizanse: We can even use `string(kebruntime.StateXYZ)` instead of strings.

resources/kcp/charts/kyma-metrics-collector/prometheus/alerts.yaml
```diff
@@ -64,4 +64,13 @@ groups:
       labels:
         severity: critical
       annotations:
-        description: Rate of increase in errors for the requests from KMC to Gardener.
\ No newline at end of file
+        description: Rate of increase in errors for the requests from KMC to Gardener.
+  - name: kmc.rules.processing
+    rules:
+    - alert: KMCProcessingClustersFromKEBFailed
```
mfaizanse, line 70: The name is conflicting with other alerts. How about something like `KMCFetchingClustersFromKEBFailed` or similar?

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,25 +353,75 @@ func isSuccess(status int) bool {
...
+		"updating":  true,
+	}
+
+	if trackableStates[runtime.Status.State] ||
+		(runtime.Status.Provisioning != nil &&
+			runtime.Status.Provisioning.State == string(kebruntime.StateSucceeded) &&
+			runtime.Status.Deprovisioning == nil) {
+
+		// record kebTotalClusters metric
+		recordKEBTotalClusters(
```
mfaizanse, line 371: `isClusterTrackable` is a helper function which called by `populateCacheAndQueue` (which is more of a business logic function), IMO we should track the metrics in `populateCacheAndQueue`.

Also we are resetting the metric in `populateCacheAndQueue`, so it would be good if we manage the metric in this function.

What do you think?

resources/kcp/charts/kyma-metrics-collector/prometheus/alerts.yaml
```diff
@@ -64,4 +64,13 @@ groups:
...
+      expr: sum(rate(kmc_process_clusters_total[15m])) > 0
```
mfaizanse, line 71: Can you explain this rule?

resources/kcp/charts/kyma-metrics-collector/prometheus/alerts.yaml
```diff
@@ -64,4 +64,13 @@ groups:
...
+      for: 10m
+      labels:
+        severity: critical
+      annotations:
+        description: Processing clusters obtained from KEB failed.
```
mfaizanse, line 76:
```suggestion
        description: Fetching clusters obtained from KEB failed.
```

resources/kcp/charts/kyma-metrics-collector/prometheus/alerts.yaml
grischperl: Changed rule to use percentage

resources/kcp/charts/kyma-metrics-collector/prometheus/alerts.yaml
grischperl: Done

resources/kcp/charts/kyma-metrics-collector/prometheus/alerts.yaml
grischperl: Done

components/kyma-metrics-collector/pkg/process/metrics.go
grischperl: true, done!

components/kyma-metrics-collector/pkg/process/metrics.go
grischperl: done

components/kyma-metrics-collector/pkg/process/process.go
grischperl: good point, done

components/kyma-metrics-collector/pkg/process/process.go
grischperl: Done!

components/kyma-metrics-collector/pkg/process/metrics.go
```diff
@@ -18,6 +19,7 @@ const (
 	globalAccountLabel = "global_account_id"
 	successLabel       = "success"
 	withOldMetricLabel = "with_old_metric"
+	trackable          = "trackable"
```
marcobebway, line 22:
```suggestion
	trackableLabel          = "trackable"
```

marcobebway, line 22: Please change `trackable` to `trackableLabel`.

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,25 +353,59 @@ func isSuccess(status int) bool {
 	return false
 }
 
-func isClusterTrackable(runtime *kebruntime.RuntimeDTO) bool {
-	if runtime.Status.Provisioning != nil &&
-		runtime.Status.Provisioning.State == string(kebruntime.StateSucceeded) &&
-		runtime.Status.Deprovisioning == nil {
+// Helper function which returns true if the runtime state is trackable
```
marcobebway, line 356:
```suggestion
// isTrackableState returns true if the runtime state is trackable, otherwise returns false.
```

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,25 +353,59 @@ func isSuccess(status int) bool {
...
+func isTrackableState(state kebruntime.State) bool {
+	switch state {
+	case kebruntime.StateSucceeded, kebruntime.StateError, kebruntime.StateUpgrading, kebruntime.StateUpdating:
```
marcobebway, line 359: Are these all the available states?

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,25 +353,59 @@ func isSuccess(status int) bool {
...
 		return true
 	}
 	return false
 }
 
+func (p *Process) isClusterTrackable(runtime *kebruntime.RuntimeDTO) bool {
```
marcobebway, line 365: The function name implies that we only need read access for the passed `runtime` argument, so please do not pass a pointer here so we do not accidentally change the object internal state.

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,25 +353,59 @@ func isSuccess(status int) bool {
...
+	// Check if the cluster is in a trackable state
+	if isTrackableState(runtime.Status.State) ||
+		(runtime.Status.Provisioning != nil &&
+			runtime.Status.Provisioning.State == string(kebruntime.StateSucceeded) &&
+			runtime.Status.Deprovisioning == nil) {
```
marcobebway, line 370: This is hard to read, could you please encapsulate this logic in a function?
From the logic I read, the function signature could be `func isProvisioned(runtime) bool`, what do you think?

components/kyma-metrics-collector/pkg/process/metrics.go
grischperl: Done

components/kyma-metrics-collector/pkg/process/process.go
grischperl: Done

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,25 +353,59 @@ func isSuccess(status int) bool {
...
+		return true
+	}
+
+	// Set provisioning and deprovisioning states if available
+	provisioning := getOrDefault(runtime.Status.Provisioning, "")
+	deprovisioning := getOrDefault(runtime.Status.Deprovisioning, "")
+
+	p.namedLogger().With(log.KeySubAccountID, runtime.SubAccountID).With(log.KeyRuntimeID, runtime.RuntimeID).
+		Infof("Not tracking cluster of state: %s, provisioning state: %s, deprovisioning state: %s",
+			runtime.Status.State, provisioning, deprovisioning)
```
marcobebway, line 384: As per the comment `// Set provisioning and deprovisioning states if available`, is this true, because I see `getOrDefault` then logging, so no set operation was done. Also the function name implies that it checks if a cluster is trackable or not, IMO it should not internally set anything. What do you think?

marcobebway, line 384: Please apply the same logging pattern described [here](https://github.com/kyma-project/control-plane/pull/3334/files#r1500365859).

components/kyma-metrics-collector/pkg/process/process.go
grischperl: Done

components/kyma-metrics-collector/pkg/process/process.go
marcobebway, line 365: Does it make sense to simplify the function logic to do something similar to the following pseudo-code?

```golang
func (p *Process) isRuntimeTrackable(runtime *kebruntime.RuntimeDTO) bool {
	// log some stuff

	return isTrackableState(runtime.Status.State) || isProvisioned(runtime)
}
```

grischperl: Done

marcobebway, line 365: Does it make sense to change the name from `isClusterTrackable` to `isRuntimeTrackable`.

components/kyma-metrics-collector/pkg/process/process.go
grischperl: There are more possible states, but only the ones provided here make the runtime trackable.

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,25 +353,59 @@ func isSuccess(status int) bool {
...
+	return false
+}
+
+// Helper function to get state or default value if nil
```
marcobebway, line 388:
```suggestion
// getOrDefault Helper function to get state or default value if nil
```

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,25 +353,59 @@ func isSuccess(status int) bool {
...
+func getOrDefault(runtimeStatus *kebruntime.Operation, defaultValue string) string {
+	if runtimeStatus != nil {
+		return runtimeStatus.State
```
marcobebway, line 391: Could `runtimeStatus.State` be nil?

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -383,6 +417,15 @@ func (p *Process) populateCacheAndQueue(runtimes *kebruntime.RuntimesPage) {
 				Metric:          nil,
 			}
 
+			// record kebFetchedClusters metric for trackable cluster
+			recordKEBFetchedClusters(
+				true,
```
marcobebway, line 422: Please create a new const and use it here:

```golang
const trackable = true
```

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -416,6 +459,16 @@ func (p *Process) populateCacheAndQueue(runtimes *kebruntime.RuntimesPage) {
 			}
 			continue
 		}
+
+		// record kebFetchedClusters metric for not trackable clusters
+		recordKEBFetchedClusters(
+			false,
```
marcobebway, line 465: Please create a new const and use it here:

```golang
const trackableFalse = false
```

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,25 +353,59 @@ func isSuccess(status int) bool {
...
+func (p *Process) isClusterTrackable(runtime *kebruntime.RuntimeDTO) bool {
+	// Check if the cluster is in a trackable state
+	if isTrackableState(runtime.Status.State) ||
+		(runtime.Status.Provisioning != nil &&
+			runtime.Status.Provisioning.State == string(kebruntime.StateSucceeded) &&
+			runtime.Status.Deprovisioning == nil) {
+
+		p.namedLogger().With(log.KeySubAccountID, runtime.SubAccountID).With(log.KeyRuntimeID, runtime.RuntimeID).
+			Infof("Tracked cluster of state: %s, provisioning state: %s, no deprovisioning state.",
+				runtime.Status.State, runtime.Status.Provisioning.State)
```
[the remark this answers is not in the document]
grischperl: Done

components/kyma-metrics-collector/pkg/process/process_test.go
```diff
@@ -37,7 +38,7 @@ const (
 	bigTimeout = 10 * time.Second
 
 	// KEB related variables.
-	kebRuntimeResponseFilePath = "../testing/fixtures/runtimes_response.json"
```
marcobebway, line 40: Is the `../testing/fixtures/runtimes_response.json` file still used in the tests?

components/kyma-metrics-collector/pkg/process/process.go
grischperl: Done

grischperl: Done

components/kyma-metrics-collector/pkg/process/process.go
grischperl: `runtime.Status.Provisioning` and `runtime.Status.Deprovisioning` can be nil. If we try to log them, when they are nil, we get an error. As a work around, the logic `getOrDefault` returns an empty string if there is not available provisioning or deprovisioning status. 

I changed the comment to `Get provisioning and deprovisioning states if available otherwise return empty string` to be more clear.

components/kyma-metrics-collector/pkg/process/process.go
grischperl: Done

components/kyma-metrics-collector/pkg/process/process.go
grischperl: Done

components/kyma-metrics-collector/pkg/process/process.go
grischperl: Done

components/kyma-metrics-collector/pkg/process/process_test.go
grischperl: Yes, it is used in a different test.

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -362,16 +377,47 @@ func isClusterTrackable(runtime *kebruntime.RuntimeDTO) bool {
 	return false
 }
 
+func isRuntimeTrackable(runtime kebruntime.RuntimeDTO) bool {
+	// Check if the runtime is trackable
+	if isTrackableState(runtime.Status.State) || isProvisionedStatus(runtime) {
+		return true
+	}
+	return false
+}
```
marcobebway, line 386:
```suggestion
func isRuntimeTrackable(runtime kebruntime.RuntimeDTO) bool {
	return isTrackableState(runtime.Status.State) || isProvisionedStatus(runtime)
}
```

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,7 +358,17 @@ func isSuccess(status int) bool {
 	return false
 }
 
-func isClusterTrackable(runtime *kebruntime.RuntimeDTO) bool {
+// isTrackableState returns true if the runtime state is trackable, otherwise returns false.
+func isTrackableState(state kebruntime.State) bool {
```
marcobebway, line 362: Is there a unit-test for this function?

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -353,7 +358,17 @@ func isSuccess(status int) bool {
...
+	switch state {
+	case kebruntime.StateSucceeded, kebruntime.StateError, kebruntime.StateUpgrading, kebruntime.StateUpdating:
+		return true
+	}
+	return false
+}
+
+// isProvisionedStatus returns true if the runtime is successfully provisioned, otherwise returns false.
+func isProvisionedStatus(runtime kebruntime.RuntimeDTO) bool {
```
marcobebway, line 371: Is there a unit-test for this function?

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -362,16 +377,47 @@ func isClusterTrackable(runtime *kebruntime.RuntimeDTO) bool {
 	return false
 }
 
+func isRuntimeTrackable(runtime kebruntime.RuntimeDTO) bool {
```
marcobebway, line 380: Is there a unit-test for this function?

components/kyma-metrics-collector/pkg/process/process.go
```diff
@@ -362,16 +377,47 @@ func isClusterTrackable(runtime *kebruntime.RuntimeDTO) bool {
...
+	// Check if the runtime is trackable
+	if isTrackableState(runtime.Status.State) || isProvisionedStatus(runtime) {
+		return true
+	}
+	return false
+}
+
+// getOrDefault returns the runtime state or a default value if runtimeStatus is nil
+func getOrDefault(runtimeStatus *kebruntime.Operation, defaultValue string) string {
```
marcobebway, line 385: Is there a unit-test for this function?

marcobebway (approved):

PR #3369, opened 2024-02-20
[CCEE] Set Cloud Profile as default

mvshao:
**Description**

- Provisioner sets `cloud-profile` OpenStack parameters field with default value

### DIFF
--- a/components/provisioner/internal/provisioning/input_converter.go
+++ b/components/provisioner/internal/provisioning/input_converter.go
@@ -15,6 +15,7 @@ import (
 const (
 	OpenStackFloatingPoolName  = "FloatingIP-external-kyma-01"
 	OpenStackExposureClassName = "converged-cloud-internet"
+	OpenStackCloudProfileName  = "converged-cloud-kyma"
 )
 
 type InputConverter interface {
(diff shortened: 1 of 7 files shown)


tobiscr (approved):

components/provisioner/internal/provisioning/input_converter.go
```diff
@@ -234,6 +235,9 @@ func (c converter) providerSpecificConfigFromInput(input *gqlschema.ProviderSpec
 		if input.OpenStackConfig.FloatingPoolName == nil {
 			input.OpenStackConfig.FloatingPoolName = util.PtrTo(OpenStackFloatingPoolName)
 		}
+		if input.OpenStackConfig.CloudProfileName == nil {
+			input.OpenStackConfig.CloudProfileName = util.PtrTo(OpenStackCloudProfileName)
+		}
```
VOID404, line 240: We could use [`OkOrDefault`](https://github.com/kyma-project/control-plane/blob/ee936771fbc59754f47cdf8c95f7db70d79e9a67/components/provisioner/internal/util/pointers.go#L26-L32) for both of those

PR #3370, opened 2024-02-20
fix: changed retry policy for EDP and KEB in KMC

mfaizanse:
**Description**

- change retry policy for EDP and KEB in KMC, because the EDP retry was getting stuck in backoff.

**Related issue(s)**

- https://github.com/kyma-project/control-plane/issues/3306

### DIFF
--- a/components/kyma-metrics-collector/pkg/edp/client.go
+++ b/components/kyma-metrics-collector/pkg/edp/client.go
@@ -74,12 +74,11 @@ func (eClient Client) Send(req *http.Request, payload []byte) (*http.Response, e
 	var resp *http.Response
 	var err error
 	defer func() {
-		if resp == nil {
-			return
-		}
-		err := resp.Body.Close()
-		if err != nil {
-			eClient.namedLogger().Warn(err)
+		if resp != nil && resp.Body != nil {
+			err := resp.Body.Close()
+			if err != nil {
+				eClient.namedLogger().Warn(err)
+			}
 		}
 	}()
 
--- a/components/kyma-metrics-collector/pkg/edp/client.go
+++ b/components/kyma-metrics-collector/pkg/edp/client.go
@@ -71,28 +71,18 @@ func (eClient Client) getEDPURL(dataTenant string) string {
 }
 
 func (eClient Client) Send(req *http.Request, payload []byte) (*http.Response, error) {
-	var resp *http.Response
-	var err error
-	defer func() {
-		if resp != nil && resp.Body != nil {
-			err := resp.Body.Close()
-			if err != nil {
-				eClient.namedLogger().Warn(err)
-			}
-		}
-	}()
-
+	// define retry policy.
 	retryOptions := []retry.Option{
 		retry.Attempts(uint(eClient.Config.EventRetry)),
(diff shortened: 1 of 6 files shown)


components/kyma-metrics-collector/pkg/edp/client.go
```diff
@@ -74,13 +73,7 @@ func (eClient Client) getEDPURL(dataTenant string) string {
...
 	var resp *http.Response
 	var err error
-	customBackoff := wait.Backoff{
-		Steps:    eClient.Config.EventRetry,
-		Duration: eClient.Config.Timeout,
-		Factor:   5.0,
-		Jitter:   0.1,
-	}
-	err = retry.OnError(customBackoff, func(err error) bool {
+	err = retry.OnError(retry.DefaultRetry, func(err error) bool {
```
mfaizanse, line 76:
```
var DefaultRetry = wait.Backoff{
	Steps:    5,
	Duration: 10 * time.Millisecond,
	Factor:   1.0,
	Jitter:   0.1,
}
```

components/kyma-metrics-collector/pkg/edp/client.go
```diff
@@ -74,60 +73,62 @@ func (eClient Client) getEDPURL(dataTenant string) string {
...
-		resp, err = eClient.HttpClient.Do(req)
-		duration := time.Since(reqStartTime)
-		// check result.
-		if err != nil {
-			urlErr := err.(*url.Error)
-			responseCode := http.StatusBadRequest
-			if urlErr.Timeout() {
-				responseCode = http.StatusRequestTimeout
+	defer func() {
+		if resp != nil && resp.Body != nil {
```
k15r, line 77: according to the spec the body will never be nil, unless you get an error from the http.client

> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.


but you should defer the close after getting the response and checking it for error (line 94)

components/kyma-metrics-collector/pkg/keb/client.go
```diff
@@ -79,68 +76,78 @@ func (c Client) GetAllRuntimes(req *http.Request) (*kebruntime.RuntimesPage, err
 }
 
 func (c Client) getRuntimesPerPage(req *http.Request, pageNum int) (*kebruntime.RuntimesPage, error) {
+	var resp *http.Response
+	var err error
+	defer func() {
+		if resp != nil && resp.Body != nil {
+			err := resp.Body.Close()
```
k15r, line 83: same as the previous situation:
do the request, get the response, check for error, defer the body close

components/kyma-metrics-collector/pkg/edp/client.go
mfaizanse: refactored.

components/kyma-metrics-collector/pkg/keb/client.go
mfaizanse: refactored.

k15r (approved):

PR #3415, opened 2024-03-20
Provisioner creates "istio-system" namespace as part of "create operator bindings" step

koala7659:
One of installed Kyma componenets "compass-runtime-agent" requires `istio-system` namespace to be correctly initialised. This namespace was created by the Provisioner as part of Runtime configuration step where Provisioner created secrets on the cluster that allowed to establish connection between compass-runtime-agent and Compass.

New version of the provisioner delivered by [this PR ](https://github.com/kyma-project/control-plane/pull/3399)makes Compass Registration process optional, controlled by env variables.

There was a bug in the code where after disabling Compass Registration - The Provisioner has stopped creating `istio-system` namespace as well. 

This PR is suppose to mitigate this problem by moving namespace creation code into "create operator bindings" step


m00g3n (approved):
LGTM

PR #3429, opened 2024-04-25
provisioner: add support for configuring AWS IMDSv2 access method

ebensom:
**Description**

- Add new Gardener configuration `DefaultEnableIMDSv2` as a feature flag to enabled IMDSv2 support during provisioning and shoot upgrade
- Add new fields in AWSProviderConfig and AWSProviderConfigInput GQL schema
- In input converter, support using the default AWS IMDVSv2 configuration if it was provided as input
- Enhance `AWSGardenerConfig.ExtendShootConfig()` to add the needed providerConfig to the single worker if `enableIMDSv2` is true
- Enhance `AWSGardenerConfig.EditShootConfig()` to check if the needed providerConfig is present in the shoot, and ensure it is configured with in the desired state, in case the `enableIMDSv2` is true

**Related issue(s)**

https://github.tools.sap/kyma/backlog/issues/5218

### DIFF
--- a/components/provisioner/internal/model/gardener_config.go
+++ b/components/provisioner/internal/model/gardener_config.go
@@ -553,8 +553,9 @@ func (c AWSGardenerConfig) AsProviderSpecificConfig() gqlschema.ProviderSpecific
 	}
 
 	return gqlschema.AWSProviderConfig{
-		AwsZones: zones,
-		VpcCidr:  &c.input.VpcCidr,
+		AwsZones:     zones,
+		VpcCidr:      &c.input.VpcCidr,
+		EnableIMDSv2: c.input.EnableIMDSv2,
 	}
 }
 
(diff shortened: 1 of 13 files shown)


components/provisioner/internal/model/gardener_config.go
```diff
@@ -590,7 +591,42 @@ func (c AWSGardenerConfig) ValidateShootConfigChange(shoot *gardener_types.Shoot
...
+		return err
+	}
+
+	if c.input.EnableIMDSv2 != nil && *c.input.EnableIMDSv2 {
+		var (
+			workerConfig            *aws.WorkerConfig
+			httpPutResponseHopLimit int64 = 2
+		)
+		if shoot.Spec.Provider.Workers[0].ProviderConfig == nil {
+			workerConfig = NewAWSWorkerConfig(2)
```
mvshao, line 605: Instead of passing the magic number we can use, defined couple lines above, variable `httpPutResponseHopLimit`. That will improve code transparency.

components/provisioner/internal/model/gardener_config.go
```diff
@@ -612,6 +648,15 @@ func (c AWSGardenerConfig) ExtendShootConfig(gardenerConfig GardenerConfig, shoo
 		return apperrors.Internal("error encoding control plane config: %s", err.Error())
 	}
 
+	if c.input.EnableIMDSv2 != nil && *c.input.EnableIMDSv2 {
+		awsWorkerConfig := NewAWSWorkerConfig(2)
```
mvshao, line 652: Same here, passing the magic number to this function could led to the confusion. We could think about a good way of passing this value, but it up to you.

components/provisioner/internal/model/gardener_config.go
ebensom: Introduced global variable for AWS IMDSv2 HTTPPutResponseHopLimit, and it's used everywhere instead of the "2" magic number.

components/provisioner/internal/model/gardener_config.go
ebensom: Introduced global variable for AWS IMDSv2 HTTPPutResponseHopLimit, and it's used everywhere instead of the "2" magic number.

mvshao (approved):
LGTM

